PR #100, opened 2026-04-09
feat: add storage create-bucket, create-table, upload-table

ottomansky:
## Summary

- Adds `kbagent storage create-bucket` — creates a bucket (stage, name, optional description/backend)
- Adds `kbagent storage create-table` — creates a typed table with `--column name:TYPE` syntax (STRING, INTEGER, NUMERIC, FLOAT, BOOLEAN, DATE, TIMESTAMP) and optional `--primary-key`
- Adds `kbagent storage upload-table` — uploads a CSV file into an existing table (full or `--incremental` load, custom delimiter/enclosure)

All three follow the existing 3-layer pattern (client → service → command) and are tested in `tests/test_storage_write.py` matching the style of `test_storage_delete.py`.

Also fixes a pre-existing Windows bug in `config_store.py`: `os.replace()` failed with `WinError 5` because the lock fd held the destination file open. On Windows (no fcntl), the fd is now closed before `os.replace()`. This was blocking all config writes and 561 tests on Windows.

## Verified against connection.europe-west3.gcp.keboola.com

```
kbagent storage create-bucket --project p --stage in --name test-write-01
kbagent storage create-table  --project p --bucket-id in.c-test-write-01 --name users \
  --column id:INTEGER --column name:STRING --column created_at:TIMESTAMP --primary-key id
kbagent storage upload-table  --project p --table-id in.c-test-write-01.users --file data.csv
# → {"imported_rows": 2, "warnings": []}
```

## Notes for reviewer

- `upload-table` import endpoint is **synchronous** (returns results directly, not a Storage job) — confirmed from real API response. Did not use `_wait_for_storage_job` for this reason.
- Row count field from API is `importedRowsCount` (not `importedRows` or `rowsCount`).

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,6 @@ markers = [
 target-version = "py312"
 line-length = 100
 src = ["src", "tests"]
-exclude = ["tests/_test_upload_client_append.py"]
 
 [tool.ruff.lint]
 select = ["E", "F", "I", "W", "UP", "B", "SIM", "RUF"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ markers = [
 target-version = "py312"
 line-length = 100
 src = ["src", "tests"]
+exclude = ["tests/_test_upload_client_append.py"]
 
 [tool.ruff.lint]
 select = ["E", "F", "I", "W", "UP", "B", "SIM", "RUF"]
--- a/plugins/kbagent/skills/kbagent/SKILL.md
+++ b/plugins/kbagent/skills/kbagent/SKILL.md
@@ -68,6 +68,9 @@ If kbagent is not installed or you need the full standalone reference, run `kbag
 | List storage buckets with sharing/linked bucket information | `kbagent storage buckets` |
 | Show detailed bucket info including Snowflake direct access paths | `kbagent storage bucket-detail --project PROJECT --bucket-id BUCKET-ID` |
 | List storage tables from a project | `kbagent storage tables --project PROJECT` |
+| Create a new storage bucket | `kbagent storage create-bucket --project PROJECT --stage STAGE --name NAME` |
+| Create a typed storage table | `kbagent storage create-table --project PROJECT --bucket-id BUCKET-ID --name NAME --column col:TYPE` |
+| Upload a CSV file into a storage table | `kbagent storage upload-table --project PROJECT --table-id TABLE-ID --file FILE` |
 | Delete one or more storage tables | `kbagent storage delete-table --project PROJECT --table-id TABLE-ID` |
 | Delete one or more storage buckets | `kbagent storage delete-bucket --project PROJECT --bucket-id BUCKET-ID` |
 | List shared buckets available for linking | `kbagent sharing list` |
--- a/src/keboola_agent_cli/client.py
+++ b/src/keboola_agent_cli/client.py
@@ -884,12 +884,11 @@ class KeboolaClient(BaseHttpClient):
         Returns:
             File resource dict including 'id' (fileId), 'url', and 'uploadParams'.
         """
-        body: dict[str, Any] = {
-            "name": name,
-            "sizeBytes": size_bytes,
-            "isPermanent": False,
-            "isPublic": False,
-        }
+        # Only send the two required fields. Omit optional booleans (isPermanent,
+        # isPublic) so Python's False doesn't serialize to the string "False" in
+        # form-data — which some backends treat as truthy, potentially making
+        # files permanent or publicly accessible.
+        body: dict[str, Any] = {"name": name, "sizeBytes": size_bytes}
         response = self._request("POST", "/v2/storage/files/prepare", data=body)
         return response.json()
 
@@ -901,7 +900,9 @@ class KeboolaClient(BaseHttpClient):
         """Upload a file to cloud storage using the presigned URL from files/prepare.
 
         Handles S3 and GCS presigned POST uploads. The uploadParams fields are
-        posted as form data with the file appended last (required by S3).
+        posted as multipart form data with the file appended last (S3 requires
+        the file to be the last part). Azure Blob Storage (ABS) backends use a
+        different upload mechanism (PUT to a SAS URL) and are not yet supported.
 
         Args:
             upload_info: Response from prepare_file_upload().
@@ -984,7 +985,8 @@ class KeboolaClient(BaseHttpClient):
             Import results dict with importedRowsCount, warnings, etc.
         """
         p = Path(file_path)
-        upload_info = self.prepare_file_upload(name=p.name, size_bytes=p.stat().st_size)
+        size_bytes = p.stat().st_size
+        upload_info = self.prepare_file_upload(name=p.name, size_bytes=size_bytes)
         file_id = upload_info["id"]
         self._upload_to_cloud(upload_info, file_path)
         job = self.import_table_async(
--- a/src/keboola_agent_cli/client.py
+++ b/src/keboola_agent_cli/client.py
@@ -882,7 +882,9 @@ class KeboolaClient(BaseHttpClient):
             size_bytes: File size in bytes.
 
         Returns:
-            File resource dict including 'id' (fileId), 'url', and 'uploadParams'.
+            File resource dict including 'id' (fileId), 'url', 'uploadParams',
+            and 'gcsUploadParams' (present on GCP stacks; contains bearer token
+            and GCS bucket/key for direct PUT upload).
         """
         # Only send the two required fields. Omit optional booleans (isPermanent,
         # isPublic) so Python's False doesn't serialize to the string "False" in
--- a/src/keboola_agent_cli/client.py
+++ b/src/keboola_agent_cli/client.py
@@ -897,29 +897,63 @@ class KeboolaClient(BaseHttpClient):
         upload_info: dict[str, Any],
         file_path: str,
     ) -> None:
-        """Upload a file to cloud storage using the presigned URL from files/prepare.
+        """Upload a file to cloud storage using credentials from files/prepare.
 
-        Handles S3 and GCS presigned POST uploads. The uploadParams fields are
-        posted as multipart form data with the file appended last (S3 requires
-        the file to be the last part). Azure Blob Storage (ABS) backends use a
-        different upload mechanism (PUT to a SAS URL) and are not yet supported.
+        Three upload paths based on what the API returns:
+
+        GCP stack (``gcsUploadParams`` present):
+            PUT to ``https://storage.googleapis.com/{bucket}/{key}`` with an
+            OAuth2 ``Authorization: Bearer`` header. The ``url`` field in the
+            response is a download URL and is NOT used for upload.
+
+        S3 stack (``uploadParams`` non-empty):
+            Multipart form POST to ``url``. All ``uploadParams`` fields are
+            included as form fields first; the file field is appended last
+            (S3 policy requires the file to be the final part).
+
+        ABS/other signed URL (``uploadParams`` empty/null):
+            Raw PUT of the file bytes directly to ``url``.
 
         Args:
-            upload_info: Response from prepare_file_upload().
+            upload_info: Full response dict from prepare_file_upload().
             file_path: Local path to the file.
         """
-        url = upload_info["url"]
-        upload_params = upload_info.get("uploadParams", {})
-
-        # Build ordered fields: uploadParams first, file last (S3 requires file last)
-        form_fields: list[tuple[str, Any]] = [(k, (None, str(v))) for k, v in upload_params.items()]
         p = Path(file_path)
-        with p.open("rb") as fh:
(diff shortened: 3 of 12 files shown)


padak (commented):
# Detailed Review of PR #100

Hey Max, thank you so much for this PR! Adding storage write operations is a great addition -- `create-bucket`, `create-table`, and `upload-table` are the natural complement to the existing delete commands, and the 3-layer architecture is followed cleanly. The test coverage (688 lines!) is solid, and the Windows fix in `config_store.py` is a nice bonus catch. Really appreciate the work here.

That said, I did a thorough comparison with the official Keboola CLI ([keboola-as-code](https://github.com/keboola/keboola-as-code)) to see how Keboola Engineering implements the same operations, and found several issues that need addressing before we can merge. I've numbered them so we can track resolution.

---

## 1. CRITICAL: `upload-table` uses sync endpoint -- must switch to async import

**The problem:**
Your `client.py` uploads CSV via direct multipart POST to `/v2/storage/tables/{id}/import` (synchronous endpoint). This endpoint has a **hard 100 MB file size limit**. For anything larger, the API returns an error.

**How Keboola Engineering does it (keboola-as-code):**
The official CLI uses a **three-step async flow** -- see [`file/upload/operation.go`](https://github.com/keboola/keboola-as-code/blob/main/pkg/lib/operation/project/remote/file/upload/operation.go) and [`table/import/operation.go`](https://github.com/keboola/keboola-as-code/blob/main/pkg/lib/operation/project/remote/table/import/operation.go):

1. **Create file resource**: `POST /v2/storage/files/prepare` -- returns presigned URL (S3/GCS/Azure) and `fileId`
2. **Upload bytes** to cloud storage via the presigned URL (handles files up to 5 GB)
3. **Trigger import**: `POST /v2/storage/tables/{id}/import-async` with `{"fileId": "..."}` -- returns a Storage Job
4. **Poll job**: `GET /v2/storage/jobs/{jobId}` until status is `success` or `error` (KBC CLI uses 10-minute timeout)

**What to change:**
- Implement the file-first upload flow in `client.py`:
  - Add `create_file_resource(branch_id, name, ...)` method → `POST /v2/storage/files/prepare`
  - Upload the CSV to the presigned URL (use `httpx` to PUT the file)
  - Add `import_table_async(table_id, file_id, ...)` method → `POST /v2/storage/tables/{id}/import-async`
  - Reuse existing `_wait_for_storage_job()` for polling
- The `upload_table()` client method should orchestrate all three steps
- Consider showing upload progress in human mode (KBC CLI shows a progress bar)

Ref: Keboola docs confirm sync endpoint is 100 MB max, async supports up to 5 GB: https://help.keboola.com/storage/tables/

---

## 2. CRITICAL: Missing permission engine registration

The firewall-style permission system (`permissions.py`) doesn't know about the new commands. Currently registered:

```python
"storage.delete-table": "destructive",
"storage.delete-bucket": "destructive",
```

Missing:
```python
"storage.create-bucket": "write",
"storage.create-table": "write",
"storage.upload-table": "write",
```

**Why this matters:** Without registration, an agent running with a restrictive policy (e.g., `--mode deny --allow 'cli:read'`) can still call these write operations -- the permission check passes silently for unregistered operations. This is a security gap.

**What to change:** Add the three entries to the `OPERATION_CATEGORIES` dict in `src/keboola_agent_cli/permissions.py`.

---

## 3. CRITICAL: `kbagent context` must include new commands

The `AGENT_CONTEXT` string in `src/keboola_agent_cli/commands/context.py` is the primary reference for AI agents. The `### Storage` section doesn't mention the three new commands. An agent relying on `kbagent context` (which is the standard bootstrapping step) won't know these commands exist.

**What to change:** Add to the `### Storage` section in `AGENT_CONTEXT`:

```
  kbagent storage create-bucket --project NAME --stage STAGE --name BUCKET_NAME [--description D] [--backend B]
    Create a new storage bucket. Stage: "in" or "out".

  kbagent storage create-table --project NAME --bucket-id BUCKET_ID --name TABLE_NAME --column col:TYPE [...] [--primary-key COL]
    Create typed table. Types: STRING, INTEGER, NUMERIC, FLOAT, BOOLEAN, DATE, TIMESTAMP. Column type defaults to STRING if omitted.

  kbagent storage upload-table --project NAME --table-id TABLE_ID --file PATH [--incremental] [--delimiter D] [--enclosure E]
    Upload CSV into existing table. Full load by default, --incremental to append.
```

---

## 4. CRITICAL: SKILL.md decision table must include new commands

The `<!-- BEGIN AUTO-GENERATED COMMANDS -->` table in `plugins/kbagent/skills/kbagent/SKILL.md` is missing the new commands. Per our convention (see main CLAUDE.md): *"Adding/removing/renaming CLI commands → update SKILL.md (decision table, workflows)"*.

**What to change:** Add rows to the decision table:

```
| Create a new storage bucket | `kbagent storage create-bucket --project PROJECT --stage STAGE --name NAME` |
| Create a typed storage table | `kbagent storage create-table --project PROJECT --bucket-id BUCKET-ID --name NAME --column col:TYPE` |
| Upload a CSV file into a storage table | `kbagent storage upload-table --project PROJECT --table-id TABLE-ID --file FILE` |
```

---

## 5. MAJOR: `create-table` must validate column types

Currently, `--column foo:BANANA` passes through to the API where it fails with an opaque error. The help text promises 7 types but doesn't enforce them.

**Valid base types** (from [Keboola docs](https://help.keboola.com/storage/tables/data-types/)): `STRING`, `INTEGER`, `NUMERIC`, `FLOAT`, `BOOLEAN`, `DATE`, `TIMESTAMP`

Note: The API also accepts **native types** (e.g., Snowflake's `VARCHAR`, `DECIMAL(20,2)`, `TIMESTAMP_NTZ`), but for our CLI, validating against base types is the safe default since they're backend-agnostic.

**What to change in `storage_service.py`:**

```python
VALID_BASE_TYPES = {"STRING", "INTEGER", "NUMERIC", "FLOAT", "BOOLEAN", "DATE", "TIMESTAMP"}

for col_spec in columns:
    if ":" in col_spec:
        col_name, col_type = col_spec.split(":", 1)
        col_type = col_type.upper()
    else:
        col_name, col_type = col_spec, "STRING"
    if col_type not in VALID_BASE_TYPES:
        raise ValueError(
            f"Invalid column type '{col_type}' for column '{col_name}'. "
            f"Valid types: {', '.join(sorted(VALID_BASE_TYPES))}"
        )
    parsed_columns.append({"name": col_name, "definition": {"type": col_type}})
```

The `ValueError` should be caught in the command layer and mapped to exit code 2 (usage error). You could also consider putting `VALID_BASE_TYPES` in `constants.py` for reuse.

If we ever want to support native types, we can add a `--native-type` flag later. For now, base types keep it simple and cross-backend safe.

For reference: the official KBC CLI ([dialog.go](https://github.com/keboola/keboola-as-code/blob/main/internal/pkg/service/cli/cmd/remote/table/create/dialog.go)) uses the Go SDK's typed `keboola.Column` struct for this. We may also want to consider adding the [sapi-python-client](https://github.com/keboola/sapi-python-client) as a dependency eventually if we need deeper API integration -- but for now, a constant set is fine.

---

## 6. MAJOR: `create-bucket` must validate stage

Same pattern as #5 -- `--stage foo` passes to API with no local validation. The only valid values are `"in"` and `"out"`.

**What to change:** Easiest fix is in the Typer option definition in `commands/storage.py`:

```python
stage: str = typer.Option(
    ...,
    "--stage",
    help="Bucket stage: 'in' or 'out'",
    click_type=click.Choice(["in", "out"], case_sensitive=False),
)
```

Or validate in the service. Either way, fail fast with a clear message.

---

## 7. MAJOR: `upload_table` docstring says "async" but implementation is sync

```python
def upload_table(self, ...):
    """Upload a CSV file into an existing table (async, waits for completion)."""
```

Once you implement the async flow from item #1, this docstring will become correct -- but make sure it matches the final implementation. If the method orchestrates file-prepare → upload → import-async → poll, then "async, waits for completion" is accurate. Just double-check after the refactor.

---

## 8. MINOR: Auto-create bucket + table on upload (better UX)

The official KBC CLI auto-creates the bucket and table if they don't exist during `kbc remote table upload` (see [`EnsureBucketExists()`](https://github.com/keboola/keboola-as-code/blob/main/pkg/lib/operation/project/remote/table/import/operation.go#L104) and `checkTableExists()`). Our CLI currently requires three separate commands.

**Suggestion:** In `upload-table`, if the table doesn't exist:
1. Check if the bucket exists → if not, create it
2. Infer columns from CSV header (first row) → create the table
3. Proceed with upload

This would make `kbagent storage upload-table --project p --table-id in.c-mybucket.users --file data.csv` a one-shot command. The KBC CLI even reads the CSV header to auto-detect column names when no `--columns` flag is provided ([`extractColumnsFromCsv()`](https://github.com/keboola/keboola-as-code/blob/main/pkg/lib/operation/project/remote/table/import/operation.go#L151)).

If you implement this, add `--no-auto-create` flag to disable the behavior (for strict mode). And log what's being auto-created so the user knows.

---

## 9. MINOR: Show file size info on upload

The official KBC CLI shows a progress bar with file size during upload. Our CLI currently gives no feedback about file size -- the user doesn't know if they're uploading 1 KB or 99 MB.

**What to change:**
- In human mode: print file size before upload starts (e.g., `Uploading data.csv (4.2 MB) to in.c-mybucket.users...`)
- In JSON mode: include `file_size_bytes` in the response dict
- If you implement the async flow, consider a simple progress indicator (even just `Uploading... done.` is better than silence)

---

## 10. MINOR: Use `pathlib.Path` instead of `os.path`

The project convention (per CLAUDE.md) is to prefer `pathlib.Path` over `os.path`. The PR uses `os.path.basename()` and `os.path.isfile()`.

**What to change:**
- `client.py`: `os.path.basename(file_path)` → `Path(file_path).name`
- `commands/storage.py`: `os.path.isfile(file)` → `Path(file).is_file()`
- Remove `import os` where it's no longer needed

---

## 12. NIT: Future commits -- no `Co-Authored-By`

The commit `3204686` includes `Co-Authored-By: Claude Sonnet 4.6`. Per our project conventions, we don't include AI attribution in commits or PR descriptions. For future commits in this PR, please omit the `Co-Authored-By` line.

For reference on our commit/PR conventions, take a look at: https://github.com/padak/claude-code-kit/blob/main/CLAUDE.md

---

## Bonus suggestion: Full file upload support?

Since you'll be implementing the File API flow for `upload-table` anyway (item #1), would you also want to add explicit `kbagent storage file-upload` and `kbagent storage file-download` commands? The KBC CLI has `kbc remote file upload` and `kbc remote file download` as standalone commands. This would give users the ability to manage Storage Files directly (useful for large dataset staging, backup/restore, etc.).

Just a thought for a follow-up PR if you're interested -- not blocking this one.

---

## Summary

| # | Severity | Action needed |
|---|----------|---------------|
| 1 | CRITICAL | Switch upload to async file-first flow |
| 2 | CRITICAL | Register 3 ops in permission engine |
| 3 | CRITICAL | Update `kbagent context` output |
| 4 | CRITICAL | Update SKILL.md decision table |
| 5 | MAJOR | Validate column types against known base types |
| 6 | MAJOR | Validate bucket stage (in/out) |
| 7 | MAJOR | Fix docstring after async refactor |
| 8 | MINOR | Auto-create bucket/table on upload |
| 9 | MINOR | Show file size info |
| 10 | MINOR | Use pathlib.Path |
| 12 | NIT | No Co-Authored-By in future commits |

Thanks again for the solid foundation here -- the architecture, tests, and overall approach are great. Once these items are addressed, this will be a really strong addition to the CLI!

pyproject.toml
```diff
@@ -42,6 +42,7 @@ markers = [
 target-version = "py312"
 line-length = 100
 src = ["src", "tests"]
+exclude = ["tests/_test_upload_client_append.py"]
```
padak: One small thing: this adds an exclude for `tests/_test_upload_client_append.py` but that file doesn't appear anywhere in the PR. Is this a local development artifact that got committed by accident? If the file doesn't exist in the repo, the exclude line should be removed.

padak (approved):
Great work, Max! All 4 critical and 3 major items from the review are properly addressed. The async file-first upload flow is clean, permissions are registered, context + SKILL.md are updated, and input validation catches bad types/stages early.

I especially appreciate the extras you added beyond what was requested:
- GCP bearer-token upload path (`gcsUploadParams`) -- shows good understanding of Keboola's multi-cloud architecture
- Boolean serialization fix in `prepare_file_upload` (omitting `isPermanent`/`isPublic` to avoid Python `False` -> `"False"` in form-data)
- UTF-8 BOM handling in `_read_csv_header`
- The Windows `config_store.py` fix

Test coverage is solid -- 975 lines in `test_storage_write.py` plus 313 lines of client-level tests for the async upload flow.

One minor thing flagged inline on `pyproject.toml` (the ruff exclude for a file that doesn't exist in the PR). Please address that and this is ready to merge.

ottomansky:
fixed

Issue #102, opened 2026-04-10
docs: clarify org setup as primary onboarding path (vs. single-project token)


AndreHeller:
## Problem

The current README and quick-start guide lead with `kbagent project add --token YOUR_TOKEN`, which requires the user to manually look up and copy a Storage API token for each project. In practice, most users who manage multiple Keboola projects hit a dead end here — the token they try first is often invalid or wrong scope.

The more natural path for most users is `kbagent org setup`, which creates tokens automatically. But this command is buried after the single-project example.

## Suggestion

Restructure the quick-start section to present **two explicit variants** upfront, so users can pick their path immediately:

**Variant A – Org admin** (requires Manage API token):
```bash
KBC_MANAGE_API_TOKEN=xxx kbagent org setup --org-id 123 --url https://connection.keboola.com --yes
```
→ Automatically discovers and registers all projects in the organization.

**Variant B – Project member** (requires Personal Access Token):
```bash
KBC_MANAGE_API_TOKEN=<personal_access_token> kbagent org setup --project-ids 123,456 --url https://connection.keboola.com --yes
```
→ Works without org-admin rights; PAT is available in Keboola UI under Account Settings → Personal Access Tokens.

The single `project add --token` path can stay as an advanced/manual option, but should not be the first thing a new user sees.

## Why

- `org setup` is safer (tokens are created with minimal scope, named consistently)
- `org setup` is faster for multi-project setups
- New users don't know where to find a Storage API token, but they can easily generate a PAT

Reported based on real onboarding experience.

PR #103, opened 2026-04-10
docs: clarify /plugin commands require Claude Code interactive terminal mode

AndreHeller:
## Problem

The current README presents `/plugin` commands in a `bash` code block, which misleads users into running them in a regular shell. They only work inside Claude Code's **interactive terminal mode** (`claude` in your terminal) -- not in VS Code extension chat, not in the desktop app, and definitely not in bash.

Related to: #102

## Changes

- Moved `/plugin` snippet out of the `bash` fenced block (it's not a shell command)
- Added a clear note explaining where these commands work and where they don't
- Added a note that `kbagent context` is the fallback for users on VS Code / desktop app

## Before / After

**Before:**
```bash
/plugin marketplace add padak/keboola_agent_cli
/plugin install kbagent@keboola-agent-cli
```
(Looked like a bash command -- multiple users tried running it in a terminal and got `bash: /plugin: No such file or directory`)

**After:** Explicit warning listing the three environments where it does NOT work, plus guidance for VS Code/desktop users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,14 +17,27 @@ KBC_MANAGE_API_TOKEN=xxx kbagent org setup --org-id 123 --url https://connection
 
 ## For AI agents (Claude Code, Codex, Gemini)
 
-Install the Claude Code plugin and the agent learns all commands automatically:
+Install the Claude Code plugin and the agent learns all commands automatically.
 
-```bash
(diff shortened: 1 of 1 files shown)


padak:
Thanks for the fix! Merged.

PR #190, opened 2026-04-20
feat(0.21.0): sync push deploys config rows + first-class variables UX

ottomansky:
## Summary

Unlocks two workflows that previously required dropping to `kbc` or direct API calls:

**Commit 1 — `sync push` deploys config rows (+ row encryption).** Pull records per-row hashes (`pull_hash` + `pull_config_hash`) in the manifest; diff performs a 3-way compare against local/base/remote; push dispatches create/update via the existing `create_config_row` / `update_config_row` client methods. `#`-prefixed secret values in rows flow through the Encryption API with fail-closed semantics (`ENCRYPTION_FAILED`, exit non-zero) matching parent-config behavior. Manifest bumps v2→v3 with backward-compat load via `extra="allow"`. `keboola.variables` / `keboola.shared-code` row payloads hoist `values` / `code_content` to top-level YAML for byte-for-byte round-trip.

**Commit 2 — `kbagent config variables-{set,get,clear}`.** Agents manage variable values on any config without treating `keboola.variables` as a component. `variables-set` auto-creates a sibling `keboola.variables` config + default row on first use and patches the parent's `runtime.variables_id` / `variables_values_id` linkage; subsequent calls merge by default or replace under `--replace`. Secret values encrypt via the Encryption API (same fail-closed contract). Dual output (`--json` / human), `--dry-run`, `--variables-id` / `--values-id` to attach to an existing values row, `--allow-plaintext-on-encrypt-failure` bootstrap escape hatch, `--yes` to skip confirmation.

Version → `0.21.0`. Changelog entry, SKILL.md decision-table rows, `references/variables-workflow.md`, and updated `references/gotchas.md` all land together.

### Architecture

- 3-layer discipline preserved: `client.py` (existing row endpoints get dedicated test class) → `services/` (new `variables_service.py`, `_encryption.py` extracted so `sync_service` and `variables_service` share the encryption contract without cross-service imports) → `commands/config.py` (thin Typer wrappers).
- Zero existing code paths rewritten. Row handling is additive in `sync_service.py`; config-only users see identical behavior. The one unavoidable tweak — `newline=""` in `_write_config_file` — fixes a pre-existing Windows `\n`→`\r\n` hash-drift that was masking 4 test failures.
- `permissions.py`: `config.variables-set = write`, `config.variables-get = read`, `config.variables-clear = destructive`.

## Test plan

- [x] Client layer: `TestConfigRowMethods` (9 tests) — URL/body/branch routing + error paths for create/update/delete.
- [x] Service layer — sync row push: `TestPushRows` (5 tests) incl. `test_push_encryption_failure_aborts_fail_closed`.
- [x] Service layer — variables: 14 tests (get 3 / auto-create 2 / update 4 / encryption 2 / validation 1 / clear 2) with `client.close.assert_called_once()` on every error path.
- [x] CLI layer — sync push rows: `TestSyncPushCli` +3 tests (json, human, encryption-failure).
- [x] CLI layer — variables: 12 tests covering both output modes, `ENCRYPTION_FAILED`, `INVALID_ARGUMENT`, `CONFIG_ERROR` → exit 5, `--dry-run` non-invocation contract.
- [x] Round-trip: `TestVariablesRowRoundTrip` (5 tests) proves byte-for-byte YAML round-trip for `keboola.variables` + `keboola.shared-code`.
- [x] Manifest v2→v3: backward-compat load + row `metadata` field.
- [x] E2E vs live Keboola project 1143:
  - `TestE2ESyncWorkflow::test_sync_push_variable_row_round_trip` ✅
  - `TestE2ESyncWorkflow::test_config_variables_round_trip` ✅ (auto-create → get → merge → replace → clear → verify-unlinked, with cleanup)
- [x] `uv run ruff check src/ tests/` — clean.
- [x] `uv run ruff format --check src/ tests/` — 157 files formatted.
- [x] Targeted suite post-review: 174 / 174 passing.

### Gaps vs plan — now closed (commit 3)

1. **`_push_delete_row`** — extracted as a named helper (symmetric with `_push_create_row` / `_push_update_row`). Behaviour was already wired inline; this gives tests a clean call-site.
2. **`_find_untracked_rows`** — paralleling `_find_untracked_configs`. Dropping a `rows/<name>/_config.yml` under a tracked config now surfaces it as an `added` row change; push POSTs it via `create_config_row`, writes the API-assigned id back into the manifest, and records pull hashes. Covered by `test_push_untracked_row_dir_calls_create_config_row`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,7 @@ ENV/
 
 # Environment variables
 .env
-.env.local
+.env.*
 
 # IDE
 .idea/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -250,6 +250,9 @@ kbagent config detail --project NAME --component-id ID --config-id ID [--branch 
 kbagent config search --query PATTERN [--project NAME] [--component-type TYPE] [--ignore-case] [--regex] [--branch ID]
 kbagent config update --project NAME --component-id ID --config-id ID [--name N] [--description D] [--configuration JSON|@file|-] [--configuration-file PATH] [--set PATH=VALUE ...] [--merge] [--dry-run] [--branch ID]
 kbagent config rename --project NAME --component-id ID --config-id ID --name "New Name" [--branch ID] [--directory DIR]
+kbagent config variables-set --project NAME --component-id ID --config-id ID --var KEY=VALUE [--var ...] [--replace] [--variables-id ID] [--values-id ID] [--branch ID] [--dry-run]
+kbagent config variables-get --project NAME --component-id ID --config-id ID [--branch ID]
+kbagent config variables-clear --project NAME --component-id ID --config-id ID [--branch ID] [--yes]
 
 kbagent job list [--project NAME] [--component-id ID] [--status STATUS] [--limit N]
 kbagent job detail --project NAME --job-id ID
--- a/plugins/kbagent/.claude-plugin/plugin.json
+++ b/plugins/kbagent/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kbagent",
-  "version": "0.20.6",
+  "version": "0.21.0",
   "description": "AI-friendly interface to Keboola Connection projects — explore configs, jobs, lineage, call MCP tools, manage dev branches, and debug SQL in workspaces",
   "author": {
     "name": "Keboola",
--- a/plugins/kbagent/skills/kbagent/SKILL.md
+++ b/plugins/kbagent/skills/kbagent/SKILL.md
@@ -207,6 +207,7 @@ For detailed response parsing rules and common pitfalls, see [gotchas](reference
 | Dev branches | [branch-workflow](references/branch-workflow.md) |
 | Encrypting secrets for MCP tools | [encrypt-workflow](references/encrypt-workflow.md) |
 | Sync & Git-branching (GitOps) | [sync-workflow](references/sync-workflow.md) |
+| Sync row-level internals (manifest v3, hoist, encryption) | [sync-rows-workflow](references/sync-rows-workflow.md) |
 | **Variables (attach to any config)** | [variables-workflow](references/variables-workflow.md) |
 | Reading synced data | [reading-synced-data](references/reading-synced-data.md) |
 | SQL migration (input mapping removal) | [sql-migration-workflow](references/sql-migration-workflow.md) |
--- a/plugins/kbagent/skills/kbagent/SKILL.md
+++ b/plugins/kbagent/skills/kbagent/SKILL.md
@@ -93,9 +93,9 @@ When working inside a git repository or project directory, run `kbagent init` (o
 | Rename a configuration (update name via API + rename local sync directory) | `kbagent config rename --project PROJECT --component-id COMPONENT-ID --config-id CONFIG-ID --name NAME` |
 | Delete a configuration from a project | `kbagent config delete --project PROJECT --component-id COMPONENT-ID --config-id CONFIG-ID` |
 | Generate boilerplate configuration files for a Keboola component | `kbagent config new --component-id COMPONENT-ID` |
-| Assign variables to a config (auto-creates backing keboola.variables on first call) | `kbagent config variables-set --project PROJECT --component-id COMPONENT-ID --config-id CONFIG-ID --var KEY=VALUE` |
-| Read variable values attached to a config | `kbagent config variables-get --project PROJECT --component-id COMPONENT-ID --config-id CONFIG-ID` |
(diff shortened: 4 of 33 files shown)


padak:
Hi @ottomansky — tested this end-to-end in `/tmp/apify` against project `padak` (901). Full test suite passes locally (1815/1815) and the happy paths I threw at the new commands all behaved as advertised. One finding worth addressing before merge, plus a small cosmetic note.

## Finding: asymmetric encryption contract between `variables-set` and `sync push`

**`variables-set` encrypts `#`-prefixed values correctly. `sync push` on `keboola.variables` rows does not — plaintext lands on Storage.**

### Reproduction

1. `sync pull` a project that has a `keboola.variables` row.
2. Edit `rows/<name>/_config.yml` and append:
   ```yaml
   values:
   - name: '#api_key'
     value: 'my-plaintext-secret-to-encrypt'
   ```
3. `sync push --project <alias>` → exit 0, row updated.
4. `config detail` against `keboola.variables/<id>` shows the value stored **verbatim as plaintext** — no `KBC::ProjectSecure::` prefix, no Encryption API roundtrip.

I reproduced this against `keboola.python-transformation-v2/1221677380` in project 901; the raw plaintext `my-plaintext-secret-to-encrypt` was visible in the row config response. (I've since cleaned that up.)

Compare with `variables-set --var '#api_key=…'` against the same row: that one correctly yields `KBC::ProjectSecure::eJw…` on the server. So both paths *should* converge on the same ciphertext, but today only one of them does.

### Root cause

`services/_encryption.py::collect_secrets` walks for `#`-prefixed **dict keys**. That matches the canonical shape `{\"#api_token\": \"...\"}`. But `keboola.variables` stores entries as `{values: [{name: \"#api_key\", value: \"plaintext\"}]}` — the `#` is the **value** of a `name` field, not a dict key — so the walker skips it.

You already handle this exact asymmetry in `variables_service._encrypt_variable_values` (the \"flat dict → encrypt → rebuild list\" trick), and the docstring there calls out why: *\"that shape is invisible to encrypt_secrets_in_config, which scans for #-prefixed keys.\"* That invisibility is still present in `sync_service._push_create_row` / `_push_update_row`, which call `encrypt_secrets_in_config(configuration, …)` directly on the hoisted `{values: [...]}` body without the pre-flatten step.

### Suggested fixes (pick whichever feels lighter)

**Option A — teach `encrypt_secrets_in_config` about row-hoisted shapes.** Before recursing, check if `obj` is `{\"values\": [{\"name\": \"#x\", \"value\": \"...\"}, ...]}` and pivot it into a flat dict for that one level. Keeps the \"one encryption contract, one function\" invariant.

**Option B — apply the same flatten-then-encrypt dance in `_push_*_row` when `component_id in ROW_HOIST_COMPONENTS`.** Mirror `variables_service._encrypt_variable_values` exactly. Slightly more duplicated surface but keeps `_encryption.py` shape-agnostic.

I'd lean toward A since the fail-closed contract is advertised at the helper level — it's easy for a future row-hoist component to repeat this mistake if the helper itself doesn't cover the shape.

### Test coverage gap

`TestPushRows::test_push_encryption_failure_aborts_fail_closed` checks that encryption **failure** aborts the push, but I don't see a test asserting that a `#`-prefixed value inside `values: [...]` actually hits `encrypt_values`. A test like:

```python
# local row:
# values:
#   - name: '#api_key'
#     value: 'plaintext'
# After push:
assert client.encrypt_values.called  # currently: not called
```

would have caught this, and pairs nicely with whichever fix you pick.

## Minor: dry-run `--replace` leaks full ciphertext

In `variables-set --replace --dry-run`, **unchanged** (`=`) rows mask secrets as `<encrypted>`, but **dropped** (`-`) rows print the full `KBC::ProjectSecure::...` ciphertext. Not a security issue (ciphertext without the project key is inert), but inconsistent with the same output's masking policy two lines up. One-line fix in the dry-run formatter.

## What works well

- Manifest v3 backward-compat load + `ManifestConfigRow.metadata` with `pull_hash` + `pull_config_hash` — row-level diffs are pleasantly precise (`values[0].value changed: '123' -> 'pr190-test'`).
- Row hoisting for `keboola.variables` / `keboola.shared-code` produces editable YAML that round-trips clean — manifest hashes match after pull→push→diff with no drift.
- `variables-set` auto-create path: one CLI call creates the sibling config + schema + default row + parent PUT, exactly as `references/variables-workflow.md` claims. Description auto-populated with parent reference, schema includes all keys (`#`-prefixed ones too).
- `variables-clear` unlinks parent without deleting backing config — correct design for shared variables configs.
- `--dry-run` output is readable in both human and JSON modes; `<encrypted>` masking on unchanged rows is a nice touch.
- Edge cases all handled: missing `--var`, malformed `KEY=VALUE`, nonexistent parent config, already-encrypted values (no double-encrypt).

Overall really solid shape — the 3-layer separation stayed clean, `_encryption.py` as a shared helper was the right call, and the manifest migration is thought-through. Just the one encryption gap before this ships.

PR #199, opened 2026-04-22
fix(0.21.2): job run banner reads resolved id, hint threads branch_id, whitespace strip

ottomansky:
## Summary

- **Banner fix**: `kbagent job run` (rich mode) now prints "Bound variable values row: `<id>`" after the job is created, reading `resolvedVariableValuesId` from the service response instead of echoing the raw `--variable-values-id` flag. This means the banner fires even when the row was auto-resolved (no flag supplied), and is silent when resolution was skipped.
- **Whitespace strip**: `--variable-values-id "  row-1  "` is stripped of leading/trailing whitespace before the empty-string guard, so a padded value no longer slips past the guard and reaches the service unstripped.
- **Hint branch threading**: `--hint client job run --branch <id>` now passes `branch_id` to all three client calls (`get_config_detail`, `list_config_rows`, `create_job`). Previously the arg was silently dropped, causing the rendered snippet to target the production branch even when a dev branch was specified.
- **Import hoist**: `from rich.markup import escape` moved to module level in `commands/job.py`.

## Resolution order / Behavior

| Scenario | Before | After |
|---|---|---|
| Auto-resolved row (no `--variable-values-id`) | No banner shown | Banner shows resolved row id |
| Explicit `--variable-values-id "  row  "` | Padded string reaches service | Stripped `"row"` reaches service |
| `--no-variables` | No banner (correct) | No banner (unchanged) |
| `--hint client job run --branch 42` | `branch_id` absent from hint | `branch_id=42` in all 3 calls |

## Architecture

All changes are confined to the CLI command layer and hint definitions — no service or client logic was modified except a docstring cleanup on `resolve_variable_values_id` in `job_service.py`. The service already stamped `resolvedVariableValuesId` onto the result dict correctly (PR #191); this PR only makes the command layer consume it correctly.

## Test plan

- [x] `TestJobRun::test_job_run_rich_mode_echoes_resolved_values_id` — banner fires from response, not from flag
- [x] `TestJobRun::test_job_run_strips_whitespace_around_variable_values_id` — stripped value reaches service
- [x] `TestJobRun::test_job_run_no_variables_flag_forwarded` — JSON output does NOT contain `resolvedVariableValuesId` when skipped
- [x] `TestJobServiceVariableValuesResolution::test_run_job_wait_preserves_resolved_variable_values_id` — stamp survives `job = wait_for_queue_job(...)` dict replacement
- [x] `TestJobServiceVariableValuesResolution::test_run_job_no_variables_skips_resolution` — `resolvedVariableValuesId` absent from result
- [x] `TestClientRenderer::test_job_run_hint_threads_branch_id_through_all_steps` — `branch_id=42` appears ≥3 times in rendered hint (get_config_detail + list_config_rows + create_job)

### Live E2E

`TestE2EJobRunVariableValues`: **6/6 passed** (14.8 s against `connection.europe-west3.gcp.keboola.com`):
- `test_resolve_variable_values_id_live` PASSED
- `test_job_run_surfaces_resolved_variable_values_id` PASSED
- `test_job_run_explicit_override_wins_over_resolver` PASSED
- `test_job_run_no_variables_skips_resolution` PASSED
- `test_job_run_no_variable_rows_surfaces_error_code` PASSED
- `test_resolver_prefers_explicit_values_id_over_first_row` PASSED

Full suite: 16 passed, 2 pre-existing failures (`TestFullE2E` + `TestE2EJsonConsistency` — both fail on `lineage show --load` missing, identical on clean `main`).

## Pre-PR review loop

Three parallel Explore agents reviewed: wire-contract, silent-failure, test-coverage.

**Wire-contract:** All paths pass. Agent #1 initially flagged `if resolved_values_id:` as a `"0"`-falsy trap; agent #2 independently confirmed Python's `bool("0") == True` — no bug. None-omission in hint renderer confirmed correct (branch_id omitted when `{branch}` resolves to `None`).

**Silent-failure:** All pass. Whitespace strip + empty-string guard ordering confirmed correct; `no_variables` mutual-exclusion check uses the post-strip value; `escape` import raises no circular dependency.

**Coverage gaps found and fixed:**
1. `test_job_run_no_variables_flag_forwarded` was only asserting service kwargs, not the JSON output — added parse assertion that `resolvedVariableValuesId` is absent from the `data` dict.
2. No service-level test verified that `resolvedVariableValuesId` survives the `job = wait_for_queue_job(...)` dict replacement — added `test_run_job_wait_preserves_resolved_variable_values_id`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,8 +18,6 @@ ENV/
 # Environment variables
 .env
 .env.*
-!.env.example
-!.env.template
 
 # IDE
 .idea/
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@ ENV/
 # Environment variables
 .env
 .env.*
+!.env.example
+!.env.template
 
 # IDE
 .idea/
--- a/plugins/kbagent/.claude-plugin/plugin.json
+++ b/plugins/kbagent/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kbagent",
-  "version": "0.21.1",
+  "version": "0.21.2",
   "description": "AI-friendly interface to Keboola Connection projects — explore configs, jobs, lineage, call MCP tools, manage dev branches, and debug SQL in workspaces",
   "author": {
     "name": "Keboola",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "keboola-agent-cli"
-version = "0.21.1"
+version = "0.21.2"
 description = "AI-friendly CLI for managing Keboola projects"
 readme = "README.md"
 requires-python = ">=3.12"
--- a/src/keboola_agent_cli/changelog.py
+++ b/src/keboola_agent_cli/changelog.py
@@ -8,6 +8,12 @@ from __future__ import annotations
 
 # Ordered newest-first.  Each value is a list of brief one-line descriptions.
 CHANGELOG: dict[str, list[str]] = {
+    "0.21.2": [
+        "Fix: `kbagent job run` rich-mode banner now reads `resolvedVariableValuesId` from the service response instead of echoing the raw `--variable-values-id` flag -- shows the auto-resolved row even when the flag was omitted",
+        "Fix: `--variable-values-id` value is stripped of surrounding whitespace before reaching the service -- prevents a padded input from bypassing the empty-string guard",
+        "Fix: `--hint client job run --branch ID` now threads `branch_id` through all three client calls (get_config_detail, list_config_rows, create_job) -- previously the branch arg was silently dropped, causing the hint to target production",
+        "Chore: `rich.markup.escape` import hoisted to module level in commands/job.py",
+    ],
     "0.21.1": [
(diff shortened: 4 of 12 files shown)


padak (dismissed):

padak (approved):
## Approve — merge-ready

Clean bugfix, no blockers. All 50 related tests pass, ruff OK, version sync OK, conventional commits without AI attribution.

### Nice-to-have (non-blocking)
- The `.gitignore` change (lines 20-21, `!.env.example`, `!.env.template`) is out of scope for a `job run banner` bugfix. Consider a separate micro-PR `chore: whitelist .env templates`.
- The PR description phrase "branch arg was silently dropped" is mildly overstated — the renderer already injected `branch_id` through a fuzzy fallback. A more accurate framing would be "relied on renderer fallback; now explicitly threaded for robustness".

Thanks!

ottomansky:
Dropped the out-of-scope `!.env.example` / `!.env.template` whitelist from `.gitignore` per review — will open a separate `chore: whitelist .env templates` micro-PR for those.

Commit: `902bbf3`

`make lint format-check skill-check version-check changelog-check` all green.

PR #236, opened 2026-05-01
feat(0.26.1): project member & invitation lifecycle

ottomansky:
## Summary

Closes the long-standing Manage API gap that forced every Keboola-internal automation (most recently `17_CuestaDemo/scripts/replicate_master.py:173` and `invite_participants.py:19`) to bypass kbagent and POST raw HTTP at `/manage/projects/{id}/invitations`. Adds a full member/invitation lifecycle to `kbagent project`, validated end-to-end against the real Manage API on `connection.us-east4.gcp.keboola.com`.

Patch bump (0.26.0 → 0.26.1). Additive surface inside an existing command group, no breaking changes.

### New commands

```
kbagent project invite --project ALIAS --email EMAIL --role admin|guest|readOnly|share [--reason TEXT] [--dry-run]
kbagent project invite --from-csv FILE [--default-role ROLE] [--workers N] [--dry-run]
kbagent project member-list --project ALIAS [--include-pending]
kbagent project invitation-list --project ALIAS
kbagent project invitation-cancel --project ALIAS --email EMAIL [--invitation-id ID] [--yes]
kbagent project member-remove --project ALIAS --email EMAIL [--yes]    # destructive
kbagent project member-set-role --project ALIAS --email EMAIL --role admin|guest|readOnly|share
```

All seven authenticate via `KBC_MANAGE_API_TOKEN` (Manage API), not the project's Storage token.

### Architecture (3-layer; respects `CONTRIBUTING.md` §boundaries)

| Layer | What changed |
|---|---|
| Client (`manage_client.py`) | 6 new methods. Role-change uses **PATCH** -- PUT returns 404 on this endpoint even on real members. |
| Service (new `services/member_service.py`) | Alias→(stack,project_id) resolution; email→user_id / invitation_id case-insensitive lookup; idempotent "already invited / already member" handling (HTTP 400 + msg-substring marker, NOT 422); parallel CSV bulk via `ThreadPoolExecutor` (default 8 workers); single-stack-URL invariant per CSV. |
| Commands (`commands/project.py`) | 7 thin Typer wrappers. `click.Choice` enforces the role whitelist `('admin','guest','readOnly','share')` -- lifted verbatim from the API's own validation error message. |

### Live validation against project 5725 (`[Cuesta training] - Master`)

| Step | Result |
|---|---|
| `member-list --project cuesta-master` | 3 active admins (max, mfiser, VFisa) |
| `invite --email ottomansky.max@gmail.com --role guest` | `status=ok, invitation_id=1742` |
| `invitation-list` | invitation appears |
| Re-invite same email | `status=noop, note=already_invited` (HTTP 400 normalized) |
| `invitation-cancel --email ottomansky.max@gmail.com --yes` | `status=cancelled` |
| `invitation-list` | empty (project back to 1 pre-existing pending invitation) |
| Bulk `--from-csv --dry-run` | 2 rows previewed, no API calls |
| `--hint client` / `--hint service` | both produce runnable Python |

### Verifications I performed against the live Manage API (2026-05-01)

| Endpoint | Verdict |
|---|---|
| `GET /users` | Plain `list[dict]`. Role at top level (not nested under `user`). |
| `GET /invitations` | Plain `list[dict]` with `user.{id,name,email}`. |
| `POST /invitations` happy path | 201 + invitation object. |
| `POST /invitations` already-invited | **HTTP 400** + `error: "This user has already been invited to this project."` (NOT 422 -- the heuristic from `invite_participants.py:25` was wrong about the code). |
| `POST /invitations` already-member | HTTP 400 + `error: "This user is already a member of this project."`. |
| `POST /invitations` invalid role | HTTP 400 + `error: "Role 'X' is not valid. Allowed roles are: admin, guest, readOnly, share"` -- API self-reports the whitelist. |
| `DELETE /invitations/{id}` | 204 success; 404 + "Invitation not found" if already deleted. |
| `DELETE /users/{userId}` | 204 success; 400 "Administrator not found" if not a member. |
| `PUT /users/{userId}` | **404 -- endpoint doesn't exist in this shape.** |
| `PATCH /users/{userId}` | **200 + updated user dict -- canonical role-change.** |

### Plugin synchronization map (every silent-drift surface walked)

| File | Change |
|---|---|
| `commands/context.py` (`AGENT_CONTEXT`) | New `### Project Members & Invitations (since v0.26.1)` block. |
| `CLAUDE.md` `## All CLI Commands` | 7 new signatures. |
| `plugins/kbagent/.claude-plugin/plugin.json` | `make version-sync` (0.26.1). |
| `plugins/kbagent/agents/keboola-expert.md` | Rule 6 VERSION GATE updated; 7 new matrix rows under "Project administration" (with `First choice / Fallback / NEVER`); 3 new inline gotchas (HTTP 400 noop, PATCH-not-PUT, parallel-bulk ordering). |
| `plugins/kbagent/skills/kbagent/SKILL.md` | 6 new description triggers; new `member-workflow` row in workflow links table; auto-table regenerated by `make skill-gen`. |
| `plugins/kbagent/skills/kbagent/references/commands-reference.md` | New "Project members & invitations (since v0.26.1)" section. |
| `plugins/kbagent/skills/kbagent/references/gotchas.md` | 3 new entries tagged `(since v0.26.1)`. |
| `plugins/kbagent/skills/kbagent/references/member-workflow.md` | **New** -- single + bulk + audit + role-change + remove golden paths + idempotency cheat sheet. |
| `permissions.py` `OPERATION_REGISTRY` | 6 new entries. `member-remove` is `destructive`; `member-list` / `invitation-list` are `read`; rest are `admin`. |

### Tests

| Layer | New | File |
|---|---:|---|
| Service | 23 | `tests/test_member_service.py` |
| CLI (CliRunner) | 15 | `tests/test_member_cli.py` |
| Client (httpx_mock) | 12 | `tests/test_manage_client.py` (extended) |
| E2E | 1 | `tests/test_e2e.py::test_project_invite_e2e` (marker `e2e_invite`, gated on `E2E_MANAGE_TOKEN` + `E2E_INVITE_PROJECT_ID`, opt-in via `make test-e2e-invite`) |

Full suite: **2410 tests pass**, 0 fail. `make check` clean (lint + format + skill freshness + version sync + changelog completeness + error-code enum + tests).

## Test plan

- [x] `make check` clean
- [x] `kbagent project invite/member-list/invitation-list/invitation-cancel/member-remove/member-set-role --help` reads like a man page
- [x] `--hint client` and `--hint service` both produce runnable Python
- [x] Live test against `cuesta-master` (project 5725 / us-east4): single invite + idempotency 400→noop + cancel + verify-gone + bulk dry-run
- [x] Token-leak grep across changed files: no real secrets
- [x] Master cuesta project verified back to its original state (3 members + 1 pre-existing pending invitation, untouched)
- [ ] `make test-e2e-invite` against the upstream/CI runner (gated on env vars; not run in default CI)
- [x] Self-review (kickoff §6 iteration 1) — 0 BLOCKING / 4 NON-BLOCKING / 3 NIT (see PR comment)
- [x] Independent agent review (kickoff §6 iteration 2+) — 0 BLOCKING / 7 NON-BLOCKING / 3 NIT (see PR comment)

## Out of scope (noted for a future PR)

- `POST /manage/projects/{id}/users` direct member-add (bypasses email confirmation; conflicts with `invite` semantics).
- Org-level user management (`/manage/organizations/{id}/users`).
- Re-send / resend-expired-invitation if the API exposes it as a distinct endpoint.


ottomansky:
## Self-review iteration 1 — 1 BLOCKING fixed

**BLOCKING (1, fixed)**
- `services/member_service.py:128-138` — `invite_bulk` raised `ConfigError` upfront when ANY row referenced an unregistered alias / project_id, aborting the whole batch and violating the documented "one bad row never aborts the rest" contract. The single ConfigError surfaced to the command layer as exit-5 with no per-row visibility, so a user with a 52-row CSV and one typo would see zero progress and have no idea which row failed.
  - Fix: per-row resolution accumulates `MemberInviteRow(status="failed", note=...)` upfront; the executor runs only on resolved rows. Multi-stack-URL invariant still raises (it's a security boundary, not partial-success). New regression test `test_unknown_alias_in_csv_row_is_per_row_failure_not_global_abort` covers this. Mirrors `OrgService.refresh_tokens` accumulation pattern.

**NON-BLOCKING (0)** / **NIT (0)** — no other material findings in iteration 1.

`make check` clean. **2411 tests pass** (+1 regression test). Force-pushed `e17c575 → 8b2db76`. CI re-running.

Iteration 2 (independent agent + §6.5 security checklist) coming next.

ottomansky:
## Iteration 2 — independent agent review (5 NON-BLOCKING fixed, 5 NIT deferred)

Independent agent verdict: **APPROVE FOR MERGE** — code clean, security clean, no BLOCKING. Of 6 NON-BLOCKING + 5 NIT findings, fixed the 5 material ones; documented why the rest are deferred.

### NON-BLOCKING (fixed)

| # | File:line | Finding | Fix |
|---|---|---|---|
| 1 | `models.py:194,202` (now `:198,217`) | `ProjectMember`/`ProjectInvitation` had `extra: "ignore"`, dropping audit-relevant fields (`created`, `expires`, `invitor`, `approver`, `creator`) from `--json` | Switched to `extra: "allow"`; full API payload passes through. Docstrings explain. |
| 2 | `member_service.py:128` | `--dry-run` short-circuited before the multi-stack-URL guard, so a CSV mixing stacks "previewed ok" then aborted on the real run | Multi-stack guard now also fires inside `_bulk_dry_run`. Regression test `test_dry_run_rejects_multi_stack_csv`. |
| 3 | `member_service.py:484` | CSV opened with `encoding="utf-8"`, so Excel's BOM made the first header read as `﻿email` and triggered "missing 'email' column" | `encoding="utf-8-sig"` strips the BOM. Regression test `test_csv_with_utf8_bom_parses`. |
| 5 | `commands/project.py:837` | `project invite --hint client --from-csv FILE` silently fell through to the live path and prompted for the manage token | Now exits 2 with a usage error explaining hints are single-shot only. Regression test `test_hint_with_from_csv_emits_clear_error_not_silent_skip`. |
| 6 | `member_service.py:399` | `_invite_one(alias=...)` was misleading -- bulk path passes the raw CSV cell (which may be a numeric `project_id` string) | Renamed to `project_label`; docstring clarifies single-shot vs bulk semantics. |

### NON-BLOCKING (deferred with rationale)

| # | Finding | Rationale |
|---|---|---|
| 4 | Single-invite returns plain dict, not a `MemberInviteResult` Pydantic model | The four call sites are stable; introducing a new model is a refactor without behavior change. Tracked for a future PR if the single-invite shape needs to evolve. |

### NITs (deferred)

| # | Finding | Rationale |
|---|---|---|
| 7 | Rich-markup escape on user-controlled strings | Consistent with existing kbagent output patterns; not a regression introduced by this PR. |
| 8 | `member-set-role` lacks `--yes` confirm | Plan didn't request it; demoting `admin → guest` is recoverable. |
| 9 | `invitation-cancel` is `admin`, not `destructive` | Reversible (re-send `invite`); category is consistent with the other invitation/membership ops. |
| 10 | CSV role column is case-sensitive | Defensible -- the Manage API is camelCase; matches the API's own validation. |
| 11 (security NIT) | Failed-row `note` may include a *masked* token marker | Not a leak -- token is `mask_token`'d per the existing convention; trimming that prefix is a stylistic preference. |

`make check` clean. **2414 tests pass** (+3 new regression tests). Force-pushed `8b2db76 → 6156a9b`. CI re-running.

Iteration 3 next: re-spawn an independent reviewer over the new diff to verify the fixes don't introduce new findings.

ottomansky:
## Iteration 3 — CONVERGED

Independent reviewer over the post-iteration-2 diff: **zero material findings** in code or security. Verdict: "**CONVERGED — PR is ready to merge.**"

The iteration-2 fixes were verified correct (Pydantic `extra: "allow"`, `_bulk_dry_run` multi-stack guard, `utf-8-sig` BOM strip, `--hint --from-csv` clear error, `_invite_one(project_label=...)` rename), each pinned by a regression test.

### Final state

- **CI**: `check` passed on `6156a9b` (`mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`).
- **Tests**: 2414 pass, 68 skipped (e2e by default), 0 fail.
- **Review-loop tally**: 1 BLOCKING (self-iter-1, fixed) + 5 NON-BLOCKING (iter-2, fixed) + 6 NIT (deferred with rationale) + 0 findings (iter-3 convergence).
- **Sync map**: every silent-drift surface walked (`AGENT_CONTEXT`, `CLAUDE.md`, `keboola-expert.md` Rule-6/matrix/gotchas, `SKILL.md` triggers + workflow links, `commands-reference.md`, `gotchas.md`, new `member-workflow.md`, `permissions.py`, `plugin.json` / `marketplace.json` via `make version-sync`).
- **Live verified** against project 5725 (`[Cuesta training] - Master`, us-east4): single invite + idempotency 400→noop + cancel + verify-gone + bulk dry-run all work end-to-end. Master project untouched in its original 3-member + 1-pre-existing-pending state.

Ready for human review whenever you have a moment.

ottomansky:
**Review iterations complete (kickoff §6).** Two independent agents with fresh context each walked the full diff + the security checklist from the kickoff prompt §6.5. Both verdicts: **APPROVED, 0 BLOCKING in CODE / 0 BLOCKING in SECURITY.** Sharing findings here so they don't get lost — none gate the merge.

### Both reviewers flagged (worth a follow-up PR)

- **Single-shot `invite()` / `cancel_invitation()` / `remove_member()` / `set_member_role()` don't `.strip()` the email** — the CSV path does (`member_service.py:522`), but the public single-shot methods don't. Trailing whitespace from a shell paste produces a 404 instead of an obvious match. Two-line fix: `email = email.strip()` at the top of each.
- **Bulk CSV has no email-dedup**: same email twice → both threads POST → second hits HTTP 400 → noop. Wastes one API call and produces a confusing `rows[]` entry. Fix: drop or warn on duplicate `(project, casefold(email))` after `_parse_invite_csv`.
- **Default `--workers 8`**: empirically fine, but Manage API rate limit isn't published. `BaseHttpClient` retries 429 with `Retry-After` so worst case is auto-throttle. Consider a soft cap of 4 if 429s start appearing in practice.

### Self-review (iter 1) extras

- `_parse_invite_csv` raises on the first malformed row (header / role-column issues are fatal) — inconsistent with the bulk philosophy elsewhere where one bad row never aborts the rest. Acceptable as documented; consider downgrading to per-row `MemberInviteRow(status="failed")` in a follow-up.
- Bulk path catches only `KeboolaApiError`; non-API exceptions inside `_invite_one` (e.g. `httpx.RequestError`) escape `as_completed(futures)` and abort the batch.
- `--default-role` and `--workers` silently ignored in single-shot mode; `--reason` silently ignored in `--from-csv` mode. Reject or document.
- Single-shot `dry_run` echoes `reason: ""` while the live request omits the key — phantom diff for users comparing dry-run output to a request log. Cosmetic.

### Independent reviewer (iter 2) extras

- **Pre-existing renderer bug now amplified**: `--hint client` for the new commands renders `ManageClient(base_url=..., token=...)` but the actual constructor is `ManageClient(stack_url=..., manage_token=...)`. Hint code passes `ast.parse` but throws `TypeError` on copy-paste. Existed pre-PR (org setup hint affected too) but this PR multiplies the surface. One-line fix in `hints/renderer.py:171-174`.
- `invitation-cancel --hint client` with no `--invitation-id` renders `cancel_project_invitation(..., invitation_id="None")` (literal string `"None"`, not Python `None`). Copy-paste foot-gun.
- Mixed bad+good multi-stack CSV: per-row failures are accumulated, then `ConfigError` fires on the multi-stack invariant — the accumulated failures never reach the user. Either include them in the error or surface a degraded `BulkInviteResult`.
- No self-removal guard on `member-remove` / `member-set-role` (admin → guest self-demotion). Foot-gun: user can lock their own automation out mid-flight. The API gatekeeps, but a `verify_token()` + email comparison would prevent the foot-shoot.
- Test coverage gaps: no test for duplicate-email-in-CSV race, no test for 429 mid-bulk, no test for Ctrl-C mid-bulk (`ThreadPoolExecutor` won't cleanly cancel in-flight POSTs).

### Hint-renderer compile-checks

I `ast.parse`'d all 14 rendered hints (7 commands × 2 modes). All pass. Note that **PR #237 added a parametrized regression test** that runs `ast.parse` on every data-app hint — same idea applies here; would be cheap to extend it across the project / member commands in a follow-up.

### Token / secret hygiene

- Token-leak grep: clean. No real Manage token, no Storage token in the diff.
- `MemberInviteRow.note = str(getattr(exc, "message", exc))` — protected today by upstream `mask_token` in `BaseHttpClient`, but adding a defensive `mask_token(str(exc.message))` before storing would harden against future changes to `KeboolaApiError.message`.

### Test plan checkboxes updated

- [x] Self-review (kickoff §6 iter 1)
- [x] Independent agent review (kickoff §6 iter 2+)
- [ ] `make test-e2e-invite` — still needs `E2E_MANAGE_TOKEN` + `E2E_INVITE_PROJECT_ID` in the env. Run it with the manage token when convenient.

`make check`: 2,414 passed, 5 skipped — same as the PR description claims.

padak (commented):
## Verdict

Mechanically clean — every CONTRIBUTING.md plugin-sync surface walked correctly, security matches existing manage-token discipline (env-var-or-hidden-prompt, never persisted, never on argv, never logged on the wire). Ready to merge once we agree on the manage-token UX below.

### The mental-model shift this PR introduces

Until now, manage tokens were a *setup-only* credential: `org setup` once, get per-project Storage tokens, throw the manage token away. Steady-state ran on Storage tokens which are scoped, persisted with `0600`, and gated by the kbagent permission firewall.

This PR introduces *steady-state Manage API operations* (invite, member-list, member-remove, set-role). Manage tokens now need to live in `KBC_MANAGE_API_TOKEN` long-term, or be typed via TTY prompt for every single invocation. That changes the threat surface and pulls two friction points to the surface:

### 1. Multi-stack ambiguity

Manage tokens are stack-specific (US ≠ EU ≠ Azure ≠ GCP). When `config.json` has aliases on multiple stacks, one `KBC_MANAGE_API_TOKEN` env var only covers aliases on the same stack as the token. The service correctly enforces a single-stack-URL invariant per CSV (`member_service.py:158-164`) — that safe-fails — but it forces users to split CSVs per stack and rotate env vars manually.

**Suggestion** — extend `_helpers.py:resolve_manage_token()` to take the resolved `stack_url` and look up a per-stack token. Two designs:

```bash
# Option A: per-stack env vars (mirrors how Storage tokens map to aliases today)
export KBC_MANAGE_TOKEN_US=xxx          # connection.keboola.com
export KBC_MANAGE_TOKEN_EU=yyy          # connection.eu-central-1.keboola.com
export KBC_MANAGE_TOKEN_GCP_US=zzz      # connection.us-east4.gcp.keboola.com

# Option B: single JSON map env var
export KBC_MANAGE_TOKENS='{"us": "xxx", "eu": "yyy", "gcp_us": "zzz"}'
```

Existing `KBC_MANAGE_API_TOKEN` keeps working as a single-stack fallback, so zero migration cost. For `--from-csv`, mixed-stack CSVs would just work end-to-end.

Option A is closer to existing token-resolution patterns; Option B reduces env pollution but is awkward to compose in CI/CD pipelines (escaping JSON in shell, rotating individual values). I lean Option A.

### 2. AI agent exfiltration risk

`kbagent init --read-only` blocks Storage write commands across 3 layers (kbagent policy + filesystem chmod + Claude Code deny rules). Permission firewall in this PR correctly classifies `member-remove=destructive`, so `--deny-destructive` blocks it at the CLI surface. But:

- The env var is *outside* the firewall — any subprocess (including the AI agent itself) sees the plain manage token in `os.environ`.
- The agent can shell out raw `curl -H "X-KBC-ManageApiToken: $KBC_MANAGE_API_TOKEN" ...` and the firewall does not catch raw HTTP — it's not a kbagent operation.
- Manage tokens have org-level scope — broader than per-project Storage tokens.

This is not a new risk (`org setup` already exposes the same surface), but this PR makes manage tokens persist in env *longer* than before, because steady-state operations need them. Read-only sandboxes are quietly more permissive than they look once a manage token is in env.

**Suggestion** — add a `(since v0.26.1)` entry to `gotchas.md` and an Inline Gotcha row in `keboola-expert.md`:

> When `KBC_MANAGE_API_TOKEN` is set in a shell, AI agents inherit org-level Manage API access via env, even when `kbagent init --read-only` is active. The CLI permission firewall blocks `kbagent project member-remove`, but the env-var token is plain to any subprocess. Read-only sandboxes should `unset KBC_MANAGE_API_TOKEN` before spawning agents.

A future `--deny-manage-env` flag (refuse member-API commands when the local config is read-only AND the manage token comes from env) would be a 0.26.2 follow-up — out of scope for this PR.

### Smaller asks (can be in this PR)

- `gotchas.md` — 4th `(since v0.26.1)` entry: **"`KBC_MANAGE_API_TOKEN` is stack-specific; CSVs spanning multiple stacks must be split per-stack."** Service enforces this at runtime, the gotcha shortens the user-feedback loop.
- PR description and changelog say "seven commands"; there are 6 distinct subcommands (`invite` covers single + bulk via `--from-csv`). Permission registry, hint registry, AGENT_CONTEXT all correctly count 6. Cosmetic.

Ship-as-is is a defensible call; the multi-stack manage-token UX is the one I'd really like to see addressed before users hit it in the wild.

### What I verified

- Full diff against CONTRIBUTING.md plugin synchronization map (every silent-drift surface touched: AGENT_CONTEXT, CLAUDE.md, keboola-expert.md Rule 6 + 7 matrix rows + 3 inline gotchas, SKILL.md description triggers + `member-workflow` link, commands-reference.md, gotchas.md with 3× `(since v0.26.1)` tags, new `member-workflow.md`).
- HTTP-400-not-422 normalization in `member_service.py:_noop_note_for()`; PATCH-not-PUT in `manage_client.py`; role whitelist exactly `('admin', 'guest', 'readOnly', 'share')` in `constants.py`.
- 26 service + 15 CLI + 12 manage-client + 1 E2E (`e2e_invite` marker, gated `E2E_MANAGE_TOKEN + E2E_INVITE_PROJECT_ID`, idempotent cleanup) tests pass.
- Security: `resolve_manage_token()` accepts only env or hidden TTY prompt; tokens never logged; CSV bulk parses via `csv.DictReader` (no shell exec, no path traversal); email goes into JSON body via `httpx.post(json=...)` (no URL interpolation).

ottomansky:
## Coordination across the two open PRs

**PR #236 (this PR)** — currently CONFLICTING with `main` since v0.27.0
(#237) shipped after your review. Whenever you're ready to land it,
rebase onto v0.27.0 and bump `0.26.1 → 0.27.2` (since #238 takes 0.27.1
— see below). Your "ship-as-is is defensible" verdict still stands;
nothing in the review needs re-litigation.

**PR #238 (just opened, ready for review)** — addresses the manage-
token UX you flagged in this thread. Multi-stack ambiguity + AI-
exfiltration risk, both fixed in one PR. CI green, mergeable.

Recommended design: per-stack hostname-derived env vars
(`KBC_MANAGE_TOKEN_<SUFFIX>`) + opt-in `--no-env-manage-token` session
flag + persisted policy (`kbagent permissions deny-manage-env`) +
auto-deny inside `kbagent init --read-only` workspaces. Existing
`KBC_MANAGE_API_TOKEN` keeps working as a single-stack fallback (zero
migration cost). Full scoring vs. four alternative scenarios (per-
project config.json storage, OS keychain, AWS-style profiles, refuse-
env-by-default) in `docs/manage-token-design.md` on the branch.

**Sequencing:** #238 (production code) sequences AFTER this PR lands.
The new resolver signature is backwards-compatible —
`resolve_manage_token(stack_url=None, *, allow_env=True)` keeps the
zero-arg legacy form working. When you rebase #236 it picks up the
new kwargs without conflict; the `MemberService` calls in #236 just
need a one-line update at each call site to pass `stack_url` and
`allow_env=ctx.obj["allow_manage_env"]`. I'm happy to push that as a
follow-up commit on the #236 branch once you signal you want it
rebased.

**Four review iterations on #238 already done** (self → fresh-context
→ fix → fresh-context convergence). 2467 tests pass. The PR body has
the full design rationale + test plan. Four open questions for your
gut reaction live in §9 of the design doc — happy to defer any of
them to a follow-up PR if you'd rather see them land separately.

Both PRs aim at the same goal: getting #236 landed cleanly, and
getting the hardening landed before users hit the multi-stack / AI-
exfil pain in the wild.

padak (commented):
Coupling note (full discussion on #238): this PR and #238 ship together as **0.28.0**, with REPL-based manage-token caching replacing the env-var-default model. Six of seven new commands here go through the manage-token resolver being hardened in #238; landing this PR alone would put new admin/destructive surface on the model we're retiring.

Suggested mechanic: rebase onto #238's branch, single 0.28.0 bump, consolidated plugin sync map updates. Detailed design rationale on #238.

padak:
Closing in favor of new PR from padak/keboola_agent_cli — branch has been rebased onto feat-0.29.0 and will be merged from there.

Issue #263, opened 2026-05-07
v0.30.1: keboola-mcp-server auto-update banner re-fires on every kbagent invocation


ottomansky:
> **Status update (2026-05-07):** v0.30.2 ([release](https://github.com/padak/keboola_agent_cli/releases/tag/v0.30.2), PR #262) fixes **Bug A** — the local-version probe now reads `uv tool list` instead of running `keboola_mcp_server --version`. The user-visible banner is gone for healthy installs because the probe now returns a real version → `_is_up_to_date` short-circuits as `True` → bugs B / C / D never fire.
>
> **Bugs B, C, D remain latent.** @padak is working on the remaining three. They are masked today by the bug-A fix but will reappear if any of these conditions hold: (i) MCP server is genuinely behind PyPI (Bug B fires the broken `uvx --version` upgrade subprocess), (ii) probe still returns `None` on some install path we haven't tested (Bug C falls through to the upgrade attempt), (iii) any code path triggers `maybe_auto_update()` more than once per process (Bug D amplifies). Tracking continues here until all four are addressed.

## Symptom

After upgrade to v0.30.1, every kbagent command (and every nested command
inside `kbagent repl`) prints:

```
Updating keboola-mcp-server vunknown -> v1.59.1 (via uvx)...
keboola-mcp-server upgrade skipped: ... AuthlibDeprecationWarning ...
usage: python -m keboola-mcp-server [-h] ...
python -m keboola-mcp-server: error: unrecognized arguments: --version;
continuing with current version.
```

The banner fires even after `kbagent doctor --fix` runs `uv tool install
keboola-mcp-server` and `doctor` reports MCP **PASS** (`MCP server
available via: keboola_mcp_server (transport=stdio)`).

> **Edit (corrected 2026-05-07):** the original issue body framed bug #2
> as "`vunknown` gets cached and never satisfies up-to-date". After
> reading the actual v0.30.1 source, that framing is wrong — `vunknown`
> is just a format-string artifact (`f"v{local_version or 'unknown'}"`
> in `auto_update.py:_maybe_update_mcp`), nothing is poisoned in the
> cache. The real chain is four bugs, not three; corrected analysis
> below.

## Root cause (four chained bugs)

Code references against `padak/keboola_agent_cli` `main` @ commit
`5e90febd9` (v0.30.1 tip).

### Bug A — local-version probe always returns `None` for non-pip-env installs ✅ FIXED in v0.30.2 (#262)

[...]

PR #265, opened 2026-05-07
fix(0.30.3): close issue #263 -- bugs B + C + D + E in MCP auto-update flow

padak:
## Summary

PR #262 (v0.30.2) addressed only **Bug A** from issue #263 (probe returned `None` for `uv tool`-managed installs). The reporter analyzed the code and identified **four** chained bugs; this PR addresses the remaining three.

Closes the rest of #263.

### Bug B -- uvx upgrade subprocess used a flag the binary does not honour

`_perform_mcp_update` for `uvx`-cache installs ran:

```bash
uvx --refresh --from keboola-mcp-server keboola_mcp_server --version
```

The trailing `--version` is rejected by the upstream MCP binary (no such flag), so the upgrade subprocess always exited non-zero -- the user-facing banner reported failure even when the cache refresh itself worked.

**Fix**: promote uvx to `uv tool install --upgrade keboola-mcp-server`, matching what `kbagent doctor --fix` already does. The binary lands on PATH, so subsequent runs use the faster `uv_tool` detection path. Self-healing.

### Bug C -- probe-`None` fell through to the upgrade attempt

`_maybe_update_mcp` had no gate for the case where `_get_local_mcp_version()` returns `None`. `up_to_date == None` (not `True`) bypassed the short-circuit, the function fell through to a broken upgrade subprocess every TTL window, and the user saw an `Updating ... vunknown -> v1.59.1` banner once per kbagent invocation.

**Fix**: add `if local_version is None: return mcp_latest` -- opts out of the upgrade for this TTL window. Cache TTL still ticks; the next fresh-cache pass retries detection.

### Bug D -- `kbagent repl` re-ran auto-update on every prompt

`maybe_auto_update()` is called from `cli.py:main()` on every CLI entry. `kbagent repl` re-enters Click on every prompt iteration, so the auto-update flow fired (and printed banners) once per command typed at the prompt.

**Fix**: a module-level `_AUTO_UPDATE_RAN: bool = False` sentinel. First call flips it to True; subsequent in-process calls short-circuit. The flag flips BEFORE any work so a crash mid-flow still gates subsequent re-entries. Re-exec'd processes (kbagent self-upgrade -> `execvpe` to new binary) start with a fresh sentinel because the module is reloaded into a new interpreter, so the kbagent-self-upgrade -> re-exec -> MCP-stage chain from PR #257 is preserved.

## Test plan

- [x] `make check` green: **2,779 tests passed**, 7 skipped, lint + format + skill + version + changelog clean.
- [x] **4 new regression tests** pinning all three contracts:
  - `TestPerformMcpUpdate.test_uvx_promotes_to_uv_tool_install` -- asserts the new uvx cmd; explicitly checks `--version` is GONE
  - `TestPerformMcpUpdate.test_uvx_promotion_requires_uv` -- missing-`uv` failure path
  - `TestProbeNoneSkipsUpgrade.test_local_version_none_skips_upgrade` -- the AC from #263: probe -> None; `_perform_mcp_update` NOT called
  - `TestProcessLevelSentinel.test_second_call_short_circuits` -- the AC from #263: `maybe_auto_update` body runs once across N calls in the same process
  - `TestProcessLevelSentinel.test_sentinel_is_set_even_when_body_raises` -- guards against flaky upstream PyPI re-firing the banner per prompt
- [x] Existing `TestMaybeAutoUpdate`, `TestMaybeAutoUpdateMcpIntegration`, and `TestReExecPathStillRunsMcp` autouse fixtures extended to reset `_AUTO_UPDATE_RAN` between tests so the sentinel does not gate the second test in each class.

## Acceptance criteria from issue #263 -- final state

- [x] `kbagent project list` prints zero auto-update noise on a healthy install (already true after Bug A fix in v0.30.2; still true here)
- [x] `kbagent repl` runs the auto-update check at most once per session (Bug D)
- [x] Probe-`None` does not fall through to the upgrade attempt (Bug C)
- [x] `uvx`-cache installs upgrade cleanly via promotion to `uv tool install --upgrade` (Bug B)
- [x] Regression test mocks `_get_local_mcp_version() -> None` and asserts `_perform_mcp_update` is NOT called
- [x] Regression test asserts a second invocation in the same REPL session does not re-trigger `maybe_auto_update()`


padak:
Picked up @ottomansky's update on issue #263 -- **Bug E** is now addressed in commit `046831a` on top of the previous B/C/D commit:

`uv tool upgrade keboola-mcp-server` exits 0 even when its resolver backtracks to the previously installed version (e.g. fastmcp constraint conflict). Pre-fix, kbagent printed `Updated keboola-mcp-server to v1.32.0.` -- the same version the user started from. Post-fix, both upgrade paths (`auto_update._maybe_update_mcp` and `VersionService._update_mcp`) compare pre/post versions:

- pre != post → claim updated (the happy path)
- post is None → probe failed post-upgrade; surface diagnostic
- pre == post → subprocess exit 0 but version unchanged; emit a diagnostic pointing to `uv tool install --reinstall keboola-mcp-server` so the user can investigate the underlying packaging conflict

The `updated` boolean in `kbagent update` JSON output also now reflects the actual version delta, not just the exit code.

New regression test `test_subprocess_succeeds_but_version_unchanged_reports_not_updated` simulates @ottomansky's reproducer (pre and post both v1.32.0; mock_perform returns success) and pins both the `updated == False` contract and the diagnostic message shape.

Updated changelog entry. `make check` green: 2,780 tests passed.

padak (commented):
# Review of #265 — fix(0.30.3): close issue #263 -- bugs B + C + D + E in MCP auto-update flow

## Summary

PR closes four chained bugs in the MCP auto-update flow (issue #263, partially addressed in v0.30.2). The fixes are architecturally sound: Bug B promotes the `uvx`-cache upgrade path to a persistent `uv tool install --upgrade` (self-healing); Bug C adds an early-out gate when the local-version probe returns `None` in the auto-update path; Bug D adds a per-process sentinel in `maybe_auto_update` to prevent re-firing on `kbagent repl` prompt iterations; Bug E adds pre/post version comparison in both the auto-update path and the explicit `kbagent update` path to distinguish a real upgrade from a uv resolver backtrack. Two issues found. The first is a **BLOCKING** UX regression in `VersionService.get_versions()` (displayed by `kbagent version`) where the suggested upgrade command for users with a `uvx`-cache install still shows the old broken `uvx --refresh --from <pkg> <bin> --version` invocation that was specifically removed by Bug B. The second is a **BLOCKING** logic bug in `_update_mcp()`: the `actually_updated` guard ANDs on `local_version`, so when a fresh uvx-to-uv-tool promotion succeeds (`local_version=None` pre-upgrade, `post_version="1.59.1"` post-upgrade), `updated` is silently reported as `False` and the message says "still vNone" instead of reporting a successful install.

**Verdict**: REQUEST CHANGES — 2 blocking findings.

## Verdict

- **Blocking findings**: 2
- **Nits**: 1

## Blocking findings

### `[B-1]` `src/keboola_agent_cli/services/version_service.py:437` — `get_versions()` still suggests the old broken `uvx --refresh` command to users

The `mcp_upgrade_cmd_by_method` dict in `get_versions()` still maps `"uvx"` to `f"uvx --refresh --from {MCP_PACKAGE_NAME} {MCP_BINARY_NAME} --version"` -- the exact command that Bug B was introduced to remove. `kbagent version` displays this string to the user as the "how to upgrade" hint. A user who sees `install_method: "uvx"` and copies the displayed command will get the same failure that triggered Bug B. The fix promoted `_perform_mcp_update("uvx")` to `uv tool install --upgrade` but left the user-facing display instruction unchanged.

Fix: change line 437 to `"uvx": f"uv tool install --upgrade {MCP_PACKAGE_NAME}"` (matching what `_perform_mcp_update` now actually runs). Optionally also update the `_detect_mcp_install_method` docstring at line 222 which still says `"uvx -- upgrade with uvx --refresh ..."`.

### `[B-2]` `src/keboola_agent_cli/services/version_service.py:641-663` — fresh-install via uvx promotion reports `updated=False` and emits "still vNone"

`_update_mcp()` computes `actually_updated = bool(success and post_version and local_version and post_version != local_version)`. When a user has only a uvx-cache install (`_get_local_mcp_version()` returns `None` pre-upgrade, method is `"uvx"`), `_perform_mcp_update` now runs `uv tool install --upgrade` which succeeds, and the post-upgrade probe returns the real version (e.g. `"1.59.1"`). The `and local_version` clause short-circuits to `False` because `local_version` is `None`, so `actually_updated=False`. The message branch falls to the `else` arm and emits `"upgrade exit 0 but local version still vNone"` — both the result (`updated=False`) and the message are wrong for what is actually a successful fresh install.

Note: the same code path in `auto_update._maybe_update_mcp` is correctly protected by the Bug C gate (`if local_version is None: return mcp_latest` at line 325) which prevents reaching the pre/post comparison. `_update_mcp` intentionally skips that gate (explicit `kbagent update` should attempt the upgrade even without a detected version), but it therefore inherits the edge case.

Fix: in `_update_mcp`, change `actually_updated` to also be `True` when `success` and `post_version` is set and `local_version` was `None` (i.e. it's a fresh install, not a backtrack). A minimal fix: `actually_updated = bool(success and post_version and (local_version is None or post_version != local_version))`. Add a corresponding test simulating `local_version=None` + `post_version="1.59.1"` asserting `updated=True` and a success message (not the diagnostic "still vNone" message).

## Non-blocking findings

### `[NB-1]` `plugins/kbagent/skills/kbagent/references/gotchas.md:13-14` — stale `uvx --refresh` description in the v0.30.1 MCP auto-update gotcha

The gotcha added in v0.30.1 lists the uvx upgrade command as `uvx --refresh` on line 13-14. After Bug B's fix, the actual upgrade command for uvx-method users is now `uv tool install --upgrade keboola-mcp-server`. An AI agent reading this gotcha would suggest the broken command to users who ask how to manually upgrade the MCP server when they have a uvx-cache install. Per CONTRIBUTING.md §17 (silent-drift surfaces), this file has no CI freshness check.

Fix: update line 13-14 to reflect the promoted command, and add a `(since v0.30.3)` tag to note the change.

### `[NB-2]` `src/keboola_agent_cli/services/version_service.py:480` — docstring still says "uvx --refresh" for the MCP upgrade description in `self_update`

The `self_update` docstring at line 480 describes Stage 2 as running `uvx --refresh` as one of the upgrade options. After Bug B, the actual behavior is `uv tool install --upgrade`. This is documentation drift within the code — less severe than the user-facing display issue (B-1) but still misleads contributors reading the docstring.

Fix: update line 480 to say `uv tool install --upgrade` for the uvx case.

### `[NB-3]` PR body "What changed" section has no "### Bug E" description

The PR body documents Bugs B, C, and D with dedicated subsections but has no `### Bug E` entry. Bug E (the false-success detection from the ottomansky reproducer) is mentioned only in the test plan checkbox and the changelog entry. The commit `046831a` that added Bug E was apparently a late addition. This makes the PR history harder to navigate when bisecting future regressions against the fix.

Fix: add a `### Bug E` subsection to "What changed" matching the level of detail present for B, C, and D.

## Nits

- `[NIT-1]` `tests/test_version_service.py:590-598` — the comment "Bug E fix: pre-upgrade returns 1.49.0; post-upgrade returns 1.59.1" is correct but the pre-existing test name `test_only_mcp_stale_kbagent_uptodate_still_runs_mcp` no longer captures the Bug E contract being tested. Consider a companion test named `test_local_version_none_post_set_is_fresh_install` to cover the B-2 fix when you add it.

## Verification log

- `gh pr view 265 --json title,body,files,additions,deletions,baseRefName,headRefName,labels,state` → 9 files, +323/-33, state OPEN, conventional `fix(0.30.3):` prefix ✓
- `git rev-parse --abbrev-ref HEAD` in `/tmp/kbagent-263-fix` → `fix/0.30.3-issue-263` ✓ (matches PR branch)
- `wc -l /tmp/kbagent-pr-265.diff` → 529 lines ✓
- Layer violation checks (typer in services, httpx in commands, formatter in clients) → empty (no violations) ✓
- `grep -E '@.*\.command' /tmp/kbagent-pr-265.diff` → empty (no new CLI commands; OPERATION_REGISTRY and --hint checks N/A) ✓
- `make check` in `/tmp/kbagent-263-fix` → 2780 passed, 7 skipped, 14 warnings ✓ (green)
- Layer 3 (version_service.py:437): `grep 'uvx.*refresh' version_service.py` → confirmed stale `uvx --refresh --from` upgrade_command at line 437 while `_perform_mcp_update("uvx")` now runs `uv tool install --upgrade` — **B-1 confirmed**
- `_update_mcp` logic trace: `actually_updated = bool(success and post_version and local_version and ...)` — when `local_version=None`, `and local_version` short-circuits to `False`; message falls to `else` branch emitting `"still vNone"` — **B-2 confirmed**
- `auto_update._maybe_update_mcp` same path: Bug C gate at line 325 returns early when `local_version is None`, so the pre/post comparison is never reached → B-2 does NOT apply to the auto-update path, only to `_update_mcp` ✓
- `grep 'uvx.*refresh' gotchas.md` → line 13-14 still says `uvx --refresh` — **NB-1 confirmed**
- `grep 'uvx.*refresh' version_service.py` → line 480 (docstring) still says `uvx --refresh` — **NB-2 confirmed**
- PR body reviewed: no `### Bug E` section in "What changed" — **NB-3 confirmed**
- PR body test plan says "2,779 tests"; `make check` shows 2,780 — one test added in commit `046831a` after the original description was written (minor inconsistency only)
- Behavior reproduction: could not run against a real uvx-cache install (no matching environment available). B-1 confirmed by static analysis; B-2 confirmed by code trace and absence of a test covering `local_version=None` + `post_version=set`.
- Plugin synchronization map: no new commands added, no command renames — OPERATION_REGISTRY, --hint definitions, AGENT_CONTEXT, CLAUDE.md All CLI Commands, commands-reference.md: all N/A for this PR. Version bump 0.30.2 → 0.30.3: changelog.py ✓, plugin.json ✓, marketplace.json ✓, pyproject.toml ✓, uv.lock ✓. keboola-expert.md VERSION GATE: no new version-gated commands added — no update required.
- Security checks (token masking, bare except, magic numbers, print in production): all clean ✓

- `_detect_mcp_install_method` (line 222) still documents `"uvx"` as using `uvx --refresh`. After Bug B, a user with `uvx`-only install who runs `kbagent update` gets promoted to a `uv tool install`. Is there a scenario where the user has `uvx` but NOT `uv`? If yes, Bug B's guard `if uv_path is None: return False, "uv not found on PATH ..."` is correct, but the error message should probably suggest installing `uv` explicitly. This is not a new gap but worth confirming the expected UX for uvx-without-uv users.

padak:
Addressed both blocking findings from the latest review iteration in commit `196f385`:

**B-1 ( user-facing cmd drift)**: `mcp_upgrade_cmd_by_method["uvx"]` now reads `uv tool install --upgrade keboola-mcp-server` -- matching what `_perform_mcp_update` actually runs internally since Bug B. Users running `kbagent version` no longer see broken-cmd guidance.

**B-2 (fresh-install guard hole)**: Rewrote the guard to handle all four success-branch cases explicitly:
- `pre=None, post=set` → fresh install (the case that broke; now updated=True)
- `pre=set, post=set, !=` → normal upgrade
- `pre=set, post=set, ==` → Bug E no-op (diagnostic message)
- `pre/post unknown` → probe failure

The auto-update startup path (`_maybe_update_mcp`) skips case 1 by design via Bug C gate; `_update_mcp` (explicit `kbagent update` path) handles it.

New regression tests: `test_uvx_user_facing_command_uses_uv_tool_install` (B-1) and `test_fresh_install_pre_none_post_set_reports_updated` (B-2).

`make check` green: 2,782 tests passed.

PR #298, opened 2026-05-14
fix(self-update): preserve [server] extras across kbagent update (0.40.3)

padak:
## Summary

Patch release **0.40.3**. Closes a compounding UX regression on every kbagent self-update.

**Before:** every \`kbagent update\` (and the silent auto-update on startup) silently strips the optional \`[server]\` extras off the tool install, so a user who ran \`kbagent serve --ui\` yesterday gets a \`ModuleNotFoundError: No module named 'fastapi'\` after auto-update today. The friendly error from 0.40.1 was a band-aid; this is the actual fix.

**After:** \`_update_kbagent\` probes for FastAPI (the only thing the \`[server]\` extra drags in) and, when present, runs

\`\`\`
uv tool install --force --with 'keboola-agent-cli[server]' git+...
\`\`\`

instead of the bare \`--upgrade\`. CLI-only users (no \`[server]\` installed) see no behaviour change.

## Why \`--force\` instead of \`--upgrade\`

\`uv tool install\` rejects \`--upgrade\` together with \`--with\` when the additional spec resolves to a different version than the existing tool environment. \`--force\` is uv's documented way to reapply both flags in one shot. Pip fallback uses PEP 508 spec syntax: \`keboola-agent-cli[server] @ git+...\`.

## Verification

- \`make check\`: **3087 passed**
- Sanity probe: \`VersionService._has_server_extras()\` → True in my dev env (has [server] installed)
- Detection method: \`importlib.util.find_spec('fastapi')\` — reliable proxy because FastAPI is in \`[project.optional-dependencies][server]\` only

## Compound effect

This + 0.40.1 (friendly error) + 0.40.2 (banner exports) together close the full \`serve --ui\` paper-cut cycle: install is forgiving, banner is copy-paste-able, and upgrades preserve the original install shape. After this lands, the only way a user hits \`No module named 'fastapi'\` is by installing without \`--with\` in the first place — at which point the 0.40.1 hint tells them how to fix it.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
   "plugins": [
     {
       "name": "kbagent",
-      "version": "0.40.2",
+      "version": "0.40.3",
       "source": "./plugins/kbagent",
       "description": "AI-friendly interface to Keboola Connection projects — explore configs, jobs, lineage, call MCP tools, manage dev branches, and debug SQL in workspaces",
       "category": "development"
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
   "plugins": [
     {
       "name": "kbagent",
-      "version": "0.40.1",
+      "version": "0.40.2",
       "source": "./plugins/kbagent",
       "description": "AI-friendly interface to Keboola Connection projects — explore configs, jobs, lineage, call MCP tools, manage dev branches, and debug SQL in workspaces",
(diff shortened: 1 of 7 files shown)


padak:
Consolidated into PR #297 -- the [server]-extras-preservation fix is shipping together with the banner export hint as a single 0.40.2 release. Both fixes belong to the same 'serve lifecycle UX' theme so a single coordinated release gives users a cleaner upgrade story than two back-to-back patch versions.

PR #325, opened 2026-05-21
fix(0.43.8): MCP install/upgrade opts into pre-releases (closes #324)

### LINKED ISSUE #324: Auto-update of keboola-mcp-server is silently blocked on every startup — uv refuses pre-release toon-format dependency
## TL;DR

Every `kbagent` command prints a misleading "upgrade exit 0 but local version still vX" diagnostic to stderr because the auto-update of `keboola-mcp-server` is blocked by an unsatisfiable pre-release dependency (`toon-format~=0.9.0b1`). The diagnostic blames "Python or dependency-version mismatch" and points the user at `uv tool install --reinstall`, which fails the same way. Users get silently downgraded to `keboola-mcp-server` v1.32.0 on every machine, even though v1.61.3 is on PyPI.

## Affected versions

- `kbagent` **v0.43.7** (latest at time of writing — released 2026-05-20). Confirmed not fixed in v0.43.4–v0.43.7 nor v0.44.0b1.
- `keboola-mcp-server` **v1.32.0** installed locally; **v1.61.3** on PyPI.
- `uv` 0.10.9, Python 3.12.13 (both versions declare `requires-python: >=3.10`, so Python is not the actual blocker despite what the diagnostic says).

## Symptoms

Every kbagent invocation — including `--json` calls — prints two lines to **stderr** (verified via `cmd >/tmp/out 2>/tmp/err`):

```
Updating keboola-mcp-server v1.32.0 -> v1.61.3 (via uv_tool)...
keboola-mcp-server upgrade exit 0 but local version still v1.32.0 (latest: v1.61.3).
Possible Python or dependency-version mismatch -- run `uv tool install --reinstall keboola-mcp-server` to diagnose.
```

The suggested remediation (`uv tool install --reinstall keboola-mcp-server`) "succeeds" — output ends in `Installed 2 executables: keboola-mcp-server, keboola_mcp_server` — but `uv tool list` afterwards still shows `keboola-mcp-server v1.32.0`. The fix it suggests does not fix anything.

## Root cause (verified end-to-end)

1. **kbagent runs a bare `uv tool upgrade`.** Source: `keboola_agent_cli/services/version_service.py:389`:
   ```python
   cmd = [uv_path, "tool", "upgrade", MCP_PACKAGE_NAME]
   ```
   No `--prerelease=allow` (or `--prerelease=if-necessary`) is passed.

2. **`keboola-mcp-server` ≥ 1.55.0 has a pre-release-only transitive dependency.** Its `requires_dist` includes:
   ```
   toon-format~=0.9.0b1
   ```
   `toon-format` has exactly two releases on PyPI: `0.1.0` (stable) and `0.9.0b1` (the only one matching the constraint). So the constraint can only be satisfied by a pre-release.

3. **`uv` refuses pre-releases by default.** Direct reproducer:
   ```
   $ uv tool install 'keboola-mcp-server==1.61.3' --force
   error: ... your requirements are unsatisfiable.
   hint: `toon-format` was requested with a pre-release marker (e.g.,
   toon-format>=0.9.0b1), but pre-releases weren't enabled (try:
   `--prerelease=allow`)
   ```

4. **`uv tool upgrade` exits 0 anyway** because its resolver backtracks to a previously installed version that does satisfy all constraints — here, `keboola-mcp-server` v1.32.0 (the last release before the `toon-format` pin was introduced). No version change, no error.

5. **kbagent's post-version check** (added in the v0.43.x "Bug E" fix for issue #263, at `auto_update.py:353-376`) correctly detects that `pre_version == post_version` and prints the diagnostic. But the diagnostic text was written assuming the cause is Python or transitive-dep version mismatch — it doesn't recognise the "pre-release marker refused" failure mode that `uv` actually surfaces.

## Why this matters beyond noise

- **The fleet is silently pinned to keboola-mcp-server v1.32.0.** Anyone whose machine installed kbagent before v1.55.0 of the MCP shipped is stuck. They see "v1.61.3 available" forever, and the suggested fix does nothing.
- **Skills built on the dataapp-developer plugin assume newer MCP behaviour.** E.g. its `storage-access.md` instructs the agent to call `mcp__keboola__get_project_info` and read `branch_id` and `workspace_id` from the response. On v1.32.0 those fields are not returned, so the agent has to detour through `kbagent branch list` + `kbagent workspace list` — extra round-trips that wouldn't be needed on a current MCP. This is what surfaced the bug in my session.
- **The diagnostic actively misleads.** It blames Python (the user's Python is fine) and points at a remediation that has the same defect.

## Confirmed workaround for users today

```bash
uv tool install 'keboola-mcp-server' --force --prerelease=allow
```

After this, `uv tool list` shows `keboola-mcp-server v1.61.3` and the warning disappears on subsequent kbagent calls.

## Suggested fixes

Either or both:

### Fix A — kbagent side (smallest, low-risk)

In `services/version_service.py:389`, pass `--prerelease=allow`:

```python
cmd = [uv_path, "tool", "upgrade", "--prerelease=allow", MCP_PACKAGE_NAME]
```

`--prerelease=if-necessary` is arguably cleaner — it keeps uv's default behaviour when a stable resolution exists and only permits pre-releases when required to satisfy a declared constraint. Either flag closes the bug.

While in that area, the diagnostic at `auto_update.py:373-377` could parse the resolver's stderr for `"pre-release marker"` and emit a precise hint, rather than the current "Possible Python or dependency-version mismatch" guess.

### Fix B — keboola-mcp-server side (root cause)

Drop the `toon-format~=0.9.0b1` pin until `toon-format` ships a stable 0.9.0. Either wait for upstream, swap in a different dep, or vendor the few functions actually needed. This removes the foot-gun entirely rather than working around it on every install.

## Reproducer summary

```bash
# 1) Confirm the install is stuck despite kbagent reporting v1.61.3 is available
uv tool list | grep keboola-mcp-server
# → keboola-mcp-server v1.32.0

# 2) Confirm the warning is on stderr (so --json output is clean, but every command is noisy)
kbagent --json branch list --project <alias> >/tmp/out 2>/tmp/err
head /tmp/err   # → the two-line warning
head /tmp/out   # → clean JSON

# 3) Confirm the "fix" in the diagnostic does nothing
uv tool install --reinstall keboola-mcp-server
uv tool list | grep keboola-mcp-server
# → still v1.32.0

# 4) Confirm uv's underlying error
uv tool install 'keboola-mcp-server==1.61.3' --force
# → "your requirements are unsatisfiable. hint: `toon-format` was requested
#    with a pre-release marker ... try: `--prerelease=allow`"

# 5) Confirm the workaround
uv tool install 'keboola-mcp-server' --force --prerelease=allow
uv tool list | grep keboola-mcp-server
# → v1.61.3
```

## Environment

- macOS 25.5.0 (Darwin), Apple Silicon
- `kbagent` v0.43.7 installed via `uv tool install` at `~/.local/share/uv/tools/keboola-agent-cli`
- `keboola-mcp-server` venv at `~/.local/share/uv/tools/keboola-mcp-server`, Python 3.12.13
- `uv` 0.10.9

padak:
## Summary

Every `kbagent` invocation silently failed to auto-update `keboola-mcp-server` and printed a misleading two-line warning to stderr, leaving the fleet pinned to the stale MCP server **v1.32.0** even though v1.61.3 is on PyPI. Closes #324 (thanks @ottomansky for the end-to-end diagnosis).

## Root cause (verified end-to-end in this PR)

- `keboola-mcp-server` >= 1.55.0 declares a **pre-release-only** transitive dependency: `toon-format~=0.9.0b1`.
- On PyPI, `toon-format` ships exactly two releases — `0.1.0` (stable) and `0.9.0b1` (pre-release) — so the `~=0.9.0b1` constraint can **only** be satisfied by the pre-release.
- `uv` refuses pre-releases by default, so the bare `uv tool upgrade keboola-mcp-server` could not resolve the latest MCP. Instead of erroring, uv's resolver **backtracked to v1.32.0** (the last release predating the pin) and exited 0.
- kbagent's post-upgrade version check (the #263 "Bug E" guard) then saw `pre == post` and emitted a diagnostic — but the text blamed Python and pointed at `uv tool install --reinstall keboola-mcp-server`, which hits the **identical wall**.

Reproduced locally (uv 0.10.x): a dry-run resolve of `keboola-mcp-server==1.61.3` fails with `No solution found ... try: --prerelease=allow`; adding `--prerelease=allow` resolves cleanly (`+ toon-format==0.9.0b1`).

> ⚠️ The issue suggested `--prerelease=if-necessary` as a cleaner alternative. **It does not fix the bug.** A *stable* `toon-format` (0.1.0) exists, so uv judges a pre-release "unnecessary" and then fails the pin. Empirically, only `--prerelease=allow` resolves it (`if-necessary`, `explicit`, and `if-necessary-or-explicit` all fail).

## Changes

Fixes **every** path that installs or upgrades the MCP server — not just the one line cited in the issue:

| Path | File | Fix |
|---|---|---|
| Auto-update `uv_tool` / `uvx` | `version_service._perform_mcp_update` | `--prerelease=allow` |
| Auto-update `pip_env` | same | `--pre` |
| User-facing `upgrade_command` (all 4 methods, incl. fresh `none`) | `version_service.get_versions` | opt-in flags |
| `doctor --fix` install (`ensure_mcp_installed`) + uvx fallback hint | `mcp_service` | `--prerelease=allow` |
| Backtrack diagnostic | `auto_update` | recommends `--reinstall --prerelease=allow`; no longer blames Python |
| First-time-setup install command | `SKILL.md` | `--prerelease=allow` |

Shared `MCP_UV_PRERELEASE_FLAG` / `MCP_PIP_PRERELEASE_FLAG` constants live in `constants.py` (next to the other MCP constants) with the full toon-format rationale; `version_service` re-exports them for existing callers.

## Out of scope (intentional)

The **runtime** uvx spawn fallback in `mcp_service.detect_mcp_server_command` (`uvx keboola_mcp_server`) is left unchanged. It is a spawn path, not an install, and its existing comment documents a real risk that `--prerelease=allow` could pull a broken pre-release dependency at runtime. After this PR fixes the install paths, that fallback is only reached when no local binary exists. Happy to fold it in if reviewers prefer.

## Tests

- `test_version_service.py`: asserts the flag in all three internal upgrade commands and the user-facing recommendation for all four install methods; new `TestMcpPrereleaseOptIn` parametrized regression class (pins literal expected flags, not the production constant).
- `test_mcp_service.py`: asserts `ensure_mcp_installed` passes `--prerelease=allow` and the fallback hint carries it.
- `test_auto_update.py`: asserts the corrected diagnostic string and that the old "Possible Python or dependency-version mismatch" wording is gone.

Existing asserts were substring checks (`"uv tool upgrade" in cmd`) that passed even with the bug present — which is exactly why it shipped. They are now tightened.

`make check` is green (3445 passed).


padak (commented):
# Review of #325 — fix(0.43.8): MCP install/upgrade opts into pre-releases (closes #324)

## Summary

This PR fixes a silent regression where every `kbagent` invocation failed to auto-update `keboola-mcp-server` past v1.32.0 because `keboola-mcp-server >= 1.55.0` pins a pre-release-only transitive dependency (`toon-format~=0.9.0b1`) that uv refuses to resolve without `--prerelease=allow`. The fix is thorough: all three internal upgrade paths in `_perform_mcp_update`, all four user-facing `upgrade_command` strings in `get_versions`, the `doctor --fix` install path in `ensure_mcp_installed`, and the backtrack diagnostic in `auto_update.py` all receive the opt-in flags. The PR also adds targeted regression tests that pin the literal expected flag strings (not the production constant), which is the correct approach given that the bug shipped precisely because the old tests used substring-only checks. One install-hint message in `mcp_service.py` — the `kbagent doctor` check (not `--fix`) warning shown when the uvx fallback is active — was missed and still recommends `uv tool install keboola-mcp-server` without `--prerelease=allow`. This is within the PR's stated scope. No new commands; no silent-drift surfaces to update. Verdict: **COMMENT**.

## Verdict

- **Verdict**: COMMENT
- **Blocking findings**: 0
- **Nits**: 1

## Blocking findings

(none)

## Non-blocking findings

### `[NB-1]` `src/keboola_agent_cli/services/mcp_service.py:1489` — `kbagent doctor` uvx-fallback hint missing `--prerelease=allow`

The `_is_mcp_available_for_doctor()` function at line 1489 emits the user-facing message: `". For faster startup, run: uv tool install keboola-mcp-server"`. A user who follows this recommendation will encounter the same resolver backtrack the PR is fixing — uv will silently install v1.32.0 and exit 0. This path is distinct from `ensure_mcp_installed()` (which was fixed); both live in `mcp_service.py` but are called from different entry points (`doctor --fix` vs the doctor health check). The PR description explicitly claims to fix "every path that installs or upgrades the MCP server" — this one was missed.

Fix: change line 1489 to `message += f". For faster startup, run: uv tool install {MCP_UV_PRERELEASE_FLAG} keboola-mcp-server"` and add a test assertion in `TestDockerMcpCheck` (already has `test_uvx_fallback`) that the message carries `--prerelease=allow`.

## Nits

- `[NIT-1]` `plugins/kbagent/skills/kbagent/SKILL.md:366` — flag position inconsistency. The code uses `uv tool install --prerelease=allow keboola-mcp-server` (flag before package name, matching uv's canonical form), while SKILL.md writes `uv tool install keboola-mcp-server --prerelease=allow` (flag after package name). Both forms work with uv, but aligning with the implementation avoids copy-paste confusion.

## Verification log

- `gh pr view 325 --json title,body,files,additions,deletions,state` → 13 files, +195/-20, state OPEN, conventional `fix(0.43.8):` prefix matches bug-fix type ✓
- `git rev-parse --abbrev-ref HEAD` → `fix/issue-324-mcp-prerelease` (matches `<branch>`) ✓
- Read `CONTRIBUTING.md` Plugin synchronization map → no new commands added; `commands/`, `cli.py`, `context.py`, `permissions.py`, `hints/definitions/` all untouched ✓ (synchronization map not applicable)
- Read `CLAUDE.md` `## All CLI Commands` → no command surface changes; section update not required ✓
- Read `plugins/kbagent/agents/keboola-expert.md` §1 Rule 6, §2 Tool Matrix, §3 Gotchas → no new commands or behavior changes; updates not required ✓
- Layer violation checks (typer in services, httpx in commands, formatter in clients) → empty ✓
- Convention checks (magic numbers, raw error_code strings, bare `except:`, `print()` in production, token leaks) → all empty ✓
- Version consistency: `pyproject.toml` `0.43.8`, `plugin.json` `0.43.8`, `marketplace.json` `0.43.8`, `uv.lock` `0.43.8` → consistent ✓
- `changelog.py` has `"0.43.8"` entry with three bullet points ✓
- `make check` (worktree `fix/issue-324-mcp-prerelease`) → `3445 passed, 7 skipped, 99 deselected` ✓
- `grep -n "MCP_UV_PRERELEASE_FLAG\|MCP_PIP_PRERELEASE_FLAG" constants.py` → `230: MCP_UV_PRERELEASE_FLAG: str = "--prerelease=allow"`, `231: MCP_PIP_PRERELEASE_FLAG: str = "--pre"` ✓
- `_perform_mcp_update` uv_tool/pip_env/uvx paths each carry the flag (verified lines 397, 402, 415 in version_service.py) ✓
- `get_versions` mcp_upgrade_cmd_by_method dict (lines 632-635) carries flags for all four methods ✓
- `ensure_mcp_installed` (mcp_service.py:173) carries `MCP_UV_PRERELEASE_FLAG` ✓; uvx fallback hint (mcp_service.py:202) carries it ✓
- `auto_update.py:381` backtrack diagnostic: `run \`uv tool install --reinstall --prerelease=allow keboola-mcp-server\`` ✓
- `mcp_service.py:1489` `kbagent doctor` check warning: `run: uv tool install keboola-mcp-server` — missing `--prerelease=allow` (NB-1 finding)
- `SKILL.md:366` install command updated to include `--prerelease=allow` ✓; flag position is after package name rather than before (NIT-1)
- `plugins/kbagent/skills/kbagent/references/gotchas.md` line 587-591 describes the auto-update upgrade command as `uv tool upgrade / pip install -U` without mentioning `--prerelease=allow` — this entry predates the fix and was not updated; however, since the change is behavior of existing infrastructure (not a new command), this falls into NON-BLOCKING at most, and given the entry is already long and the fix is in code, left as observation only
- `TestMcpPrereleaseOptIn` class: parametrized tests pin literal `"--prerelease=allow"` / `"--pre"` strings (not production constants) — correct regression discipline, as the bug shipped because old tests used `"uv tool upgrade" in cmd` substring checks
- uvx runtime spawn fallback (`detect_mcp_server_command` line 126-129) intentionally omits `--prerelease=allow` with explanatory comment; PR description explicitly documents this as out-of-scope ✓
- No E2E tests for this PR: applicable because the fix is in installer/upgrade paths, not in CLI command behavior (no new `kbagent` commands), and E2E tests require spawning real uv + PyPI, which is not practical in CI
- Behavior reproduction: cannot reproduce against live uv + PyPI (no network access in sandbox), but the fix is mechanically correct — the flag is present in every subprocess list that gets passed to `subprocess.run`; test suite confirms the behavior

(none)

PR #446, opened 2026-06-19
feat(install): quiet the install wall, add banner + next steps

jerab111:
@padak zkusil jsem vylepsit onboarding. snad jsem to moc nerozresral

---

<a href="https://app.devin.ai/review/keboola/cli/pull/446" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -9,19 +9,21 @@
 # from `git+` recompiles the bundled React SPA via npm on every install, which
 # takes 2-4 minutes on WSL2. A prebuilt wheel is a few-seconds download instead.
 #
-# Requirements: `uv` (the install backend) and `curl`. The guide installs uv a
-# couple of steps before this. If no wheel asset exists for the latest release
-# yet (older releases predate the release workflow), this falls back to the
-# `git+` source build so the install still succeeds.
+# Requirements: `curl`. `uv` (the install backend) is auto-installed from its
+# official source if missing -- opt out with KBAGENT_NO_UV_BOOTSTRAP=1 to manage uv
+# yourself. If no wheel asset exists for the latest release yet (older releases
+# predate the release workflow), this falls back to the `git+` source build.
 #
 # Env knobs:
-#   KBAGENT_NO_SERVER=1       install CLI-only (skip the [server] extras: FastAPI/
-#                             uvicorn for `kbagent serve`). Default bundles them so
-#                             `kbagent serve --ui` works out of the box.
-#   KBAGENT_INSTALL_VERBOSE=1 show the full `uv tool install` output (the resolved
-#                             dependency list) instead of a single progress line.
-#                             Also enabled with the --verbose / -v argument.
-#   NO_COLOR=1                disable ANSI colour (https://no-color.org).
+#   KBAGENT_NO_SERVER=1        install CLI-only (skip the [server] extras: FastAPI/
+#                              uvicorn for `kbagent serve`). Default bundles them so
+#                              `kbagent serve --ui` works out of the box.
+#   KBAGENT_NO_UV_BOOTSTRAP=1  do NOT auto-install uv when missing; print the manual
+#                              install command and exit instead.
+#   KBAGENT_INSTALL_VERBOSE=1  show the full `uv tool install` output (the resolved
+#                              dependency list) instead of a single progress line.
+#                              Also enabled with the --verbose / -v argument.
+#   NO_COLOR=1                 disable ANSI colour (https://no-color.org).
 
 set -eu
 
@@ -61,9 +63,10 @@ fi
 # the banner never renders as mojibake on a legacy terminal.
 UTF8=0
 case "${LC_ALL:-${LC_CTYPE:-${LANG:-}}}" in *UTF-8* | *utf-8* | *UTF8* | *utf8*) UTF8=1 ;; esac
-if [ "$UTF8" -eq 1 ]; then CHECK='✓'; CROSS='✗'; else CHECK='OK'; CROSS='x'; fi
+if [ "$UTF8" -eq 1 ]; then CHECK='✓'; CROSS='✗'; ARROW='→'; else CHECK='OK'; CROSS='x'; ARROW='->'; fi
 
 LOG_FILE=$(mktemp 2>/dev/null || echo "/tmp/kbagent-install.$$.log")
+_BG_PID="" # pid of the currently-backgrounded step, for interrupt cleanup
 
 # Fractional sleep is not guaranteed by POSIX; probe once and fall back to 1s
 # so the spinner never busy-loops on a shell whose `sleep` rejects "0.1".
@@ -117,8 +120,10 @@ run_step() {
   fi
   "$@" >"$LOG_FILE" 2>&1 &
   _pid=$!
+  _BG_PID=$_pid
   _spin "$_pid" "$_msg"
   if wait "$_pid"; then _rc=0; else _rc=$?; fi
+  _BG_PID=""
   if [ "$_rc" -eq 0 ]; then
     printf '\r  %s%s%s %s            \n' "$GREEN" "$CHECK" "$RESET" "$_msg" >&2
   else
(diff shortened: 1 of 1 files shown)


padak (commented):
# Review of #446 — feat(install): quiet the install wall, add banner + next steps

## Summary

This PR rewrites the UX layer of `install.sh`: adds a branded ASCII banner,
a spinner-with-summary for the long `uv tool install` step, colour output
(NO_COLOR-safe, TTY-gated), verbose mode (`--verbose` / `KBAGENT_INSTALL_VERBOSE=1`),
and a "Next steps" section after successful install. It also fixes a latent
bug where a source-build failure was swallowed silently by `set -e` instead
of showing an actionable error. No Python code, no CLI commands, no plugin
surfaces touched. All CI checks pass. Verdict: **APPROVE** — one non-blocking
finding and two nits.

## Verdict

- **Verdict**: APPROVE
- **Blocking findings**: 0
- **Nits**: 2

## Blocking findings

(none)

## Non-blocking findings

### `[NB-1]` `install.sh:66,118` — single `LOG_FILE` is overwritten on second `run_step` call; wheel-failure details lost when both paths fail

`run_step` redirects command output to `$LOG_FILE` (line 118). When the wheel
install is attempted first and fails (e.g. a partial download or a
non-zero `uv` exit), the file is overwritten by the subsequent source-build
`run_step`. `dump_log` (shown on final failure at line 197) then prints only
the source-build error. The wheel failure reason is silently lost.

This matters most in the unusual-but-real scenario where the wheel asset
exists (header check passes), the download starts, and `uv` aborts mid-way
(network reset, disk full, etc.) — the user sees `Install failed` and a log
that talks about `git+`, not about the wheel, making it look like the fallback
also ran.

Easiest fix: rotate the log file per step (e.g. `LOG_FILE=$(mktemp ...)` at
the top of `run_step`) and `dump_log` each on failure. Or append (`>>`
instead of `>`) and note which step wrote which section. Not a blocker because
the combined-failure path is rare and the user can always retry with
`KBAGENT_INSTALL_VERBOSE=1`.

## Nits

- `[NIT-1]` `install.sh:64,70` — `sleep 0.1` probe runs unconditionally during
  every install even when the spinner will never be used (non-TTY / verbose mode).
  Consider moving the probe inside the TTY branch of `run_step`, or accept the
  ~100 ms hit as negligible (most installers pay more). Truly minor.

- `[NIT-2]` `install.sh:217-218` — file ends without a trailing newline after
  the final `info "..."` call. `git log --oneline` shows the diff applies cleanly,
  but `wc -l` returns 218 while the file on disk has 218 lines all ending with
  `\n` — so this is actually fine. Disregard if confirmed by `tail -c1 | xxd`
  showing `0a`.

## Verification log

- `gh pr view 446 --json title,body,files,additions,deletions,state` → 1 file
  (`install.sh`), +150 / -18, state OPEN, `feat(install):` prefix ✓
- `git rev-parse --abbrev-ref HEAD` → `install-first-run-ux` ✓ (matches PR branch)
- Layer violation check (`grep typer/click/httpx in diff`) → empty ✓ (no Python touched)
- Convention checks (magic numbers, raw error codes, bare except, print(),
  token in logs) → all empty ✓
- `make check` (after `uv sync --extra server`) → **4124 passed, 8 skipped** ✓
- `grep '@.*_app\.command' git diff main...HEAD -- commands/` → empty ✓
  (no new CLI commands; Plugin synchronization map not applicable)
- Token placeholder in Next-steps output (`YOUR_TOKEN`) is a literal
  placeholder string, not a real credential ✓
- `$@` loop at lines 42-46: safe when piped via `curl ... | sh` (`$@` is empty)
  and correctly handles `sh -s -- --verbose` per comment ✓
- `_spin` char-index math (`cut -c $((_i+1))`): POSIX-correct, cycles
  `|/-\` without off-by-one ✓
- `run_step` is called only inside `if run_step ...` at both call sites (lines
  174 and 188) — satisfies the `set -e` safety requirement stated in the
  function comment ✓
- Behavior delta on source-build failure: original used bare `uv tool install`
  (silent `set -e` abort); new wraps it in `if run_step ... ; then installed=1;
  fi` + explicit `"Install failed"` + `dump_log` + `exit 1`. Net improvement ✓
- End-of-file newline: `tail -c1 | xxd` → `0a` ✓ (NIT-2 above is moot)
- `dump_log` / `cleanup` trap: `rm -f "$LOG_FILE"` on EXIT regardless of
  success/failure — no temp-file leak ✓

(none)

jerab111:
@padak #446 is updated  again — the new commit fixes all 3 of Devin's points (missing fi, the non‑UTF‑8 arrow, Ctrl+C cleanup) and adds the uv auto‑install from UT‑4281. CI's green.

padak (commented):
# Delta Review of #446 — feat(install): auto-install uv if missing (UT-4281) + review fixes

> **Delta/follow-up review** of the second commit (`3ea7035`) pushed to this PR on 2026-06-19.
> The prior review (commit `e96b15e`) returned APPROVE with one non-blocking finding (shared `$LOG_FILE`)
> and two nits. This review covers what the new commit changes and verifies the whole file remains coherent.
> Generated by `kbagent-pr-reviewer` subagent. The verdict below is advisory; the human author retains every veto.

## Summary

The new commit adds four changes to `install.sh`: (1) `ensure_uv()` — auto-installs uv via `curl | sh` when missing, with a `KBAGENT_NO_UV_BOOTSTRAP=1` opt-out; (2) adds the missing `fi` that caused a real `sh` syntax error at EOF (verified: first commit fails `sh -n`, new commit passes); (3) introduces the `ARROW` variable to eliminate the hardcoded non-UTF-8 `→` glyph from commit 1; (4) tracks `_BG_PID` and adds `trap ... INT TERM` for Ctrl+C cleanup of backgrounded children. All four changes are correct. `make check` passes (4124 tests, 0 failures). `sh -n` and `bash -n` both exit 0 on the final file. One new non-blocking finding: when the `ensure_uv` bootstrap itself fails, the captured log output from `run_step` is silently discarded — `dump_log` is not called on that exit path, leaving the user with no diagnostic detail. The prior non-blocking finding (shared `$LOG_FILE` losing the wheel error when source also fails) carries forward unchanged and now has a third writer in the same chain, but the ordering prevents any actual overlap during normal flow.

## Verdict

- **Verdict**: APPROVE
- **Blocking findings**: 0
- **Nits**: 1

## Blocking findings

(none)

## Non-blocking findings

### `[NB-1]` `install.sh:171-175` + `install.sh:194-197` — `dump_log` not called when uv bootstrap fails

`ensure_uv` calls `run_step "Installing uv ..."` which, in quiet/TTY mode, writes all uv-installer output to `$LOG_FILE`. If the `run_step` call fails (network error, 403 from astral.sh, TLS failure, `curl` reports a partial download), `ensure_uv` prints `"error: could not install uv automatically."` and returns 1. The outer `if ! ensure_uv` block at line 194 then prints `"Cannot continue without uv."` and calls `exit 1`. The `EXIT` trap fires `cleanup()`, which **deletes `$LOG_FILE`** before the user ever sees its contents. The user is told to install uv manually but cannot see why the automated install failed.

The existing `kbagent install failure` path at line 244 correctly calls `dump_log` before `exit 1`. The same pattern should be applied here.

Fix: add `dump_log` inside `ensure_uv`'s failure branch, immediately before `return 1` at line 175:

```sh
    dump_log
    return 1
```

Note: this extends the prior NB-1 finding (shared `$LOG_FILE` across wheel→source fallback) now that a third `run_step` call exists. All three are sequential so there is no concurrency overlap, but if uv bootstrap fails, `$LOG_FILE` holds the uv error; adding `dump_log` there exhausts all the cases where a failure silently discards diagnostic output.

### `[NB-2]` `install.sh:172` — silent exit-0 when `curl` fails in the uv bootstrap pipe (no `pipefail`)

The uv bootstrap runs `sh -c 'curl -LsSf https://astral.sh/uv/install.sh | sh'`. Because the script is `#!/bin/sh` with no `set -o pipefail`, the exit code of `sh -c '...'` is the exit code of the **last command in the pipeline** — the inner `sh`. If `curl` fails after the connection is established but before the inner `sh` completes (e.g., a partial download), the inner `sh` exits with whatever error its partial input caused. More importantly, if `curl -f` exits non-zero before writing anything, the inner `sh` receives EOF on stdin and exits 0 — the pipe exit code is 0, `run_step` reports success, and the script proceeds to call `command -v uv` at line 191.

The guard at line 191 (`command -v uv >/dev/null 2>&1`) catches this correctly: if uv was not installed despite the pipe appearing to succeed, `ensure_uv` returns 1 and the script exits. So the silent `curl` failure is **caught by a post-install check rather than by the pipe exit code**, and the user error message is correct. This is not a safety hole, but the failure message says "could not install uv automatically" when the real cause is a network error — and with `dump_log` absent (NB-1), no further detail is available.

A minimal improvement would be to split the `sh -c` into two steps — first download to a temp file (so curl's exit code is directly testable), then run. Or add `set -o pipefail` to the inner `sh -c` if the target `sh` supports it. Neither is required to merge; documenting the failure mode is sufficient.

## Nits

- `[NIT-1]` `install.sh:194-197` — The `if ! ensure_uv; then` block prints `"Cannot continue without uv."` as a second error message after `ensure_uv` has already printed `"error: could not install uv automatically."`. The two messages are slightly redundant in the bootstrap-fail case. Changing the outer message to just `info "${DIM}Tip: https://docs.astral.sh/uv/ for manual install.${RESET}"` (dropping the redundant "Cannot continue") would make the output cleaner; the `exit 1` already terminates the script.

## Verification log

- `gh pr view 446 --json title,state,files` → OPEN, 1 file (`install.sh`), +208/-25 across the full PR ✓
- `gh api repos/jerab111/cli/compare/e96b15e...3ea7035 --jq .files[].patch` → confirmed the 4 changes described in the PR message ✓
- `sh -n /tmp/kbagent-install-446-final.sh` → exit 0 ✓ (first commit fails with "unexpected end of file" at line 219)
- `bash -n /tmp/kbagent-install-446-final.sh` → exit 0 ✓
- `make check` → 4124 passed, 8 skipped, 0 failures ✓
- `sh -n` on the first commit's file (`e96b15e`) → exit 2, "syntax error: unexpected end of file" — confirms the `fi` fix is real and correct ✓
- Function ordering verified: `LOG_FILE` (line 68), `_BG_PID=""` (line 69), `info()` (line 75), `run_step()` (line 113), `cleanup()` (line 142), `trap` (lines 148-149), `ensure_uv()` (line 157), `ensure_uv` call (line 194) — all infra defined before `ensure_uv` uses it ✓
- `ARROW` glyph at line 66: `ARROW='→'` on UTF-8, `ARROW='->'` otherwise; used at line 257 as `%s` arg to `printf` — no embedded literal `→` in non-UTF-8 path ✓
- `_BG_PID` tracking in `run_step` (lines 123, 126) and `cleanup()` at line 145: initialized at line 69, set before background fork, cleared after `wait` — correct. Double-cleanup on INT+EXIT is idempotent (kill of already-dead pid ignored, `rm -f` of already-removed file ignored) ✓
- `UV_PYTHON_PREFERENCE` at line 188: uses `:=` form (`"${UV_PYTHON_PREFERENCE:=only-managed}"`) — only sets if unset or empty; users who pre-set the variable keep their preference ✓
- Security: `curl -LsSf https://astral.sh/uv/install.sh | sh` is the **identical command** the previous script printed as the manual step for users missing uv. Automating the same command does not expand trust surface beyond what the install flow already required. Both the kboola install script and the uv bootstrapper use HTTPS to pinned hostnames ✓
- `dump_log` absence on uv bootstrap failure: confirmed by reading lines 171-198 — `dump_log` IS present at line 248 (kbagent install failure path) but NOT at line 175 (uv bootstrap failure return path) or line 197 (outer exit) ✗ → NB-1

## Prior review status

| Prior finding | Status after new commit |
|---|---|
| NB-1 — shared `$LOG_FILE` (wheel error lost when source also fails) | **Unchanged** — still applies, now extended: uv bootstrap is a third `run_step` writer, and its failures additionally discard the log without calling `dump_log` (see NB-1 above) |
| NIT-1 — false alarm (trailing newline) | **Resolved** — `fi` now correctly terminates the file |
| NIT-2 — hardcoded `→` literal (non-UTF-8 arrow) | **Resolved** — `ARROW` variable added at line 66 and used at line 257 ✓ |

padak (approved):
Approving.

Two review passes — the initial review at `e96b15e` and the delta review of `3ea7035` — both land **APPROVE** with no blocking findings.

The new commit:
- fixes the EOF `fi` syntax error that slipped past `make check` + CI,
- restores the non-UTF-8 arrow fallback (`ARROW` var, matching the existing `✓`/`✗` pattern),
- kills the backgrounded step on Ctrl+C (`_BG_PID` + `trap ... INT TERM`, exit 130),
- adds the `uv` auto-bootstrap (UT-4281) with a clean `KBAGENT_NO_UV_BOOTSTRAP=1` opt-out and `UV_PYTHON_PREFERENCE=only-managed` to dodge the wrong-arch system-Python failure.

Verified locally: `sh -n install.sh` and `bash -n install.sh` parse clean; `make check` green (4124 tests).

One optional, non-blocking nit remains for a follow-up: the `uv`-bootstrap failure path (`install.sh:171-175`) doesn't call `dump_log`, unlike the kbagent-install failure path at line ~248 — so a quiet-mode uv bootstrap failure discards its captured log. Trivial to mirror; not a merge blocker.

PR #424, opened 2026-06-15
chore(release): rename PyPI distribution to keboola-cli

Matovidlo:
## Summary
Renames the **PyPI distribution name** from `keboola-agent-cli` to **`keboola-cli`**, to match the PyPI **Trusted Publisher** registered for this repo. Scope is deliberately limited to the distribution name; the import package (`keboola_agent_cli`) and the binary (`kbagent`) are unchanged.

## What changed (distribution name only)
- `pyproject.toml` `[project].name` → `keboola-cli`
- `constants.APP_NAME` (drives `importlib.metadata` version lookup **and** the User-Agent token)
- `auto_update.distribution(...)`, and the `[server]` self-update specs in `version_service.py` / `commands/serve.py`
- `scripts/sync_version.py` `KBAGENT_DIST_NAME`, `uv.lock` (project package now `keboola-cli`)
- `README.md` `[server]` install spec, `http_base` User-Agent docstring
- Tests updated for the new UA token + install specs (`test_client`, `test_metastore_client`, `test_manage_client`, `test_auto_update`, `test_version_service`, `test_sync_version_script`)

## Deliberately NOT changed
- **Config dir** `~/.config/keboola-agent-cli/` (`platformdirs.user_config_dir(...)`) — renaming it would **orphan every existing user's stored tokens**; that's a separate migration, out of scope.
- **Claude Code plugin/marketplace logical name** `kbagent@keboola-agent-cli` and the plugin cache path — intentionally logical names (per changelog), not the dist/repo name.
- `changelog.py` historical entries (accurate history).

## Why
`uv tool install keboola-cli` / `pip install keboola-cli` is the intended install name, and it must equal the PyPI project the Trusted Publisher is bound to. `keboola-cli` is confirmed **available** on PyPI.

## Change type
Chore (release/packaging). No behavior change; binary and import package unchanged.

## Test plan
- 371 affected tests pass (`test_client`, `test_metastore_client`, `test_manage_client`, `test_auto_update`, `test_version_service`, `test_sync_version_script`).
- Gates pass locally: ruff lint + format, `ty` (no new diagnostics), `command-sync-check`, `skill-check` (SKILL.md unchanged), `version-check` (uv.lock/plugin.json/marketplace.json in sync).

## Deployment / Rollback
Merge & automatic. Rollback = revert of this PR.

PR #457, opened 2026-06-22
fix(release): attach the pip/uv wheel on every release

soustruh:
## Summary
This fixes wheel publication for releases so the pip/uv wheel is attached every time, including releases created by the native binary pipeline.

## How I found this
I hit the broken update path myself when `kbagent` tried to update itself and failed because the expected wheel asset was missing. The update did not complete, so `kbagent` stayed on the currently installed version instead of downloading the release wheel.

## Root Cause
- `release.yml` only runs on `release: published`.
- `release-kbagent.yml` creates the GitHub Release from a tag push.
- When `release-kbagent.yml` creates the release with `GITHUB_TOKEN`, GitHub does not emit `release: published`.
- So bot-created releases can ship without the wheel unless something else triggers the wheel workflow.

## Evidence
- `v0.64.0` was authored by `github-actions[bot]`.
- Its wheel assets were uploaded later in a separate manual `workflow_dispatch` run by me.
- `v0.65.0` was authored by `padak`, and `release.yml` ran automatically on the `release` event.

## Fix
After creating the GitHub Release, `release-kbagent.yml` now dispatches `release.yml` directly for the same tag.

This keeps wheel publishing attached to every release, regardless of whether the release page was created by a human or by Actions.

## Notes
- The change is minimal: one `actions: write` permission and one `gh workflow run release.yml -f tag=...` step.
- `v0.64.0` is still a useful example because the wheel was missing at initial publish time and only appeared later after manual backfill.

### DIFF
--- a/.github/workflows/release-kbagent.yml
+++ b/.github/workflows/release-kbagent.yml
@@ -255,6 +255,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write # softprops/action-gh-release creates the GitHub Release
+      actions: write # dispatch release.yml (the wheel-attach workflow) below
     env:
       VERSION: ${{ needs.version.outputs.VERSION }}
     steps:
@@ -270,6 +271,18 @@ jobs:
           tag_name: v${{ env.VERSION }}
           files: release/*
           prerelease: ${{ needs.version.outputs.IS_PRERELEASE == 'true' }}
+      # Attach the pip/uv wheel too. release.yml builds + uploads it (the asset that
+      # auto-update and install.sh download) on `release: published` -- but a release
+      # created above with GITHUB_TOKEN does NOT emit that event, so on a bot-cut
+      # release the wheel never got attached (this is what broke v0.64.0).
+      # workflow_dispatch is the documented exception to the GITHUB_TOKEN no-cascade
+      # rule, so dispatch release.yml directly; it fires no matter who cut the release.
+      # The release already exists (created above), so release.yml's upload lands.
+      - name: Attach the pip/uv wheel (dispatch release.yml)
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: gh workflow run release.yml -f tag="v${VERSION}"
 
   # ── Publish to S3 (cli-dist.keboola.com/keboola-cli2/) and index the apt/rpm/apk
   #    repos so `apt-get install keboola-cli2` works. Real; needs AWS + GPG secrets.
--- a/.github/workflows/release-kbagent.yml
+++ b/.github/workflows/release-kbagent.yml
@@ -276,8 +276,9 @@ jobs:
       # job creates with GITHUB_TOKEN does NOT emit that event, so when no one
       # published the release by hand the wheel never got attached -- this broke
       # v0.64.0. workflow_dispatch is the documented exception to the GITHUB_TOKEN
-      # no-trigger rule, so dispatch release.yml directly; it runs whoever created
-      # the release. The release already exists (created above), so the upload lands.
+      # no-trigger rule, so dispatch release.yml directly; that runs it no matter who
+      # or what created the release. The release already exists (created above), so
+      # the upload lands.
       - name: Attach the pip/uv wheel (dispatch release.yml)
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -293,12 +294,13 @@ jobs:
         run: |
           for _ in $(seq 1 30); do
             sleep 20
-            if gh release view "v${VERSION}" --json assets --jq '.assets[].name' \
-                 | grep -qx "keboola_cli-${VERSION}-py3-none-any.whl"; then
-              echo "wheel attached"; exit 0
+            names="$(gh release view "v${VERSION}" --json assets --jq '.assets[].name')"
+            if grep -qx "keboola_cli-${VERSION}-py3-none-any.whl" <<<"$names" \
+               && grep -qx "keboola_agent_cli-${VERSION}-py3-none-any.whl" <<<"$names"; then
+              echo "both wheels attached"; exit 0
             fi
           done
-          echo "::error::wheel still missing 10 min after dispatching release.yml for v${VERSION} -- check that run"
+          echo "::error::wheel(s) still missing 10 min after dispatching release.yml for v${VERSION} -- check that run"
           exit 1
 
   # ── Publish to S3 (cli-dist.keboola.com/keboola-cli2/) and index the apt/rpm/apk
--- a/.github/workflows/release-kbagent.yml
+++ b/.github/workflows/release-kbagent.yml
@@ -272,17 +272,34 @@ jobs:
           files: release/*
           prerelease: ${{ needs.version.outputs.IS_PRERELEASE == 'true' }}
       # Attach the pip/uv wheel too. release.yml builds + uploads it (the asset that
-      # auto-update and install.sh download) on `release: published` -- but a release
-      # created above with GITHUB_TOKEN does NOT emit that event, so on a bot-cut
-      # release the wheel never got attached (this is what broke v0.64.0).
-      # workflow_dispatch is the documented exception to the GITHUB_TOKEN no-cascade
-      # rule, so dispatch release.yml directly; it fires no matter who cut the release.
-      # The release already exists (created above), so release.yml's upload lands.
+      # auto-update and install.sh download) on `release: published`. A release this
+      # job creates with GITHUB_TOKEN does NOT emit that event, so when no one
+      # published the release by hand the wheel never got attached -- this broke
+      # v0.64.0. workflow_dispatch is the documented exception to the GITHUB_TOKEN
+      # no-trigger rule, so dispatch release.yml directly; it runs whoever created
+      # the release. The release already exists (created above), so the upload lands.
       - name: Attach the pip/uv wheel (dispatch release.yml)
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GH_REPO: ${{ github.repository }}
+          GH_REPO: ${{ github.repository }} # required: no checkout in this job, gh can't read the repo from git
         run: gh workflow run release.yml -f tag="v${VERSION}"
+      # gh workflow run is fire-and-forget: it returns once GitHub accepts the
+      # dispatch, not when release.yml finishes. Poll the release for the wheel so a
+      # failed release.yml turns THIS run red instead of silently shipping no wheel.
+      - name: Confirm the wheel was attached
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          for _ in $(seq 1 30); do
+            sleep 20
+            if gh release view "v${VERSION}" --json assets --jq '.assets[].name' \
+                 | grep -qx "keboola_cli-${VERSION}-py3-none-any.whl"; then
+              echo "wheel attached"; exit 0
+            fi
+          done
+          echo "::error::wheel still missing 10 min after dispatching release.yml for v${VERSION} -- check that run"
+          exit 1
 
   # ── Publish to S3 (cli-dist.keboola.com/keboola-cli2/) and index the apt/rpm/apk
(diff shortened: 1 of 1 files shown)


soustruh (commented):
# Review of #457 — fix(release): attach the pip/uv wheel on every release

## Summary

The PR adds one `actions: write` permission and one `gh workflow run release.yml -f tag=...` step to the `github-release` job in `release-kbagent.yml`. The root cause analysis in the description is correct — `GITHUB_TOKEN` does not emit `release: published` events, so bot-created releases (the normal path for this repo) never triggered the wheel workflow. The fix is minimal, uses the only documented escape hatch (`workflow_dispatch`), and the input field name and checkout logic in `release.yml` have been verified to line up correctly. One NON-BLOCKING observation and one NIT below; no blocking issues.

## Verdict

- **Verdict**: APPROVE
- **Blocking findings**: 0
- **Nits**: 1

## Blocking findings

(none)

## Non-blocking findings

### `[NB-1]` `.github/workflows/release-kbagent.yml:285` — dispatch is fire-and-forget; a `release.yml` failure is invisible to the release run

`gh workflow run` returns as soon as GitHub accepts the event (HTTP 204). If `release.yml` fails — build error, npm timeout, asset upload failure — `release-kbagent.yml` still reports green. The missing wheel is only noticed by users hitting the broken update path, exactly as happened with v0.64.0.

This is a well-known limitation of cross-workflow dispatch in Actions (there is no blocking `gh workflow run --wait` that returns the child run's exit code). The workaround is post-dispatch polling (`gh run list --workflow=release.yml --limit 1` until it exits), but that adds complexity most projects don't bother with. Noting it here so the failure mode is documented — if a release ships without the wheel again, start the investigation at the `release.yml` run triggered by the dispatch, not at `release-kbagent.yml` itself.

## Nits

- `[NIT-1]` `.github/workflows/release-kbagent.yml:282-284` — the step sets `GH_REPO: ${{ github.repository }}` explicitly, which mirrors what `gh` would use by default when running inside Actions. Not wrong, and it makes the target repo unambiguous if the workflow is ever called from a fork context; just noting it is redundant in the normal case.

## Verification log

- `gh pr view 457 --json title,body,files,additions,deletions,state` → 1 file, +13/-0, state OPEN, conventional `fix(release):` prefix matches a bug fix ✓
- `git rev-parse --abbrev-ref HEAD` → `fix/release-attach-wheel-every-release` matches `<branch>` input ✓
- `gh auth status` → authenticated as `soustruh` ✓
- `cat .github/workflows/release.yml` → confirms `workflow_dispatch` trigger with `inputs.tag` (string, required) — the dispatch call `-f tag="v${VERSION}"` uses the exact field name ✓
- `cat .github/workflows/release.yml` → `TAG: ${{ github.event.release.tag_name || inputs.tag }}` and `ref: ${{ env.TAG }}` in checkout — the dispatched run checks out the tagged commit, not default-branch HEAD ✓
- `release.yml` line 15-16 comment: "Fires for both normal and prerelease (beta) releases" — the dispatch firing unconditionally for pre-release tags (no `IS_PRERELEASE` guard on `github-release` job) is intentional and correct ✓
- `release.yml` line 125: `gh release upload "$TAG" dist/*.whl --clobber` — `--clobber` makes re-runs idempotent; no duplicate-asset error if dispatch fires twice ✓
- Double-run scenario checked: `release-kbagent.yml` triggers only on `push: tags`, not on `release:` events. A human-created release fires `release: published` → `release.yml` runs directly, `release-kbagent.yml` never starts. A bot-created release fires `push: tags` → `release-kbagent.yml` runs → dispatches `release.yml`; `release: published` is NOT emitted. No double-run ✓
- `grep -n "actions: write"` confirms the new permission is scoped to the `github-release` job only (not widened at workflow level) ✓
- `make check` → all tests passed, 0 lint errors ✓

(none)

soustruh (commented):
_

padak (commented):
# Review of #457 — fix(release): attach the pip/uv wheel on every release

## Summary

This PR fixes a real gap in the release pipeline: when `release-kbagent.yml` creates a GitHub Release via `GITHUB_TOKEN`, GitHub does not emit the `release: published` event, so `release.yml` (the wheel-build workflow) never fires. The fix adds an explicit `workflow_dispatch` of `release.yml` after the release is created, then polls until the wheel asset appears before declaring success. The approach is sound, well-commented, and minimal. All checks pass. No Python source files are changed, so the plugin synchronization map, permission registry, and layer-compliance checks are not applicable. The one finding below is a non-blocking correctness note worth the author's consideration.

## Verdict

- **Verdict**: APPROVE
- **Blocking findings**: 0
- **Nits**: 1

## Blocking findings

(none)

## Non-blocking findings

### `[NB-1]` `.github/workflows/release-kbagent.yml:37` (polling step) — confirmation step only checks the primary wheel, not the legacy compat wheel

The "Confirm the wheel was attached" step polls for `keboola_cli-${VERSION}-py3-none-any.whl`. `release.yml` actually uploads two wheels in a single `gh release upload "$TAG" dist/*.whl --clobber` call: `keboola_cli-*.whl` (primary) and `keboola_agent_cli-*.whl` (legacy compat bridge for <=0.62 clients, added in the `#424` migration). Because both are uploaded atomically in one step, a green polling result does imply both are present. However, if a future change to `release.yml` ever splits the upload into two steps and the first one (legacy compat) fails silently after the primary succeeds, the polling will pass while the compat wheel is missing — exactly the category of silent failure this PR is trying to prevent.

Consider checking for `keboola_agent_cli-${VERSION}-py3-none-any.whl` in the same polling grep (e.g., with an `&&`), or add a second `grep -qx` line, to make the confirmation step cover the full upload contract.

## Nits

- `[NIT-1]` `.github/workflows/release-kbagent.yml:30` (dispatch step comment) — "it runs whoever created the release" is a bit misleading. `workflow_dispatch` runs the workflow under the `GITHUB_TOKEN` of the dispatching run (i.e., `github-actions[bot]`), not under the identity of whoever created the release. The functional behavior is correct; the comment description is imprecise.

## Verification log

- `gh pr view 457 --json title,body,files,additions,deletions,baseRefName,headRefName,labels,state` → state OPEN, 1 file changed, +30/-0, conventional `fix(release):` prefix ✓
- `gh pr diff 457` → diff is 48 lines, touches only `.github/workflows/release-kbagent.yml` ✓
- `uv sync --extra server && make check` → 4156 passed, 8 skipped, 127 deselected, 13 warnings in 83.57s ✓ (exit 0)
- Layer violation grep (`typer` in services, `httpx` in commands, formatter in clients) → no Python source files changed; not applicable ✓
- Plugin synchronization map check (OPERATION_REGISTRY, AGENT_CONTEXT, CLAUDE.md commands, keboola-expert.md, commands-reference.md, gotchas.md) → no new CLI commands added; not applicable ✓
- Magic numbers / bare except / print() in production / raw error-code strings → no Python changes; not applicable ✓
- Token discipline: both new steps use `${{ secrets.GITHUB_TOKEN }}` via `GH_TOKEN` env var; no hardcoded tokens ✓
- `GH_REPO: ${{ github.repository }}` correctly set for both new steps because the `github-release` job has no `actions/checkout` step, so `gh` cannot infer the repo from git metadata ✓
- `gh workflow run release.yml -f tag="v${VERSION}"` targets `release.yml` on the default branch (main); `release.yml` IS on main and already has `workflow_dispatch.inputs.tag` defined ✓
- Polling math: 30 iterations × 20 seconds = 600 seconds = 10 minutes; error message says "10 min" ✓
- Pre-release behavior: `github-release` job has no `IS_PRERELEASE` guard, so the dispatch fires for beta tags too; `release.yml` explicitly handles betas ("Fires for both normal and prerelease (beta) releases") ✓
- Behavior reproduction: cannot execute against a real tag push without creating a release in the target repo; author's evidence (`v0.64.0` missing wheel + `v0.65.0` wheel present) is consistent with the stated root cause and the fix is mechanically correct ✓

(none)

padak (commented):
# Review of #457 — fix(release): attach the pip/uv wheel on every release

## Summary

This PR fixes a real gap in the release pipeline: when `release-kbagent.yml` creates a GitHub Release via `GITHUB_TOKEN`, GitHub does not emit the `release: published` event, so `release.yml` (the wheel-build workflow) never fires. The fix adds an explicit `workflow_dispatch` of `release.yml` after the release is created, then polls until the wheel asset appears before declaring success. The approach is sound, well-commented, and minimal. All checks pass. No Python source files are changed, so the plugin synchronization map, permission registry, and layer-compliance checks are not applicable. The one finding below is a non-blocking correctness note worth the author's consideration.

## Verdict

- **Verdict**: APPROVE
- **Blocking findings**: 0
- **Nits**: 1

## Blocking findings

(none)

## Non-blocking findings

### `[NB-1]` `.github/workflows/release-kbagent.yml:37` (polling step) — confirmation step only checks the primary wheel, not the legacy compat wheel

The "Confirm the wheel was attached" step polls for `keboola_cli-${VERSION}-py3-none-any.whl`. `release.yml` actually uploads two wheels in a single `gh release upload "$TAG" dist/*.whl --clobber` call: `keboola_cli-*.whl` (primary) and `keboola_agent_cli-*.whl` (legacy compat bridge for <=0.62 clients, added in the `#424` migration). Because both are uploaded atomically in one step, a green polling result does imply both are present. However, if a future change to `release.yml` ever splits the upload into two steps and the first one (legacy compat) fails silently after the primary succeeds, the polling will pass while the compat wheel is missing — exactly the category of silent failure this PR is trying to prevent.

Consider checking for `keboola_agent_cli-${VERSION}-py3-none-any.whl` in the same polling grep (e.g., with an `&&`), or add a second `grep -qx` line, to make the confirmation step cover the full upload contract.

## Nits

- `[NIT-1]` `.github/workflows/release-kbagent.yml:30` (dispatch step comment) — "it runs whoever created the release" is a bit misleading. `workflow_dispatch` runs the workflow under the `GITHUB_TOKEN` of the dispatching run (i.e., `github-actions[bot]`), not under the identity of whoever created the release. The functional behavior is correct; the comment description is imprecise.

## Verification log

- `gh pr view 457 --json title,body,files,additions,deletions,baseRefName,headRefName,labels,state` → state OPEN, 1 file changed, +30/-0, conventional `fix(release):` prefix ✓
- `gh pr diff 457` → diff is 48 lines, touches only `.github/workflows/release-kbagent.yml` ✓
- `uv sync --extra server && make check` → 4156 passed, 8 skipped, 127 deselected, 13 warnings in 83.57s ✓ (exit 0)
- Layer violation grep (`typer` in services, `httpx` in commands, formatter in clients) → no Python source files changed; not applicable ✓
- Plugin synchronization map check (OPERATION_REGISTRY, AGENT_CONTEXT, CLAUDE.md commands, keboola-expert.md, commands-reference.md, gotchas.md) → no new CLI commands added; not applicable ✓
- Magic numbers / bare except / print() in production / raw error-code strings → no Python changes; not applicable ✓
- Token discipline: both new steps use `${{ secrets.GITHUB_TOKEN }}` via `GH_TOKEN` env var; no hardcoded tokens ✓
- `GH_REPO: ${{ github.repository }}` correctly set for both new steps because the `github-release` job has no `actions/checkout` step, so `gh` cannot infer the repo from git metadata ✓
- `gh workflow run release.yml -f tag="v${VERSION}"` targets `release.yml` on the default branch (main); `release.yml` IS on main and already has `workflow_dispatch.inputs.tag` defined ✓
- Polling math: 30 iterations × 20 seconds = 600 seconds = 10 minutes; error message says "10 min" ✓
- Pre-release behavior: `github-release` job has no `IS_PRERELEASE` guard, so the dispatch fires for beta tags too; `release.yml` explicitly handles betas ("Fires for both normal and prerelease (beta) releases") ✓
- Behavior reproduction: cannot execute against a real tag push without creating a release in the target repo; author's evidence (`v0.64.0` missing wheel + `v0.65.0` wheel present) is consistent with the stated root cause and the fix is mechanically correct ✓

(none)

padak (approved):
Approving. The automated review pass found no blocking issues — 0 blocking, 1 non-blocking, 1 nit (details in the earlier review comment).

The change correctly attaches the prebuilt `py3-none-any` wheel to every GitHub Release, so `kbagent update` can pull a built asset instead of rebuilding the SPA from source — the path that timed out on WSL / slow networks.

One non-blocking follow-up worth tracking: the verify step polls only the primary `keboola_cli-*.whl`, not the legacy `keboola_agent_cli-*.whl`. Harmless today because `release.yml` uploads both through a single `dist/*.whl` glob (atomic), but a second `grep -qx` would keep it safe if that upload is ever split into two steps.

LGTM.

PR #461, opened 2026-06-23
fix(build): npm ci --ignore-scripts in the wheel build hook (GHSA-pfvh, code-only)

padak:
## Summary

Fixes **M9** from the 2026-06-12 security audit (private advisory **GHSA-pfvh-6q5w-9m52**) — an install-time RCE surface in the wheel build hook.

`scripts/hatch_build.py` runs `npm ci` to build the React SPA at wheel-build time **and** when a user installs via `git+https://github.com/keboola/cli` (the fallback when no prebuilt wheel asset is present). Without `--ignore-scripts`, every `preinstall`/`install`/`postinstall` lifecycle script in the (transitive) npm dependency tree executes on the user's machine during install — so a single compromised transitive dependency lands arbitrary code execution on every user who takes the `git+` install path with Node on PATH.

## Fix

Add `--ignore-scripts` to the `npm ci` invocation. `tsc -b && vite build` compiles the SPA without needing any dependency install-script.

### Verified non-breaking (not assumed)
`--ignore-scripts` can break builds whose deps fetch a binary via `postinstall` (the classic esbuild trap). I **ran the real build** with the flag in `web/frontend`:
```
npm ci --ignore-scripts --prefer-offline --no-audit --no-fund   # added 393 packages, exit 0
npm run build                                                    # tsc -b && vite build -> ✓ built, dist/index.html present
```
vite 8 uses **rolldown** (Rust bundler) which ships prebuilt platform binaries as regular package files, not via a postinstall download — so `--ignore-scripts` is safe here. The lockfile (`package-lock.json`) remains committed + integrity-pinned, so this is defense-in-depth on top of the existing tarball-hash verification.

## Tests

`tests/test_build_hook.py::test_passes_resolved_npm_path_not_bare_name` now also asserts `--ignore-scripts` is in the `npm ci` command (mock-based regression test). Full suite green: **4177 passed, 135 skipped**; lint/format clean.

## ⚠️ Code-only PR
Touches only `scripts/hatch_build.py` + `tests/test_build_hook.py` — **no version bump / changelog entry** (added at the next release, same conflict-immunity rationale as #422/#460).

## Audit progress
Open after this: M7 (plaintext-on-encrypt warning — 3 services), M10 (version regex), + M1/M5 residuals.

---

<a href="https://app.devin.ai/review/keboola/cli/pull/461" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/scripts/hatch_build.py
+++ b/scripts/hatch_build.py
@@ -138,7 +138,19 @@ def _bundle_ui(repo_root: Path, log: Callable[[str], None] = print) -> None:
     log("no prebuilt dist found; running npm build")
     try:
         subprocess.check_call(
-            [npm, "ci", "--prefer-offline", "--no-audit", "--no-fund"],
+            # --ignore-scripts: do NOT run npm lifecycle scripts from the
+            # (transitive) dependency tree at wheel-build / `git+` install time.
+            # Without it, a compromised dependency's postinstall would execute
+            # arbitrary code on the user's machine during install (GHSA-pfvh).
+            # Verified non-breaking: `tsc -b && vite build` needs no dependency
+            # install-scripts -- vite 8 (rolldown) ships prebuilt platform
+            # binaries rather than fetching them via a postinstall.
+            # NOTE: this also suppresses any *root* web/frontend/package.json
+            # prepare/postinstall script (there is none today). If one is ever
+            # added AND required for the build, invoke it explicitly here rather
+            # than dropping --ignore-scripts, which would re-open the dependency
+            # lifecycle-script RCE surface.
+            [npm, "ci", "--prefer-offline", "--no-audit", "--no-fund", "--ignore-scripts"],
             cwd=frontend_dir,
         )
         subprocess.check_call([npm, "run", "build"], cwd=frontend_dir)
--- a/tests/test_build_hook.py
+++ b/tests/test_build_hook.py
@@ -187,8 +187,10 @@ class TestBundleUiBug1NpmInvocation:
         assert _target(root).is_dir()
 
     def test_passes_resolved_npm_path_not_bare_name(self, tmp_path: Path) -> None:
-        """Root-cause fix: the resolved ``shutil.which`` path (``npm.cmd`` on
-        Windows) is handed to subprocess, never the bare string ``"npm"``."""
+        """Two regressions guarded here: (1) the resolved ``shutil.which`` path
+        (``npm.cmd`` on Windows) is handed to subprocess, never the bare string
+        ``"npm"``; (2) ``npm ci`` carries ``--ignore-scripts`` (GHSA-pfvh) while
+        ``npm run build`` intentionally does NOT."""
         root = _make_repo(tmp_path)
         resolved = "C:\\Program Files\\nodejs\\npm.cmd"
         with (
@@ -203,7 +205,15 @@ class TestBundleUiBug1NpmInvocation:
         assert ci_cmd[0] == resolved
         assert build_cmd[0] == resolved
         assert ci_cmd[1] == "ci"
+        # GHSA-pfvh: dependency lifecycle scripts must NOT run at wheel-build /
+        # `git+` install time (install-time RCE surface). Verified non-breaking
+        # against the real `tsc -b && vite build`.
+        assert "--ignore-scripts" in ci_cmd
+        # Intentional asymmetry: `--ignore-scripts` gates the dependency INSTALL
+        # (`npm ci`). `npm run build` runs our OWN build script and must NOT
+        # carry the flag -- do not "fix" the asymmetry by adding it there.
         assert build_cmd[1:] == ["run", "build"]
+        assert "--ignore-scripts" not in build_cmd
 
     def test_successful_npm_build_is_bundled(self, tmp_path: Path) -> None:
         """When npm produces dist/index.html, it gets copied into _ui_dist/."""


padak (commented):
# Review of #461 — fix(build): npm ci --ignore-scripts in the wheel build hook (GHSA-pfvh, code-only)

## Summary

This PR adds `--ignore-scripts` to the `npm ci` invocation in `scripts/hatch_build.py` to close a real install-time RCE surface (GHSA-pfvh-6q5w-9m52): without the flag, any `postinstall`/`preinstall` lifecycle script in the transitive npm dependency tree executes on the user's machine during `git+` install — including on machines that have never consciously run `npm`. The fix is one-line surgical; the test assertion (`assert "--ignore-scripts" in ci_cmd`) is correctly scoped to the `npm ci` call only and does not incorrectly constrain the `npm run build` call. The deliberate deferral of the version bump and changelog entry is noted in the PR description and is consistent with the documented conflict-immunity pattern used in #422 and #460.

**Verdict: APPROVE.** No blocking findings. Two non-blocking observations and one nit follow.

## Verdict

- **Verdict**: APPROVE
- **Blocking findings**: 0
- **Nits**: 1

## Blocking findings

(none)

## Non-blocking findings

### `[NB-1]` `tests/test_build_hook.py:209` — test does not assert that `--ignore-scripts` is absent from the `npm run build` command

The assertion at line 210 (`assert build_cmd[1:] == ["run", "build"]`) implicitly confirms `--ignore-scripts` is not in the `build_cmd`, but it does so as a positive equality check on the full tail rather than an explicit negative assertion. This is correct and sufficient — `npm run` ignores the flag anyway — but a future maintainer reading the test may not realize the `build_cmd` shape is intentionally different from `ci_cmd`. A one-line comment like `# --ignore-scripts intentionally absent: npm run executes a named script, not lifecycle hooks` above the `build_cmd[1:]` assertion would make the asymmetry self-documenting and prevent a well-meaning contributor from "fixing" the apparent inconsistency.

### `[NB-2]` `scripts/hatch_build.py:148` — `--ignore-scripts` suppresses the root package's own `prepare` script if one is ever added

`npm ci --ignore-scripts` suppresses ALL lifecycle scripts, including the root package's `prepare` entry. Currently `web/frontend/package.json` has no `prepare` script so this is benign. However, if a future contributor adds `"prepare": "some-setup-tool"` to the root `package.json` (a common pattern for husky, lefthook, etc.), the flag will silently suppress it during wheel builds without any visible error. The code comment at line 141–147 explains the security rationale well, but does not call out this constraint. Adding "Note: if `web/frontend/package.json` ever gains a `prepare` script, verify it is not needed at wheel-build time" would surface the constraint before it bites.

## Nits

- `[NIT-1]` `tests/test_build_hook.py:189` — the docstring of `test_passes_resolved_npm_path_not_bare_name` still describes only the "Bug 1" Windows `.cmd` fix. Now that this test also covers the GHSA-pfvh regression, expanding the docstring to mention both responsibilities (or splitting into two focused tests) would keep the test naming honest.

## Verification log

- `gh pr view 461 --json title,body,files,additions,deletions,state` → state=OPEN, 2 files (+12/-1), conventional `fix(build):` prefix ✓
- `git rev-parse --abbrev-ref HEAD` → `fix/build-ignore-scripts` matches PR branch ✓
- `git show 7cec496 --stat` → only `scripts/hatch_build.py` (+9/-1) and `tests/test_build_hook.py` (+4/0) touched; no `src/` changes ✓ (Plugin synchronization map: no CLI command surface change, no OPERATION_REGISTRY / context.py / CLAUDE.md / keboola-expert.md updates required)
- Layer violation checks (typer in services, httpx in commands, formatter in clients): `grep` returned empty ✓
- Convention checks (magic numbers, raw error_code strings, bare except, print() in src/, token in logs): all empty ✓
- `python3` scan of `web/frontend/package-lock.json` (lockfileVersion=3, 393 packages): **0 packages with lifecycle scripts** (`postinstall`/`install`/`preinstall`/`prepare`) ✓ — confirms `--ignore-scripts` has nothing to suppress in the current lock; the security value is defense-in-depth against a future compromised dependency
- `web/frontend/package.json` root scripts: `{dev, build, preview, test}` — no `prepare`/`postinstall` entries ✓
- No `esbuild` entries in lock file (vite 8 / rolldown does not depend on esbuild's `postinstall` binary fetcher) ✓ — the classic `--ignore-scripts` breakage vector is absent
- No git-hook managers (`husky`, `lefthook`, `simple-git-hooks`) in dependencies ✓
- `uv run pytest tests/test_build_hook.py::TestBundleUiBug1NpmInvocation::test_passes_resolved_npm_path_not_bare_name -v` → 1 passed ✓
- `make check` → **4177 passed, 8 skipped**, lint/format/typecheck/skill/version/error-codes clean ✓
- Behavior verification: the PR author ran the real build with the flag (`npm ci --ignore-scripts ... && npm run build` → `dist/index.html` present); the lock file scan above independently corroborates that no lifecycle script would have fired anyway; cannot re-run the full build in this reviewer context (no npm in CI sandbox), but the evidence is sufficient

(none)

padak:
## Author response — review findings addressed (commit 35cb8b5, still code-only)

Thanks — and for the empirical confirmation that `package-lock.json` (393 pkgs) carries zero lifecycle scripts today, so this is pure defense-in-depth. All three findings were "make the intentional choice discoverable so a future maintainer doesn't undo it", and I've added that discoverability:

- **NB-1 — undocumented `ci` vs `build` asymmetry in the test:** Added an inline comment at the assertions explaining that `--ignore-scripts` intentionally gates only the dependency *install* (`npm ci`), and `npm run build` (our own build script) must NOT carry it. Also turned that into a tested invariant: `assert "--ignore-scripts" not in build_cmd`.

- **NB-2 — `--ignore-scripts` would also suppress a root `prepare`/`postinstall`:** Added a NOTE in the `hatch_build.py` comment: there is no root build script today; if one is ever added AND required for the build, it must be invoked explicitly rather than dropping `--ignore-scripts` (which would re-open the dep-tree RCE surface).

- **NIT-1 — stale test docstring:** Updated to state both regressions it now guards (the Windows `npm.cmd` path resolution AND the GHSA-pfvh `--ignore-scripts` flag).

No logic change. Still **code-only** (version + changelog deferred to the next release alongside #422 / #460). `make check` green; `tests/test_build_hook.py` 20 passed.

PR #476, opened 2026-07-03
fix(release): unbreak homebrew + chocolatey publish jobs

padak:
Two pre-existing bugs in `release-kbagent` failed the **brew** + **choco** pushes on the v0.66.0 run (they surfaced now because the macOS *Sign & notarize* step — previously blocked on an expired Apple Developer agreement — is green again, so the downstream package-manager jobs finally ran):

- **chocolatey** — the `.sha256` asset is served as `application/octet-stream`, so `pwsh` returns `Invoke-WebRequest`'s `.Content` as a `byte[]` and `.Split(" ")` threw `[System.Byte] does not contain a method named 'Split'`. Now downloads it via `-OutFile` and reads it back as text, splitting on any whitespace.
- **homebrew** — `target-directory: "."` made the push-to-tap action run `rm` against `.`/`..` (`can't remove '.' or '..'`), aborting the job. Now pushes the rendered formula into the tap's `Formula/` dir (same resulting layout `Formula/keboola-cli2.rb`); keeps the SHA-pinned action (no supply-chain change).
- **winget** (`needs: chocolatey`) was skipped as a consequence and unblocks once choco is green.

No app/runtime code changed — workflow-only. Applies to the next tag; see PR discussion for backfilling v0.66.x to brew/choco.

---

<a href="https://app.devin.ai/review/keboola/cli/pull/476" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.github/workflows/release-kbagent.yml
+++ b/.github/workflows/release-kbagent.yml
@@ -383,8 +383,13 @@ jobs:
         env:
           API_TOKEN_GITHUB: ${{ secrets.HOMEBREW_TAP_TOKEN }}
         with:
-          source-directory: "out"
-          target-directory: "."
+          # Copy the rendered formula into the tap's `Formula/` dir (its canonical
+          # location) rather than the repo root. target-directory "." made the
+          # push action run `rm` against `.`/`..` while clearing the target, which
+          # fails fatally ("can't remove '.' or '..'"); a real subdirectory clears
+          # cleanly. The resulting tap layout (`Formula/keboola-cli2.rb`) is unchanged.
+          source-directory: "out/Formula"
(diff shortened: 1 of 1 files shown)


padak:
Re Devin's note on `target-directory: Formula`: confirmed intended. The old `target: "."` never actually cleared anything — it crashed on `rm '.'` — and had it worked it would have wiped the tap's root `README.md`. The tap `keboola/homebrew-keboola-cli2` has `README.md` at root and the formula belongs at `Formula/keboola-cli2.rb`, so scoping the clear to `Formula/` (preserving the README) is the correct behavior.

Issue #267, opened 2026-05-07
sync git-branching: branch_id type confusion + walker scope guard block GitOps workflow


padak:
## TL;DR

Two chained bugs in `kbagent sync` make GitOps workflows unusable on `gitBranching.enabled: true` workspaces:

- **Bug A** — `kbagent sync pull` mis-routes dev-branch configs into `main/` and corrupts `.keboola/manifest.json` by appending a duplicate `branches[]` entry on every invocation. Root cause: Keboola branch IDs are persisted as `str` in `branch-mapping.json` but compared as `int` everywhere else, so every cross-type `==` lookup silently fails.

- **Bug B** — `kbagent sync diff` and `kbagent sync push --dry-run` silently report "no changes" when the user scaffolds a brand-new local config dir under a non-default branch path. The "scaffold locally then push" flow advertised by the docs is unreachable on git-branching workspaces.

The two bugs are independent at the symptom level but **Bug A blocks a complete fix for Bug B** (the path-resolution helper that Bug B's walker delegates to is itself broken by Bug A). They should be fixed together in one PR.

Reported externally against **v0.27.0**; reproduced in code review on **v0.30.3** (current `main`). Both bugs are still latent.

---

## Shared environment

- Manifest schema **v3**, `gitBranching.enabled: true`, `defaultBranch: main`.
- Active git feature branch linked via `kbagent sync branch-link --branch <DEV_BRANCH_ID>` to a non-default Keboola dev branch.
- Manifest pre-state:
  ```json
  "branches": [
    { "id": 38516, "path": "main",         "metadata": {} },
    { "id": 38721, "path": "branch-38721", "metadata": {} }
  ]
  ```

---

### Symptom

Single warning line on stderr:

```
WARNING Branch ID 38721 not found in manifest, falling back to default path
```

— even though `{"id": 38721, "path": "branch-38721"}` is plainly present in `manifest.branches[]`. The lookup fails, the fallback writes to `main/`, and a separate code path then re-registers the "missing" branch under a mangled path.

### Repro

```bash
kbagent --json sync pull --project <alias>
```

1. The warning above is emitted.
2. JSON response indicates a successful pull, but `branch_dir: "main"`:
   ```json
   { "status": "ok",
     "data": {
       "status": "pulled",
       "branch_id": "38721",
       "branch_dir": "main",
       "configs_pulled": 1,
       "details": [
         { "action": "updated",
           "component_id": "<component-id>",
[...]

PR #494, opened 2026-07-20
fix(sync): sync push no longer duplicates configs inherited from main on a dev branch

### LINKED ISSUE #482: sync push creates duplicate configs for every config a dev branch inherits from main (even with zero local changes)
# `sync push` creates duplicate configs for every config a dev branch inherits from main (even with zero local changes)

## Summary

On a Keboola dev branch, `sync push` treats every config **inherited from main** (not materialized branch-locally) as `added` and **creates a duplicate**, even when there are **no local edits at all**. A single `sync push` on a freshly-pulled branch created **100 duplicate configurations**. This makes `sync push` unsafe on any dev branch that inherits the bulk of its configs from main — i.e. the normal case for a small targeted change.

## Environment

- kbagent **v0.66.0**
- Stack: `north-europe.azure`
- Single-project workflow, `branch use` dev branches (not git-branching)
- Project has the `storage-branches` feature (real dev branches, not legacy fake-branch)

## Steps to reproduce

Starting from a completely clean local state (to rule out a corrupted manifest):

```bash
# 1. Clean slate + pull main
rm -rf main <branch_dir> storage .keboola/manifest.json
kbagent sync init --project <alias>
kbagent sync pull --project <alias>          # 66 configs, branchId 25376
kbagent sync diff --project <alias>          # 0/0/66 unchanged  ✅ clean baseline

# 2. Switch to an existing dev branch + pull (NO edits after this)
kbagent branch use --branch <dev_id> --project <alias>
kbagent sync pull --project <alias>          # 68 configs into <branch_dir>/

# 3. Push with ZERO local changes — should be a no-op
kbagent --json sync push --project <alias>
```

## Expected

With no local edits after a fresh pull, `sync push` should be a **no-op** (`created: 0`), exactly like the same sequence on `main`.

## Actual

```json
{ "status": "pushed", "created": 100, "updated": 0, "deleted": 0,
  "errors": [ /* 3× keboola.orchestrator: "Creation of new Flows disabled" */ ] }
```

- **`created: 100`** — new duplicate configs, with no local changes.
- The only configs *not* duplicated were 3 `keboola.orchestrator` flows, blocked by an unrelated API permission ("Creation of new Flows disabled") — otherwise it would have been 103.
- Concrete example: `keboola.app-data-gateway` had **1** config on the branch before push, **3** after (2 duplicates, same name `Data Gateway`, fresh ULIDs).

Dry-run reveals the same intent beforehand — 66 of 68 configs reported `change_type: added` with **empty `config_id`**, despite each having a valid ULID under `branchId: <dev_id>` in `.keboola/manifest.json`.

## Proof the configs already exist on the branch

```bash
kbagent config detail --component-id keboola.app-data-gateway \
  --config-id 01kc4xss8veqhgc92276scekpg --branch <dev_id> --project <alias>
# -> {id: 01kc4xss8veqhgc92276scekpg, name: "Data Gateway", version: 1,
#     last_change: "Copied from default branch configuration ... version 35"}
```

The config resolves on the dev branch with the **same ID** as main. `sync push` still creates a duplicate.

## Root cause (hypothesis)

The push diff engine decides "exists on remote branch?" from a branch-scoped listing that returns only **branch-materialized** configs, ignoring configs read-through from main. This is the documented Storage-API behaviour for buckets (kbagent's own `gotchas.md`: *"`storage buckets --branch ID` returns only locally-modified buckets ... That is Storage API behaviour, not a CLI bug"*). But `sync push` mishandles it: instead of treating an inherited config as read-through/unchanged, it classifies it as `added`, drops its manifest ULID, and POSTs a new config — producing duplicates. `sync pull` writes correct ULIDs for all 68 configs, so the pull/push pair is asymmetric.

## Impact

- **Data-corrupting.** One `sync push` on a partially-materialized branch silently creates N duplicate configs (100 here). Cleanup is hard: the push result does not reliably record the new ULIDs, so the only clean recovery was to delete the entire dev branch.
- Affects the most common workflow: create/checkout a dev branch, change one transformation, push.
- `--dry-run` surfaces the `added` entries, but there is no scoping flag to push only genuinely-changed configs, and the danger (duplicate creation vs. harmless no-op) is not obvious from the dry-run output.

## Suggested fix

`sync push` should treat a local config whose manifest ULID resolves on the target branch — including configs inherited/read-through from main — as `unchanged` when content matches, rather than `added`. The remote-existence check must account for branch inheritance, consistent with what `sync pull` and `config detail --branch` already do. At minimum, push should refuse to create a config whose manifest ULID already exists on the target branch, and warn loudly instead of silently duplicating.

## Workaround

For single-config edits on such a branch, use the MCP tool call (`update_config` / `update_sql_transformation`), which targets the config ID on the active branch and lets KBC resolve inheritance server-side. Avoid `sync push` on any branch where `sync push --dry-run` reports `added` for configs that already exist on the branch.

padak:
Fixes #482

## Root cause

Not the Storage API and not the branch listing — `sync pull` on a dev branch receives **all** inherited configs from `GET /branch/{id}/components` (that is exactly how pull materialized all 68 configs in the report). The bug is local manifest bookkeeping:

1. `sync pull` **replaces** `manifest.configurations` wholesale with the pulled branch's entries (`sync_service.py`, end of `pull()`). After `branch use <dev>` + `pull`, the previously pulled `main/` tree is still on disk but no longer tracked by the manifest.
2. `_find_untracked_configs` keeps the **default branch tree in scope unconditionally** (a deliberate choice from the issue #267 scaffold flow). Every orphaned `main/` file is therefore reported as untracked.
3. `diff()` turned each untracked file into an `added` change with **empty `config_id`** — matching the dry-run signature in the report ("66 of 68 configs `added` with empty `config_id`") — and `push` then POSTed a brand-new config **onto the dev branch** for each of them, reading the file content from the colliding path in the dev tree.

## Fix

- **Scope the untracked scan to the source branch tree.** `diff`/`push` now pass the resolved source branch path (`_resolve_source_branch_path` — the tree push actually reads from) to `_find_untracked_configs`, which scans exactly that subtree. Files belonging to another branch's tree can never surface as `added` for the target branch. The issue #267 "scaffold locally then push" flow and the KFR-07 promote-default-tree flow (`push --branch <id>` with no per-branch subtree) both keep working, because they are defined by the same source-path resolution.
- **Adopt-by-id guard** (the "at minimum" ask from the issue): an untracked `_config.yml` whose `_keboola.config_id` resolves on the target branch and is **not claimed** by any manifest entry now diffs against the existing remote config (`unchanged`/`modified`) instead of being created as a duplicate. A **claimed** id (user copied a tracked config dir to fork it) keeps the fresh-create semantics, so the fork-by-copy workflow is unchanged, and `sync clone` is unaffected (its tree is manifest-tracked and requires a fresh target).
- `sync status` scanning is unchanged (it has no branch/remote context; orphaned trees still show there as local-only `added` files, which push now correctly ignores).

## Tests

- `TestIssue482BranchSwitchDuplicates::test_push_on_dev_branch_after_pull_is_noop` — the exact repro from the issue (pull main → pull dev → push with zero edits). Fails on pre-fix `main` with 2 phantom `added` changes (empty `config_id`, one per orphaned config); now asserts diff is empty, push is `no_changes`, and `create_config` is never called.
- Adopt-by-id: untracked file with a known unclaimed remote id → `modified` + `update_config`, never `create_config`.
- Fork-by-copy: copy of a tracked config dir (claimed id) → still a single `added` with empty `config_id` and a fresh create.
- The two issue #267 Bug B walker tests updated to the new `only_branch_path` parameter (same scenarios, expressed through the source-path resolution that diff/push now perform).

`make check`: 4317 passed, 8 skipped.

---

<a href="https://app.devin.ai/review/keboola/cli/pull/494" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/keboola_agent_cli/services/sync_service.py
+++ b/src/keboola_agent_cli/services/sync_service.py
@@ -1225,26 +1225,48 @@ class SyncService(BaseService):
                 }
             )
 
-        # Also add untracked local configs (new files)
-        for added_cfg in self._find_untracked_configs(project_root, manifest, branch_id):
-            branch_path = self._resolve_source_branch_path(manifest, project_root, branch_id)
-            config_dir = project_root / branch_path / added_cfg["path"]
+        # Build set of manifest-tracked keys so that compute_changeset only
+        # flags configs that were previously pulled (not brand-new remote ones).
+        tracked_keys = {f"{cfg.component_id}/{cfg.id}" for cfg in manifest.configurations}
+
+        # Also add untracked local configs (new files). Scan ONLY the branch
+        # subtree push reads from -- scanning another branch's tree (e.g.
+        # ``main/`` while targeting a dev branch) turned configs orphaned by a
+        # branch switch into phantom "added" entries, and push then created a
+        # duplicate on the target branch for each of them (issue #482).
+        source_branch_path = self._resolve_source_branch_path(manifest, project_root, branch_id)
+        for added_cfg in self._find_untracked_configs(
+            project_root, manifest, only_branch_path=source_branch_path
+        ):
+            config_dir = project_root / source_branch_path / added_cfg["path"]
             local_data = self._read_config_file(config_dir)
             if local_data is None:
                 continue
+            component_id = added_cfg.get("component_id", "unknown")
+            merge_code_files(component_id, local_data, config_dir)
+            # Adopt-by-id guard (issue #482): an untracked file whose
+            # ``_keboola.config_id`` resolves on the target branch and is not
+            # claimed by any manifest entry refers to an EXISTING remote
+            # config -- diff against it (unchanged/modified) instead of
+            # letting push create a duplicate. A claimed id means the user
+            # copied a tracked config dir to fork it: keep the create.
+            untracked_id = added_cfg.get("config_id", "")
+            untracked_key = f"{component_id}/{untracked_id}"
+            if not (
+                untracked_id
+                and untracked_key in remote_configs
+                and untracked_key not in tracked_keys
+            ):
+                untracked_id = ""  # new config, push creates it
             local_configs.append(
                 {
-                    "component_id": added_cfg.get("component_id", "unknown"),
-                    "config_id": "",  # new config, no ID yet
+                    "component_id": component_id,
+                    "config_id": untracked_id,
                     "config_name": local_data.get("name", ""),
                     "path": added_cfg["path"],
                     "data": local_data,
                 }
             )
-
-        # Build set of manifest-tracked keys so that compute_changeset only
-        # flags configs that were previously pulled (not brand-new remote ones).
-        tracked_keys = {f"{cfg.component_id}/{cfg.id}" for cfg in manifest.configurations}
 
         # Build base hashes for 3-way diff.
         # Preferred: pull_config_hash (normalized hash stored at pull time).
@@ -2141,20 +2163,28 @@ class SyncService(BaseService):
         self,
         project_root: Path,
         manifest: Manifest,
-        resolved_branch_id: int | None = None,
+        only_branch_path: str | None = None,
     ) -> list[dict[str, str]]:
         """Scan for _config.yml files that are not tracked in the manifest.
 
-        Scans branch directories that the user is actively working with:
-        branches that already have tracked configs, the default branch
-        (production), and the branch the caller resolved for the current
-        operation (when provided). This supports the documented
-        "scaffold locally then push" workflow on git-branching workspaces
-        with empty ``manifest.configurations`` (issue #267, Bug B).
+        When ``only_branch_path`` is given, scan exactly that branch subtree.
+        ``diff`` / ``push`` pass the resolved *source* branch path (the tree
+        push reads configs from) so that files belonging to a different
+        branch's tree can never be classified as "added" for the target
+        branch: after a branch switch, ``sync pull`` re-targets
+        ``manifest.configurations`` to the new branch, orphaning the previous
+        branch's tree on disk -- and every orphaned file used to surface as
+        "added", making ``sync push`` create a duplicate config on the target
+        branch for each of them (issue #482).
 
-        Branches outside this scope are skipped to avoid phantom "added"
-        configs from orphaned dev-branch directories left over from
-        previous work.
+        Without ``only_branch_path`` (``sync status``), scan branch
+        directories the user is actively working with: branches that already
+        have tracked configs and the default branch (production). The default
+        branch is always in scope so the documented "scaffold locally then
+        push" workflow works on workspaces with empty
+        ``manifest.configurations`` (issue #267, Bug B). Branches outside
+        this scope are skipped to avoid phantom "added" configs from
+        orphaned dev-branch directories left over from previous work.
         """
         tracked_paths: set[str] = set()
         in_scope_branch_ids: set[int] = set()
@@ -2168,14 +2198,12 @@ class SyncService(BaseService):
         if manifest.branches:
             in_scope_branch_ids.add(manifest.branches[0].id)
 
-        # The branch the caller resolved is in scope (linked feature branch
-        # the user explicitly switched to via git checkout + branch-link).
-        if resolved_branch_id is not None:
-            in_scope_branch_ids.add(resolved_branch_id)
-
         added: list[dict[str, str]] = []
         for branch in manifest.branches:
-            if branch.id not in in_scope_branch_ids:
+            if only_branch_path is not None:
+                if branch.path != only_branch_path:
+                    continue
+            elif branch.id not in in_scope_branch_ids:
                 continue
             branch_dir = project_root / branch.path
             if not branch_dir.exists():
--- a/tests/test_sync_service.py
+++ b/tests/test_sync_service.py
@@ -2583,7 +2583,8 @@ class TestIssue267Regressions:
     def test_walker_finds_untracked_with_empty_configurations(self, tmp_path: Path) -> None:
         """``_find_untracked_configs`` surfaces a scaffold under a non-default
         branch even when ``manifest.configurations: []``, given that the
-        branch is the resolved one for this op (Bug B fix)."""
+        branch is the resolved *source* tree for this op (Bug B fix; the
+        walker takes the source branch path since issue #482)."""
         project_root = tmp_path / "project"
         project_root.mkdir()
 
@@ -2630,10 +2631,12 @@ class TestIssue267Regressions:
 
         manifest = load_manifest(project_root)
         svc = SyncService(config_store=MagicMock())
+        # diff/push resolve the source tree first (the scaffold materializes
+        # branch-99999) and pass its path to the walker.
         added = svc._find_untracked_configs(
             project_root,
             manifest,
-            resolved_branch_id=99999,
+            only_branch_path="branch-99999",
         )
 
         assert len(added) == 1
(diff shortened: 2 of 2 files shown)


padak (commented):
# Review of #494 — fix(sync): sync push no longer duplicates configs inherited from main on a dev branch

## Summary

This PR fixes issue #482: after `branch use <dev>` + `sync pull`, the previously-pulled `main/` tree became orphaned in the manifest, and `sync push` treated every orphaned file as a brand-new config, duplicating up to ~100 configs onto the dev branch. The fix (a) scopes `_find_untracked_configs` to the single resolved *source* branch path instead of a set of "in-scope" branch ids, and (b) adds an adopt-by-id guard so an untracked file whose `_keboola.config_id` matches an existing, unclaimed remote config is diffed as `modified` rather than created as a duplicate. The change is service-layer only (no new/removed/renamed CLI command), comes with three targeted regression tests that directly encode the reported repro, and `make check` is fully green. **Verdict: COMMENT** — no blocking issues found; a few non-blocking follow-ups worth tracking (manifest reconciliation gap for adopted configs, deferred gotcha/E2E documentation).

## Verdict

- **Verdict**: COMMENT
- **Blocking findings**: 0
- **Nits**: 1

## Blocking findings

(none)

## Non-blocking findings

### `[NB-1]` `src/keboola_agent_cli/services/sync_service.py:1580-1583` — adopted config is never written back into the manifest

When an untracked file is "adopted" via the new id-matching guard (`sync_service.py:1253-1264`) and pushed as `modified`, `push_update()` succeeds, but the post-update loop at `sync_service.py:1580-1583` (`for cfg in manifest.configurations: if cfg.component_id == component_id and cfg.id == config_id`) finds nothing, because the adopted config was never added to `manifest.configurations`. The file therefore never converges into a normal tracked state: it stays permanently outside the manifest, so a later local deletion of that file is never detected as `deleted` (`tracked_keys` never contains it), and every future `sync diff`/`push` re-derives the adopt decision from scratch instead of reading a stable manifest entry. Consider adding a manifest writeback (similar to `writeback_create_config_in_manifest`) for the adopted-by-id path so the reconciliation is durable, not re-computed every run.

### `[NB-2]` `plugins/kbagent/skills/kbagent/references/gotchas.md` / `plugins/kbagent/agents/keboola-expert.md` §3 — no gotcha entry for the pre-fix duplication risk

Given the severity of the underlying bug (up to ~100 duplicate configs created from a single push after a branch switch), this is exactly the class of behavior CONTRIBUTING.md's "Plugin synchronization map" wants tagged `(since vX.Y.Z)` once a version is cut, so AI agents on older kbagent installs following the §4.5 "Cross-project migration" workflow (`branch use` → `pull` → edit → `push --dry-run`) know to upgrade first. Not flagging BLOCKING here because the three most recent merged PRs on `main` (#486, #487, #488) show the team currently batches these hand-maintained doc updates at release time rather than per-PR (no version bump lands with this PR either — `pyproject.toml` stays at `0.70.0`). Recommend adding the gotcha in the next release's manual-review pass (CONTRIBUTING.md "Releasing a new version" steps 5-6), tagged with whatever version this fix ships in.

### `[NB-3]` `tests/test_e2e.py` — no E2E test for the specific branch-switch-then-push regression

`test_sync_diff_branch_override` (`tests/test_e2e.py:10643`) exercises `sync diff --branch <id>` but doesn't reproduce the reported sequence (`pull` main → `branch use`/`pull` dev → `push` with zero edits, asserting `no_changes`/no duplicate `create_config` calls). This isn't a "new command" so it doesn't strictly trigger CONTRIBUTING.md's per-command E2E checklist, but given the real-world impact (a live-project bug that created ~100 duplicate configs), an E2E test against a real disposable branch would catch anything the fully-mocked service-layer tests can't (real manifest serialization round-trip, real branch-id typing, real API duplicate-detection). Every CLI command must have E2E coverage per CONTRIBUTING.md, but environmental constraints can defer one cycle — flagging for awareness, not blocking.

## Nits

- `[NIT-1]` `src/keboola_agent_cli/services/sync_service.py` (2272 LOC) — already well past both the soft (1000) and hard (1500) ceiling CONTRIBUTING.md sets for `services/*.py`, and this PR adds another net +28 lines to it. The fix itself is small and surgical so it isn't worth blocking on a refactor here, but the next PR that adds material functionality to this file should extract the untracked-config walker + adopt-by-id logic into a sibling module (following the existing `_sync_push_ops.py` / `_sync_writeback.py` extraction pattern) before adding more.

## Verification log

- `gh auth status` → authenticated as `padak`, scopes `repo`/`workflow` ✓
- `gh pr view 494 --json title,body,files,additions,deletions,baseRefName,headRefName,labels,state` → OPEN, 2 files (`src/keboola_agent_cli/services/sync_service.py`, `tests/test_sync_service.py`), +256/-32, title `fix(sync): ...` — conventional-commit prefix matches (bug fix) ✓
- `git rev-parse --abbrev-ref HEAD` → `claude/issue-482-4d01d4`, matches `<branch>` input — working tree correctly checked out on the PR branch ✓
- Read `CONTRIBUTING.md` in full (Coding Style, Code Quality Patterns, Security Principles, "Checklist: Adding a New CLI Command", "Plugin synchronization map", "Releasing a new version") ✓
- Read `CLAUDE.md` convention #17 (silent-drift surfaces) and `## All CLI Commands` — confirmed this PR adds/removes/renames zero CLI commands, so the command-surface checklist (context.py `AGENT_CONTEXT`, `CLAUDE.md`, `OPERATION_REGISTRY`, `commands-reference.md`, `server/routers/`) is not applicable ✓
- Read `plugins/kbagent/agents/keboola-expert.md` §1 (non-negotiable rules) and §3 (inline gotchas) — no existing entry for issue #482; see NB-2 ✓
- `git diff --stat main...HEAD` → confirmed only `src/keboola_agent_cli/services/sync_service.py` (+82/-27 counting context lines) and `tests/test_sync_service.py` (+206/-5... net +201/-5 test additions) changed; no command/CLI/docs files touched ✓
- Layer-violation greps (`typer`/`click`/`formatter.`/`console.print` in services; `httpx`/`requests` in commands) against the diff → all empty ✓ no 3-layer violation (change is service-layer only)
- Convention greps (magic numbers, raw `error_code="..."` strings, bare `except:`, `print()` in prod code, token/secret leakage, new bare `tuple[...]` returns) against the diff → all empty ✓
- Read `diff_engine.compute_changeset` (`src/keboola_agent_cli/sync/diff_engine.py:299-430`) to verify the adopt-by-id path correctly falls into the "no base hash → 2-way compare" branch, producing `modified` (content differs) or a silent skip (content identical), and correctly marks the remote key as `seen` so it isn't double-flagged as `deleted` ✓
- Read `push_update` (`src/keboola_agent_cli/services/_sync_push_ops.py:318-370`) and the push-loop's post-update manifest hash writeback (`sync_service.py:1560-1584`) to trace the manifest-reconciliation gap in NB-1 ✓
- `uv run pytest tests/test_sync_service.py -k Issue482 -v` → 3 passed (`test_push_on_dev_branch_after_pull_is_noop`, `test_untracked_file_with_known_remote_id_updates_instead_of_creating`, `test_untracked_copy_of_tracked_config_still_creates_fresh`) — the exact issue #482 repro is asserted as `no_changes` / zero `create_config` calls ✓
- `make check` (lint + format-check + typecheck + skill-check + version-check + command-sync-check + changelog-check + check-error-codes + test, chained with hard-stop on first failure) → completed through the final test run: **4317 passed, 8 skipped, 131 deselected** ✓ — since `make` stops at the first failing prerequisite, this also confirms lint/format/typecheck/skill-freshness/version-sync/command-sync/changelog/error-code gates all passed
- `wc -l src/keboola_agent_cli/services/sync_service.py` → 2272 LOC (pre-existing over-hard-ceiling condition, not newly introduced by this PR) — see NIT-1
- Could not reproduce the fix against a live Keboola project (no token access, per policy that AI agents never handle Keboola API tokens); relied on code-level trace of the full diff/push/manifest interaction plus the new mocked-client regression test suite, which directly encodes the reported production repro — see NB-3 for the E2E gap this leaves

(none)

PR #501, opened 2026-07-20
feat(release): generate GitHub release notes from changelog.py

padak:
## Why

v0.66.1 shipped as a GitHub Release with an **empty body**: the tag was pushed alone, the `github-release` job created the release bare (softprops/action-gh-release is given no `body`), and nobody had pre-created the release by hand as previous releases relied on. With AI agents increasingly driving releases, "someone writes the notes by hand first" is not a process — the pipeline must produce the notes itself, and they must match what `kbagent changelog` shows.

## What

- **`scripts/gen_release_notes.py`** — renders the release body verbatim from the `CHANGELOG` dict in `src/keboola_agent_cli/changelog.py`, one `### vX.Y.Z` section per version. Reuses the CLI renderer's `_PREFIX_RE`, so the release page bolds exactly the prefixes `kbagent changelog` colours. With `--previous` (the latest stable release) it includes a section for **every version merged but never tagged** — the next tag after a gap (currently 0.67.0–0.71.0 are unreleased) announces all of them as a catch-up release.
- **Forward changelog gate** — a stable tag whose version has no `CHANGELOG` entry now fails the `gate` job before anything publishes. `generate_changelog.py --check` only audits already-published releases (backward direction); this closes the forward direction. Pre-releases pass `--allow-missing`, mirroring the existing changelog-check skip.
- **`release-kbagent.yml`** — the `gate` job (which already has the uv env) renders `release-notes.md` and ships it to `github-release` as an artifact. After softprops creates the release, a new step fills the body from the artifact **only when the body is empty** — a hand-authored release body is never overwritten.

## Consistency guarantees after this PR

| Direction | Enforced by |
|---|---|
| Every published release has a changelog entry | `generate_changelog.py --check` (existing) |
| Every tagged stable version has a changelog entry | `gen_release_notes.py` hard error in `gate` (new) |
| Release body == changelog content | body is generated verbatim from `CHANGELOG` (new) |
| Hand-written notes survive | fill step skips non-empty bodies (new) |

## Testing

- 17 new unit tests (`tests/test_gen_release_notes.py`): version-range collection (single, catch-up, misordered/unknown `--previous`), rendering (verbatim bullets, prefix bolding incl. `(#479)` / `(web UI)` decorations, footer), CLI behavior (missing stable version exits 1, `--allow-missing` writes nothing and exits 0).
- Full unit suite: 4361 passed.
- Live dry-run against the real changelog: `gen_release_notes.py --version 0.71.0 --previous 0.66.1` renders the expected 6-section catch-up body (0.67.0–0.71.0).

---

<a href="https://app.devin.ai/review/keboola/cli/pull/501" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.github/workflows/release-kbagent.yml
+++ b/.github/workflows/release-kbagent.yml
@@ -128,18 +128,23 @@ jobs:
       # travels to github-release as an artifact. This is also the FORWARD gate:
       # a stable tag whose version has no CHANGELOG entry fails the release now,
       # before anything publishes (generate_changelog.py --check above only
-      # audits already-published releases). `--previous` is the latest stable
-      # release, so versions that were merged but never tagged ship in the next
-      # tag as one catch-up section per version.
+      # audits already-published releases). The full released-version list goes
+      # to the script, which walks the changelog from the tagged version toward
+      # older entries until it hits one that is already released — versions that
+      # were merged but never tagged ship in the next tag as one catch-up
+      # section per version, with no reliance on `gh release list` ordering (an
+      # old-line hotfix tagged after a newer release still gets exactly its own
+      # section).
       - name: Generate release notes from changelog.py
         env:
           GH_TOKEN: ${{ github.token }}
           VERSION: ${{ needs.version.outputs.VERSION }}
           IS_PRERELEASE: ${{ needs.version.outputs.IS_PRERELEASE }}
         run: |
-          prev="$(gh release list --limit 1 --exclude-drafts --exclude-pre-releases --json tagName --jq '.[0].tagName' || true)"
           args=(--version "$VERSION" --output release-notes.md)
-          if [ -n "$prev" ]; then args+=(--previous "${prev#v}"); fi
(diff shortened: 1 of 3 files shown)


padak:
Both Devin findings addressed in 0128ae1:

1. **Null body defeats the emptiness guard** — fixed as suggested: the fill step now reads the body via `--jq '.body // ""'`, so a JSON-null body (which `--jq .body` would print as the literal string `null`) no longer masquerades as hand-written notes.

2. **`gh release list` ordering assumption** — fixed structurally rather than documented: the workflow now passes the full stable-release list (`--released`, repeatable) and `collect_versions` walks the changelog from the tagged version toward older entries, stopping at the first already-released one. No reliance on list ordering; an old-line hotfix tagged after a newer release yields exactly its own section, and an empty or changelog-disjoint released set conservatively yields a single section (never the entire history). Covered by new unit tests (adjacent release, catch-up walk, hotfix-after-newer, pre-created same-version release, no-boundary fallback).

Issue #503, opened 2026-07-20
linux native binaries require glibc >= 2.38 — fail on Ubuntu 22.04 / Debian 12 (freeze runs on ubuntu-latest)


padak:
## Summary

The frozen Linux binaries shipped since (at least) v0.71.0 require **glibc >= 2.38** and fail to start on older-but-current distros (Ubuntu 22.04 LTS, Debian 12):

```
[PYI-2687:ERROR] Failed to load Python shared library '/tmp/_MEI.../libpython3.12.so.1.0':
dlopen: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found
```

Surfaced by the v0.71.0 release run's `test-install` job ([run 29752342509](https://github.com/keboola/cli/actions/runs/29752342509)), **Homebrew (Linux)** leg. This is the first release where that leg got far enough to execute the binary — on v0.66.1 it died earlier on the untrusted-tap error, which #483 fixed. So this is a **pre-existing property of the freeze setup made visible by #483**, not a regression introduced in 0.71.0.

## Root cause

- The `freeze (ubuntu-latest, linux, amd64)` job runs on **ubuntu-latest = 24.04 (glibc 2.39)**. PyInstaller output inherits the build host's glibc floor; the resulting binary needs `GLIBC_2.38` symbols from the *system* libm/libc.
- The `homebrew/brew` docker image is Ubuntu **22.04-based (glibc 2.35)** → dlopen fails.
- Note: brew even pours its own `glibc 2.39` bottle as a formula dependency, but that does not help — the PyInstaller bootloader resolves the system loader/`/lib/x86_64-linux-gnu`, not linuxbrew's Cellar glibc.

Current `test-install` matrix results (v0.71.0): apt on `ubuntu:latest` (24.04) passes, `fedora:latest` passes, Homebrew (Linux, 22.04 base) fails.

## Impact

Any user on a distro with glibc < 2.38 (Ubuntu 22.04 LTS until 2027, Debian 12, RHEL 9 [glibc 2.34], Amazon Linux 2023 [2.34]) cannot run the native binary from **any** channel — zip, deb, rpm, or Homebrew — because all Linux artifacts come from the same freeze job. The `pip`/`uv` wheel path is unaffected.

## Suggested fix

Freeze the Linux builds on the **oldest supported base** instead of `ubuntu-latest`:

[...]

PR #504, opened 2026-07-20
fix(release): freeze linux binaries on ubuntu-22.04 to restore glibc 2.35 compatibility

padak:
## Why

A PyInstaller binary requires the build machine's glibc (or newer) at runtime, so the freeze runner sets the compatibility floor for every end user. The linux amd64 freeze leg runs on `ubuntu-latest`, which is 24.04 (glibc 2.39) — the v0.71.0 binary fails on any glibc < 2.38 system:

```
[PYI-2687:ERROR] Failed to load Python shared library '/tmp/_MEI188JGu/libpython3.12.so.1.0':
dlopen: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found
```

Caught by `test-install`'s Homebrew (Linux) leg in the v0.71.0 release run (29752342509), which runs the binary inside the `homebrew/brew` image (Ubuntu 22.04-era glibc). The apt/rpm legs passed only because `ubuntu:latest` and `fedora` carry new-enough glibc. This is an end-user regression, not a CI quirk: Ubuntu 22.04 / Debian 12 systems cannot run the frozen binary at all. It was previously masked because the v0.66.1 Homebrew leg failed earlier on the missing `brew trust` step (fixed in #483).

## What

Pin both linux freeze legs to the oldest GA runners — `ubuntu-22.04` (amd64) and `ubuntu-22.04-arm` (arm64), glibc 2.35 — and document in the matrix why `-latest` must never be used for freeze legs. Both runner labels are GA per actions/runner-images (no deprecation scheduled).

## Verification

- The released v0.71.0 linux binary demonstrably requires `GLIBC_2.38` (release-run failure above).
- Workflow YAML validates; no other pipeline change.
- Full proof comes from the next release run's `test-install` Homebrew leg, which now exercises exactly the old-glibc scenario. A `workflow_dispatch` dry run of `release-kbagent` on this branch builds the binary on the pinned runners without publishing anything.

## Relation to #503

Refs #503 — this is the "mechanical change" the issue describes (option A: pin to `ubuntu-22.04`, glibc 2.35 floor). It restores Ubuntu 22.04 LTS and Debian 12. It does **not** cover RHEL 9 / Amazon Linux 2023 (glibc 2.34) — if that class of distro should run the native binary, the freeze has to move into an older container (debian 11 / manylinux-style base), which is the floor decision #503 leaves to the maintainers. Keeping #503 open for that call is reasonable; this PR removes the immediate 2.38 regression either way.

### DIFF
--- a/.github/workflows/release-kbagent.yml
+++ b/.github/workflows/release-kbagent.yml
@@ -151,9 +151,19 @@ jobs:
       matrix:
         # `platform` is the OS name baked into artifact/zip filenames (darwin/linux/
         # windows) — Homebrew/Chocolatey/WinGet URLs depend on these exact values.
+        #
+        # Linux legs are PINNED to the OLDEST GA runner, NOT ubuntu-latest: a
+        # PyInstaller binary requires the build machine's glibc (or newer) at
+        # runtime, so the build runner sets the compatibility floor for every
+        # end user. ubuntu-latest (24.04, glibc 2.39) produced a v0.71.0 binary
+        # that failed with "GLIBC_2.38 not found" on Ubuntu 22.04-era systems
+        # (caught by test-install's Homebrew leg, run 29752342509). Building on
+        # ubuntu-22.04 (glibc 2.35) keeps Ubuntu 22.04 / Debian 12 installable.
+        # Bump these only when the OS baseline is consciously dropped — never
+        # back to -latest.
         include:
-          - { os: ubuntu-latest, platform: linux, arch: amd64 }
-          - { os: ubuntu-24.04-arm, platform: linux, arch: arm64 }
+          - { os: ubuntu-22.04, platform: linux, arch: amd64 }
+          - { os: ubuntu-22.04-arm, platform: linux, arch: arm64 }
           # Single macOS environment (Apple Silicon). PyInstaller can't cross-compile,
           # so there's no Intel leg; Intel Macs use `uv tool install`. Re-add a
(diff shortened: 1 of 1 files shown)


padak:
Post-merge verification of the glibc floor (local, full freeze-job replica):

**Build** — `ubuntu:22.04` container (arm64, native), mirroring the freeze job: deadsnakes Python 3.12 (built on 22.04, same class as actions/setup-python's toolcache build), Node 20 for the SPA, `uv build --wheel`, PyInstaller 6.11.1 `--onefile --collect-all keboola_agent_cli`. Build succeeds; the bootloader ELF requires only `GLIBC_2.17` (PyInstaller ships manylinux bootloaders).

**Run** — the produced binary executed in a **clean** `ubuntu:22.04.5` container (glibc 2.35, no Python, `env -i`):

```
$ env -i /out/kbagent --version
kbagent v0.71.0
```

This is the exact scenario where the released v0.71.0 amd64 binary died with `GLIBC_2.38 not found` (the homebrew/brew image is 22.04-based). A 22.04-built binary runs there fine, confirming the pinned runners restore the 2.35 floor. The verification build is arm64 (native on the test machine); the glibc-floor mechanism is architecture-independent and both linux legs are pinned by this PR. Full end-to-end proof lands with the next release run's test-install Homebrew leg (or a `workflow_dispatch` dry run from main).

PR #535, opened 2026-07-24
DMD-1825 feat(auth): browser login via PKCE + device authorization (programmatic auth)

zajca:
# DMD-1825 feat(auth): browser login via PKCE + device authorization (programmatic auth)

## What this changes

`kbagent` can now authenticate against a Keboola stack through a **browser
login** instead of a hand-copied static Storage token. `kbagent auth login`
runs a PKCE authorization-code flow by default, falling back to the RFC 8628
device flow when a loopback browser is not reachable. The result is a
USER-scoped **programmatic session** (a `kbc_at_*` access token plus a
`kbc_rt_*` refresh token) that the CLI refreshes on its own.

Why: obtaining a static token today means opening the Keboola UI, minting a
token by hand, and pasting it into a terminal, where it lands in shell history
and stays valid until someone remembers to revoke it. A session is short-lived,
server-revocable, tied to the actual human, and never typed anywhere.

The static-token path is untouched. Both credential kinds coexist per project
inside one `config.json`, and every existing command keeps working exactly as
before for static projects.

## Observable differences

| Before | After |
|---|---|
| Registering a project required a token pasted on the command line (`project add --token`) | `kbagent auth login` opens a browser; no token is ever typed or shown |
| No notion of a session; `config.json` held a literal token per project | Session projects carry a `kbc-session://{project_id}` sentinel in `config.json`; the live credential lives in `auth.json` (`0600`, sibling of `config.json`) |
| Nothing to expire | Access tokens refresh transparently (proactively before expiry, and reactively once on a real 401); a dead session is purged and reported as `SESSION_EXPIRED` |
| n/a | `kbagent auth status` reports live / expired / missing distinctly, refreshing an expired access token rather than reporting a false negative |
| n/a | `kbagent auth logout` revokes the session server-side, reporting an unconfirmed revoke distinctly from a clean one, and always clears local state |
| n/a | `kbagent project list` shows an auth-mode column, so a session project is visibly different from a static one |

`config.json`'s schema and `CURRENT_CONFIG_VERSION` are **unchanged** -- a
sentinel is a value in the existing `token` field, not a new shape. Existing
configs load untouched, and downgrading a session project to static works via
`project edit --token` (with a warning, see D6 below).

## Commands added

```
kbagent auth login [--stack URL|alias] [--device-code] [--register-projects]
kbagent auth status [--stack URL|alias]
kbagent auth logout [--stack URL|alias] [--remove-projects] [--yes]
[...]

PR #541, opened 2026-07-29
chore(release): bump to 0.76.3, fix all open npm Dependabot alerts

Matovidlo:
## Summary
- Resolves all 22 open Dependabot alerts (npm ecosystem only; no Python alerts were open).
- `web/backend`: bump `@fastify/static` 9.x → 10.1.2 (fixes path-traversal / auth-bypass advisories GHSA-83w8-p2f5-377r, GHSA-8pvw-jcv7-9cmj); `npm audit fix` picked up patched `find-my-way` and `brace-expansion` transitively.
- `web/frontend`: `npm audit fix` picked up patched `postcss` (source-map path traversal); added an `overrides` entry pinning `dompurify` to `^3.4.12` to close the XSS/sanitization-bypass chain of advisories on the copy bundled transitively via `monaco-editor` and `mermaid`.
- Bumped `pyproject.toml` to `0.76.3` and added the changelog entry; ran `make version-sync` to propagate to `plugin.json` / `marketplace.json` / `uv.lock`.
- No application behavior change — dependency/security bumps only.

## Test plan
- [x] `cd web/backend && npm install && npm audit` → 0 vulnerabilities
- [x] `cd web/frontend && npm install && npm audit` → 0 vulnerabilities
- [x] `cd web/backend && npm run build` (tsc) → passes
- [x] `cd web/frontend && npm run build` (vite) → passes
- [x] `uv run python scripts/sync_version.py --check` → all files at 0.76.3
- [x] `uv run python scripts/generate_changelog.py --check` → all releases have changelog entries
- [x] `uv run ruff check` / `ruff format --check` on changed Python files → clean
- [ ] Neither `web/backend` nor `web/frontend` currently has test files (`vitest run` reports "No test files found") — pre-existing gap, not introduced by this PR

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
   "plugins": [
     {
       "name": "kbagent",
-      "version": "0.76.2",
+      "version": "0.76.3",
       "source": "./plugins/kbagent",
       "description": "AI-friendly interface to Keboola Connection projects — explore configs, jobs, lineage, call MCP tools, manage dev branches, and debug SQL in workspaces",
       "category": "development"
--- a/plugins/kbagent/.claude-plugin/plugin.json
+++ b/plugins/kbagent/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kbagent",
-  "version": "0.76.2",
+  "version": "0.76.3",
   "description": "AI-friendly interface to Keboola Connection projects — explore configs, jobs, lineage, call MCP tools, manage dev branches, and debug SQL in workspaces",
   "author": {
     "name": "Keboola",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "keboola-cli"
-version = "0.76.2"
+version = "0.76.3"
 description = "AI-friendly CLI for managing Keboola projects"
 readme = "README.md"
 requires-python = ">=3.12"
--- a/src/keboola_agent_cli/changelog.py
+++ b/src/keboola_agent_cli/changelog.py
@@ -24,6 +24,13 @@ from .constants import CHANGELOG_HEADLINE_MAX_CHARS
 
 # Ordered newest-first.  Each value is a list of brief one-line descriptions.
 CHANGELOG: dict[str, list[str]] = {
+    "0.76.3": [
+        "Security: bump npm dependencies flagged by Dependabot in `web/backend` and "
+        "`web/frontend` -- `@fastify/static` (path traversal / auth bypass), "
+        "`brace-expansion` and `find-my-way` (DoS), `postcss` (source-map path traversal), "
+        "and `dompurify` (XSS/sanitization bypass, pinned via an npm override on the "
+        "`monaco-editor`/`mermaid` transitive copy). No behavior change.",
+    ],
     "0.76.2": [
         "Fix (#528, #530): self-update no longer risks leaving the running Windows uv tool "
         "environment partially upgraded. kbagent now completes all network checks and command "
--- a/uv.lock
+++ b/uv.lock
@@ -590,7 +590,7 @@ wheels = [
 
 [[package]]
 name = "keboola-cli"
-version = "0.76.2"
+version = "0.76.3"
 source = { editable = "." }
 dependencies = [
     { name = "croniter" },
--- a/web/backend/package-lock.json
+++ b/web/backend/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@fastify/cors": "^11.0.0",
-        "@fastify/static": "^9.0.0",
+        "@fastify/static": "^10.1.2",
         "fastify": "^5.8.5",
         "pino-pretty": "^13.1.3",
         "undici": "^6.27.0"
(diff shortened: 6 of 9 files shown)


Matovidlo:
@claude review

padak (commented):
## Review — `chore(release): bump to 0.76.3, fix all open npm Dependabot alerts`

**Verdict: no blockers.** The dependency work is correct, and the `dompurify` override in particular is the right call for a non-obvious reason (below). Findings are about the PR description and one piece of undocumented reasoning, not the change itself.

### Verified against the lockfiles, not just `npm audit`

`npm audit` reports 0 vulnerabilities *after* `overrides` reshapes the tree, so it partly grades its own homework. I diffed installed versions `main` → PR and checked each against the advisories' `vulnerable_version_range`:

| Package | main | PR | Result |
|---|---|---|---|
| `@fastify/static` | 9.1.3 | 10.1.2 | ✅ clears GHSA-83w8-p2f5-377r + GHSA-8pvw-jcv7-9cmj |
| `find-my-way` | 9.6.0 | 9.7.0 | ✅ |
| `brace-expansion` | 5.0.7 | 5.0.8 | ✅ |
| `postcss` (backend **and** frontend) | 8.5.15 | 8.5.24 | ✅ both manifests |
| `dompurify` (top level) | 3.4.12 | 3.4.12 | already clean |
| `monaco-editor/node_modules/dompurify` | **3.2.7** | *(gone)* | ✅ nested copy eliminated |

Also checked and clean:

- **`@fastify/static` v9 → v10 is a major bump, but safe here.** The only breaking change in v10.0.0 is `setHeaders` receiving `FastifyReply` instead of `Response`. `web/backend/src/server.ts:47` registers only `root` / `prefix` / `wildcard` — no `setHeaders`, so nothing to migrate.
- **Node floor.** `brace-expansion` narrowed its engines to `20 || >=22`; backend declares `>=20` and CI pins Node 20 → consistent.
- **`_ui_dist` is not a stale artifact.** It's gitignored and populated by the hatch build hook at wheel-build time, so the frontend fix reaches shipped wheels without anything committed here.
- **Version bump is consistent.** main `0.76.2`, latest release `v0.76.2`, `0.76.3` propagated to `plugin.json` / `marketplace.json` / `uv.lock`, changelog entry present. All checks green.

---

### 🟡 1. The override is load-bearing for a reason the PR doesn't state — please capture it

I initially read the override as redundant: the PR also bumps `monaco-editor` 0.55.1 → 0.56.0, and 0.56.0 pins `dompurify` at exactly `3.4.8`, which looks patched. It isn't. Per the advisory data, `3.4.8` is still vulnerable to three of them:

- GHSA-c2j3-45gr-mqc4 — vulnerable `<= 3.4.11`, patched **3.4.12**
- GHSA-cmwh-pvxp-8882 — vulnerable `<= 3.4.10`, patched 3.4.11
- GHSA-vxr8-fq34-vvx9 — vulnerable `< 3.4.9`, patched 3.4.9

Because monaco pins **exactly** `3.4.8` (not a range), npm has no choice but to materialize a nested vulnerable copy — the override is the only way to force it up. So this is the correct fix, not a workaround.

That reasoning is exactly what a future maintainer needs and can't reconstruct from a 3-line `overrides` block. It's also a standing commitment: it globally overrides an upstream *exact* pin, and once monaco ships a release pinning `>= 3.4.12` nothing will flag it as removable. `package.json` can't carry comments, so a line in `CONTRIBUTING.md` or a `web/frontend/README` note would do:

> `overrides.dompurify` exists because `monaco-editor` pins `dompurify` at an exact version that trails the patched release (0.56.0 → 3.4.8, vulnerable to GHSA-c2j3-45gr-mqc4). Re-check on every `monaco-editor` bump; drop the override once its pin is `>= 3.4.12`.

### 🟡 2. `monaco-editor` 0.55.1 → 0.56.0 isn't mentioned

`npm audit fix` pulled a minor bump of the editor component itself. The description only mentions the `dompurify` override, so the changelog line "No behavior change" is doing quiet work for a UI component bump. Given `web/frontend` has no tests, this is the one change in the PR that could plausibly alter runtime behavior — worth naming explicitly rather than leaving it in the lockfile diff.

### 🟡 3. Alert count is 23, not 22 — and it matters more than usual here

There are currently **23** open alerts, including `postcss` on `web/backend/package-lock.json`, which the description attributes to the frontend only. The PR does fix the backend copy (8.5.15 → 8.5.24), so this is a description bug, not a coverage gap.

Flagging it because after merge the alert list auto-closes: if one alert is left standing, nobody can tell whether it's a miscount in the description or genuinely unfixed. Stating 23 makes the post-merge check unambiguous.

### 🟡 4. AI attribution footer contradicts `CONTRIBUTING.md:462`

> **No AI attribution footers** in PR descriptions

The body ends with `🤖 Generated with [Claude Code]`. Since this repo squash-merges and GitHub pre-fills the squash commit message from the PR body, it lands in `main`'s history unless overridden at merge time.

---

### 🟢 Note: CI doesn't cover `web/backend`

CI builds the wheel (which builds the frontend SPA) but never runs `npm run build` or `npm test` on the Node BFF, so the major `@fastify/static` bump rests entirely on the author's local `tsc` — that's why I checked `setHeaders` usage by hand rather than trusting the test plan. Not a blocker: `web/backend` is excluded from the distribution (`pyproject.toml:85`), so the blast radius is `make web-dev`, not shipped artifacts. Worth a follow-up issue if the BFF ever becomes shipped surface.

### ℹ️ Merge-train note (not this PR)

`0.76.3` is correct against main (`0.76.2`) and stays correct if this lands first. Heads-up though: **#535 and #538 both bump to `0.77.0`** and will collide with each other — whichever merges second needs renumbering.

padak (approved):
Approving — no blockers. Verified the fixes against the lockfiles rather than `npm audit` (which grades its own homework once `overrides` reshapes the tree): every flagged package lands on a version outside its advisory's vulnerable range, and the nested `monaco-editor/node_modules/dompurify@3.2.7` is gone. The `@fastify/static` v9→v10 major is safe here — its only breaking change (`setHeaders`) isn't used in `server.ts`.

Two asks before the squash-merge, both on the description rather than the code:

1. Drop the AI attribution footer (`CONTRIBUTING.md:462`) — the repo squash-merges and GitHub pre-fills the commit message from the PR body, so it lands in `main`'s history otherwise.
2. Mention the `monaco-editor` 0.55.1 → 0.56.0 bump that `npm audit fix` pulled in, and correct the alert count to 23. The count matters after merge: when the alert list auto-closes, a mismatch makes it ambiguous whether a leftover alert is a miscount or genuinely unfixed.

The `overrides.dompurify` entry is the right call and worth a one-line note somewhere durable — monaco pins `dompurify` at exactly 3.4.8, which is still vulnerable to GHSA-c2j3-45gr-mqc4 (patched 3.4.12), so the override is the only way up. Once monaco pins `>= 3.4.12` nothing will flag it as removable.

Unrelated heads-up: #535 and #538 both bump to 0.77.0 and will collide with each other. This PR's 0.76.3 is fine either way.

Issue #522, opened 2026-07-22
`kbagent serve --ui` crashes on startup on Windows with non-UTF-8 console codepage (UnicodeEncodeError in banner)


papousek-radan:
## Summary

On Windows with a non-UTF-8 console codepage (e.g. cp1250 on Czech Windows 11), `kbagent serve --ui` crashes immediately during startup. The startup banner writes box-drawing characters (`├─`, `└─`) via `sys.stdout.write`, which raises `UnicodeEncodeError` before `uvicorn.run` is reached — the server never binds.

## Environment

- kbagent v0.72.0 (banner code unchanged on current `main` / v0.75.0)
- Windows 11 Enterprise, Czech locale — Python stdout encoding resolves to `cp1250`
- Installed via `uv tool install`

## Steps to reproduce

1. On Windows with a non-UTF-8 console codepage (Czech/Polish/Hungarian locales default to cp1250 for Python's stdout when not in UTF-8 mode), run:
   ```
   kbagent serve --ui
   ```
2. The process crashes before binding the port.

## Actual behavior

```
UnicodeEncodeError: 'charmap' codec can't encode characters in position ...
  File "...\encodings\cp1250.py", line 19, in encode
```

raised from the `sys.stdout.write(...)` banner in `src/keboola_agent_cli/commands/serve.py` (lines ~235–273). Both the `--ui` and non-UI banner branches contain `├─` / `└─`, which cp1250 cannot encode. The same crash would affect any legacy single-byte console encoding.

## Expected behavior

The server starts; the banner degrades gracefully on terminals that can't render Unicode.

## Workaround

Set `PYTHONUTF8=1` before launching (verified working):

```
set PYTHONUTF8=1
kbagent serve --ui
```

## Suggested fix

Either of:

1. Reconfigure stdout before printing the banner so encoding errors can't be fatal, e.g.:
   ```python
   sys.stdout.reconfigure(errors="replace")  # Python 3.7+
   ```
   (or wrap in `io.TextIOWrapper(..., encoding="utf-8")`), or
2. Fall back to ASCII glyphs (`|-`, `` `- ``) when `sys.stdout.encoding` can't encode the banner — the same pattern `install.sh` already uses for its UTF-8 locale check (lines 62–66: Unicode glyphs only under a UTF-8 locale, ASCII otherwise, "so the banner never renders as mojibake on a legacy terminal").

[...]

PR #547, opened 2026-08-01
fix(output): force UTF-8 for machine output on non-UTF-8 consoles (#546)

### LINKED ISSUE #546: `--json` output crashes with UnicodeEncodeError on Windows when data contains non-ASCII characters
Running any `kbagent --json ...` command whose output contains non-ASCII characters (in our case an arrow present inside a flow name) crashes on Windows instead of printing JSON.

Reproduced with:
```
kbagent --json flow list
```

Traceback (trimmed):
```
File "keboola_agent_cli/commands/flow.py", line 200, in flow_list
    formatter.output(result)
File "keboola_agent_cli/output.py", line 57, in output
    sys.stdout.write(response.model_dump_json(indent=2) + "\n")
File ".../encodings/cp1250.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '→' in position 28156: character maps to <undefined>
```

**Environment:** Windows 11, PowerShell/Git Bash, default console codepage cp1250 (not UTF-8).

**Workaround:** forcing UTF-8 before invoking the CLI avoids the crash:
```
PYTHONIOENCODING=utf-8 PYTHONUTF8=1 kbagent --json flow list
```

**Why this matters:** `--json` mode exists specifically for machine consumption (piping to a file or another program), so it shouldn't depend on the active console codepage. Any project/flow/config name containing a non-ASCII character (arrows, accented characters, emoji, etc.) makes `--json` unusable on a default Windows terminal.

**Suggested fix:** in JSON/machine-output mode, `output.py` should write to stdout using UTF-8 explicitly (e.g. reconfigure `sys.stdout` encoding, or write UTF-8 bytes directly) instead of relying on the inherited console codepage.

padak:
## What

`kbagent --json <anything>` crashed with `UnicodeEncodeError` on Windows whenever the payload contained a non-ASCII character (the reporter hit an arrow inside a flow name via `kbagent --json flow list`).

Machine output is now written as UTF-8 regardless of the console codepage:

- New `write_machine_output()` in `output.py`, used by `OutputFormatter.output/error/success`, `kbagent http`, and the `agent --stream` NDJSON events.
- `force_utf8_stdout()` reconfigures `sys.stdout` to UTF-8 (covers every real `TextIOWrapper`).
- Belt-and-braces: on `UnicodeEncodeError` the payload is written as UTF-8 bytes to `sys.stdout.buffer`, which bypasses the text layer's codec. A stream with neither `reconfigure` nor `buffer` re-raises — for a machine consumer, a silently mangled payload is worse than a crash.

## Why

`sys.stdout` inherits the console codepage — cp1250 on Czech/Polish/Hungarian Windows 11. pydantic's `model_dump_json()` emits **raw UTF-8** rather than `\uXXXX` escapes, so the write hits the cp1250 codec and aborts the command. `--json` exists specifically for machine consumption (piping to a file or another program), so it must not depend on which terminal happens to be attached.

`agent.py`'s stream events serialize with `ensure_ascii=False` and shared the same exposure; `kbagent http` used the `json.dumps` default (`ensure_ascii=True`) so it never crashed, but it is routed through the same helper for consistency.

Unlike the `kbagent serve` startup banner (#522, fixed in #526), transliterating to ASCII is **not** an option here: the banner is decoration, this is data.

## How it was tested

- New `tests/test_json_output_encoding.py` (24 tests): a cp1250 stdout double that raises exactly like a real Windows console, plus a fidelity guard proving the double reproduces the crash (so the passing tests are not vacuous). Covers the reconfigure path, the byte-level fallback, the no-fallback re-raise, and `OutputFormatter.output/error/success` round-tripping a `→` through `json.loads`.
- Live check under `PYTHONIOENCODING=cp1250`: a raw `sys.stdout.write("extract → load")` raises `UnicodeEncodeError`, while `OutputFormatter(json_mode=True)` now emits valid UTF-8 JSON that parses back with `json.load`, both to a terminal and through a pipe.
- `make check`: lint + format + changelog-check + full suite (4707 passed, 8 skipped).

Modern UTF-8 terminals are byte-identical to before — no `\uXXXX` escaping is introduced.

Fixes #546

### DIFF
--- a/src/keboola_agent_cli/commands/agent.py
+++ b/src/keboola_agent_cli/commands/agent.py
@@ -23,7 +23,7 @@ from typing import Any
 import typer
 
 from ..errors import ConfigError, ErrorCode
-from ..output import OutputFormatter
+from ..output import OutputFormatter, write_machine_output
 from ..server.agents_store import AgentAction, Trigger
 from ..services.agent_service import AgentService
 from ._helpers import check_cli_permission, get_formatter, get_service
@@ -394,7 +394,7 @@ def _render_stream_event(formatter: OutputFormatter, evt: dict[str, Any]) -> Non
     multi-line summary with exit code, elapsed time, response preview.
     """
     if formatter.json_mode:
-        sys.stdout.write(json.dumps(evt, ensure_ascii=False) + "\n")
+        write_machine_output(json.dumps(evt, ensure_ascii=False) + "\n")
         sys.stdout.flush()
         return
     event = evt.get("event", "?")
--- a/src/keboola_agent_cli/commands/http_client.py
+++ b/src/keboola_agent_cli/commands/http_client.py
@@ -16,12 +16,12 @@ of :mod:`keboola_agent_cli.services.http_forwarder_service`.
 from __future__ import annotations
 
 import json
-import sys
 from typing import Any
 
 import typer
 
 from ..constants import HTTP_DEFAULT_TIMEOUT
+from ..output import write_machine_output
 from ..services.http_forwarder_service import (
     ForwardedResponse,
     ForwarderError,
@@ -41,7 +41,7 @@ http_app = typer.Typer(
 def _print_json(_console: Any, data: Any) -> None:
     """Human-mode renderer: pipe-safe pretty JSON.
 
-    Uses ``sys.stdout.write`` instead of ``console.print`` because Rich
+    Uses a raw stdout write instead of ``console.print`` because Rich
     can soft-wrap long strings or escape markup, which breaks downstream
     ``json.loads`` consumers. Real-world case: an AI agent piped
     ``kbagent http get /openapi.json`` into ``python3 -c "json.load(sys.stdin)"``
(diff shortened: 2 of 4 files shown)


padak:
Closing — superseded, and worth recording why rather than just deleting it.

**The bug this fixes is fixed.** #546 landed via #549 on 2 Aug and shipped in 0.78.0. `write_machine_output()` has been on `main` since then, so the substance of this PR is already in the product; this branch just never got closed behind it.

**One piece of it was deliberately not taken, though**, and I want that on the record so it does not get revived by accident. This PR also adds `force_utf8_stdout()`, which reconfigures `sys.stdout` to UTF-8 **unconditionally**.

#567 (just merged, in 0.80.1) solves the same remaining problem — human/Rich output crashing on Windows — but only when the stream is **not** a terminal. That distinction is the whole fix, and measuring it on a real Windows 11 box is what settled it:

| stdout | `encoding` | `"↔"` |
|---|---|---|
| real console (`isatty=True`) | **utf-8** | encodes fine |
| pipe / file (`isatty=False`) | **cp1252** | `UnicodeEncodeError` |

Since PEP 528, CPython writes to a real Windows console through the console API, so an interactive session already reports UTF-8 and was never affected. Forcing UTF-8 unconditionally therefore fixes nothing on the console side and actively breaks it: a console whose codepage is cp852 would start receiving UTF-8 bytes and render mojibake where it previously rendered correctly. Trading a crash nobody sees for garbled output everybody sees is a bad trade.

So `force_utf8_when_redirected()` on `main` carries an `isatty()` guard, with a test asserting a terminal is never reconfigured.

Anyone arriving here later: see #549 for the `--json` half and #567 for the human/Rich half. Nothing from this branch is lost.

Issue #545, opened 2026-07-30
Auto-update from v0.76.2 -> v0.77.0 leaves broken install (ModuleNotFoundError: keboola_agent_cli)


MichalProchazka:
While running `kbagent` (v0.76.2), the CLI's built-in auto-update check kicked in and attempted to upgrade to v0.77.0. The update failed with:

```
Updating kbagent v0.76.2 -> v0.77.0...
Auto-update failed; continuing with current version. Recover with: uv tool install --force --reinstall "keboola-cli @ https://github.com/keboola/cli/releases/download/v0.77.0/keboola_cli-0.77.0-py3-none-any.whl"
Error in sys.excepthook:
Traceback (most recent call last):
  File "typer/main.py", line 82, in except_hook
ImportError: cannot import name 'rich_utils' from 'typer'
```

Despite the message saying it would "continue with current version," the local install was left in a broken, inconsistent state -- every subsequent command (`kbagent flow --help`, `kbagent job --help`, etc.) failed immediately with:

```
ModuleNotFoundError: No module named 'keboola_agent_cli'
```

so the tool was completely unusable until manually repaired.

**Environment:** Windows 11, installed via `uv tool install`.

**Workaround:** running the exact recovery command printed in the failed-update message fixed it:
```
uv tool install --force --reinstall "keboola-cli @ https://github.com/keboola/cli/releases/download/v0.77.0/keboola_cli-0.77.0-py3-none-any.whl"
```
After that, `kbagent --version` correctly reports `v0.77.0` and `kbagent doctor` passes.

**Suggested fix:** if auto-update fails partway (e.g. dependency conflict, partial package removal), the CLI should roll back to the previously-working install rather than leaving a half-updated environment that can't even print `--help`.

padak:
Thank you for the report — and specifically for including the full traceback. This is
the same failure as #528, and your details added evidence we did not have.

**Root cause.** `uv tool install` recreates a tool environment by **removing it in place**
and then building a fresh venv at the same path
([`uv-tool/src/lib.rs`](https://github.com/astral-sh/uv/blob/main/crates/uv-tool/src/lib.rs):
`create_environment` → *"Remove any existing environment"* → `create_venv`). That is not
atomic and has no rollback.

On POSIX it is harmless: unlinking a file another process holds open leaves that
process's inode intact. On Windows uv's `kbagent.exe` trampoline loads the venv's
interpreter **in-process**, so those files are locked — the removal deletes everything it
[...]

Issue #528, opened 2026-07-23
Self-update (auto-update on startup and `kbagent update`) can corrupt the uv tool venv on Windows — non-atomic in-place upgrade of the running venv


papousek-radan:
## Summary

kbagent's self-update mechanism — both the automatic `Updating kbagent vX -> vY...` on startup and the explicit `kbagent update` — can leave the uv tool venv in a corrupted, inconsistent state on Windows. The update swaps packages in-place inside the venv the running `kbagent.exe` was launched from, so any mid-flight failure (network error, locked file) strands the venv between versions and every subsequent invocation crashes.

Observed twice in two days on the same machine.

## Environment

- Windows 11 Enterprise (10.0.26200)
- kbagent installed via `uv tool install "keboola-cli[server] @ <release wheel URL>"`, Python 3.14
- kbagent versions involved: 0.72.0 → 0.75.0 (incident 1), 0.75.0 → 0.76.1 (incident 2)

## Incident 1 — auto-update on startup (2026-07-22, 0.72.0 → 0.75.0)

On startup kbagent printed `Updating kbagent v0.72.0 -> 0.75.0...`, the update **failed but kbagent continued**. Afterwards the venv had mismatched `rich`/`typer` versions, and every subsequent invocation failed with:

```
ModuleNotFoundError: No module named 'rich._windows'
```

and

```
ImportError: cannot import name 'rich_utils' from 'typer'
```

## Incident 2 — explicit `kbagent update` (2026-07-23, 0.75.0 → 0.76.1)

`kbagent update` crashed mid-update:

```
version_service.self_update → _update_mcp → httpx create_ssl_context → FileNotFoundError
```

The `FileNotFoundError` was for certifi's CA bundle file — the old dist had already been partially removed from site-packages while the still-running process referenced it, so when the post-update MCP version check tried to create an SSL context, certifi's `cacert.pem` no longer existed on disk. The venv was left broken between versions.

## Root cause (as far as we can tell)

The self-update replaces site-packages / the entrypoint **in-place, inside the venv that the running `kbagent.exe` was launched from**:

- On Windows the running exe is locked, so the upgrade fails with os error 32 if e.g. `kbagent serve` is running (that part is expected and correctly reported).
- But even without an exe lock, the dependency swap is not atomic. If anything fails mid-flight — a network error in the MCP-update step, a locked file — the venv is stranded with a mix of old and new dists, and there is no rollback.
[...]

PR #548, opened 2026-08-01
fix(update): surface why a startup self-update failed (#545) [superseded by #543 — do not review standalone]

padak:
## Scope, up front

This is **diagnostics, not the fix**. It does not change how the self-update mutates the tool environment, so it does **not** fix the corruption reported in #545 and #528. It removes the reason we cannot diagnose that corruption. Deliberately not marked `Fixes`.

## What

The startup auto-update hook runs the installer with `capture_output=True` and discarded the transcript, so a failed upgrade printed only:

```
Auto-update failed; continuing with current version. Recover with: uv tool install --force ...
```

The explicit `kbagent update` path has always surfaced `result.stderr` (`_update_kbagent` → `message` / `output`). The startup path did not — which is why both Windows reports (#545, 0.76.2 → 0.77.0; #528, three separate upgrades) could say the venv was left broken but not what `uv` actually refused to do.

- `_perform_update()` now returns an `UpdateAttempt` frozen dataclass: `outcome` (the existing `UpdateOutcome`) plus `detail`, the last actionable line of the installer transcript — `stderr`, falling back to `stdout` so the tail is never empty by accident. `OSError` and "no installer on PATH" get their own explanation.
- The failure banner prints it:

```
Auto-update failed (error: Access is denied. (os error 5)); continuing with current version. Recover with: ...
```

- `summarize_failure_tail()` is lifted out of `VersionService._summarize_failure_tail` (now a thin delegate) so both update paths compress a transcript identically.

`TIMEOUT` and `SUCCESS` carry no detail, and a failure with nothing to report keeps the old wording — no empty parentheses.

## Why this matters for #545 / #528

Both reports describe the same failure mode: the update fails partway and leaves the uv tool venv stranded between versions (`ImportError: cannot import name 'rich_utils' from 'typer'`, then `ModuleNotFoundError: No module named 'keboola_agent_cli'`), recoverable only with the printed `uv tool install --force --reinstall` command. What we still do not know is **why `uv` exits non-zero in the first place** — and without the installer's own last line, every new report reproduces that blind spot.

The remaining architectural fix (a fresh-resolution or post-exit detached updater, as suggested in #528) changes what auto-update means on Windows — the current invocation would keep running the old version, with no re-exec and no "What's new" — so it is left as a maintainer decision rather than guessed at here.

## How it was tested

- New `TestPerformUpdateFailureDetail`: stderr tail extraction, stdout fallback, missing-installer and `OSError` explanations, and that a timeout carries no detail.
- New `TestFailureBannerText`: the detail reaches stderr alongside the recovery command, and the banner stays clean when there is no detail.
- New cross-path test in `tests/test_version_service.py`: the module-level helper and the `VersionService` delegate compress a transcript identically.
- Existing `_perform_update` assertions updated for the new return type.
- `make check`: lint + format + changelog-check + full suite (4691 passed, 8 skipped).

Refs #545, #528

### DIFF
--- a/src/keboola_agent_cli/auto_update.py
+++ b/src/keboola_agent_cli/auto_update.py
@@ -16,6 +16,7 @@ import shutil
 import subprocess
 import sys
 import time
+from dataclasses import dataclass
 from importlib.metadata import distribution
 from pathlib import Path
 
@@ -47,6 +48,7 @@ from .services.version_service import (
     prepare_kbagent_update_plan,
     prepare_mcp_update_plan,
     resolve_kbagent_wheel_url,
+    summarize_failure_tail,
 )
 
 logger = logging.getLogger(__name__)
@@ -64,6 +66,28 @@ class UpdateOutcome(enum.Enum):
     SUCCESS = "success"
     TIMEOUT = "timeout"
     FAILED = "failed"
+
+
+@dataclass(frozen=True)
+class UpdateAttempt:
+    """One self-update subprocess run: its outcome plus *why* it ended that way.
+
+    The startup hook runs the installer with ``capture_output=True`` and used to
+    throw the transcript away, so a failed auto-update printed a bare
+    "Auto-update failed" and every bug report arrived without the one line that
+    explains it -- see issues #528 and #545, where the venv was left broken on
+    Windows and neither report could say what uv actually refused to do. The
+    explicit ``kbagent update`` path has always surfaced ``result.stderr``;
+    carrying the tail here closes that asymmetry.
+
+    Attributes:
+        outcome: SUCCESS / TIMEOUT / FAILED.
+        detail: Last actionable line of the installer transcript (empty when
+            there is nothing to report, e.g. on SUCCESS or TIMEOUT).
+    """
+
+    outcome: UpdateOutcome
+    detail: str = ""
 
 
 # Process-level sentinel for the auto-update flow.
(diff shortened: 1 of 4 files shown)


padak:
**Superseded — please read before reviewing.**

This PR was opened by the nightly issue-triage run for #545. After opening it I found #543, which was already in flight for #528 and is the same root cause: the reported broken install in #545 (0.76.2 → 0.77.0, `ImportError: cannot import name 'rich_utils' from 'typer'` → `ModuleNotFoundError: No module named 'keboola_agent_cli'`) is a second report of the non-atomic in-place venv swap #543 fixes properly.

#543 rewrites `_perform_update`, `maybe_auto_update` and the `version_service` failure path, so this branch **conflicts with it** and should not be merged alongside.

Recommendation: **close this** and, if the idea is wanted, fold it into #543 as a small follow-up. The residual value after #543 is narrow but real — on POSIX the inline install stays, `_perform_update` still returns a bare `UpdateOutcome`, and the banner still says only "Auto-update failed" with no reason. What this branch adds there:

- `UpdateAttempt` dataclass carrying the last actionable line of the installer transcript (stderr, falling back to stdout);
- the failure banner printing it: `Auto-update failed (error: Access is denied. (os error 5)); continuing with current version. Recover with: ...`;
- `summarize_failure_tail()` shared by the startup hook and `kbagent update`, so both paths compress a transcript identically.

#545 should be closed as a duplicate of #528 once #543 ships.

padak:
Superseded by #543, which is merged (598eca4). The failure-detail surfacing this PR added shipped as part of that change; closing so it does not linger in the merge queue.

PR #561, opened 2026-08-07
feat(auth): add kbagent auth pat-create/pat-revoke for CI/CD credentials

Matovidlo:
## Summary
- Adds **Personal Access Token (PAT)** support: `kbagent auth pat-create --name NAME [--totp-code CODE] [--read-only] [--ttl-days N] [--project-id ID ...]` and `kbagent auth pat-revoke PAT_ID [--yes]`, built on top of the existing browser-login session from #535.
- `pat-create` spends an **already-live** `auth login` session's access token to do a TOTP step-up (`POST /v1/auth/sudo`) and then mint a scoped, long-lived `kbc_pat_...` token (`POST /v1/auth/pat`). It never opens a browser and never logs in itself -- it does not relax `auth login`'s "requires a human at a browser" boundary; a live TOTP code is still required (interactive prompt, or `--totp-code`, and REQUIRED under `--json`/non-TTY -- there is no unattended path to mint one).
- `--project-id` (repeatable) scopes the PAT to an explicit allow-list instead of every accessible project, for a one-project-per-CI-secret setup.
- The resulting PAT is a **drop-in for `KBC_TOKEN`** under `KBAGENT_PROJECT_FROM_ENV=1` (or `project add --token`): `make_client_factory` now recognizes the `kbc_pat_` prefix on a plain static token and routes it to `Authorization: Bearer` instead of `X-StorageApi-Token` -- confirmed against the Storage API's own live OpenAPI security schemes that these are genuinely distinct auth mechanisms, not interchangeable encodings of one, before writing any of this.
- `pat-revoke` needs no step-up and is idempotent.
- New error codes: `AUTH_SUDO_REQUIRED`, `AUTH_MFA_INVALID`.

**Follow-up in #563** (stacked on this branch): a `--webauthn` step-up alternative to `--totp-code`, opening a browser for a passkey ceremony instead of a typed code. Split out because it depends on a not-yet-confirmed server-hosted ceremony page/redirect contract that needs live-stack testing before it's ready for review; this PR (TOTP-only) has no such dependency and is ready now.

## Why
CI/CD pipelines need a long-lived, non-interactive credential, but the only two options today are a raw Storage API token (no scoping, no easy revoke story tied to a person) or the browser-login session from #535 (explicitly human-only, never usable in CI). The Connection Auth Service already has first-class PAT support (`/v1/auth/pat`, `/v1/auth/sudo`) purpose-built for exactly this "authenticate once interactively, mint a durable scoped credential, use it non-interactively forever after" pattern -- this PR wires kbagent up to it. The `kbagent-cicd-migration` skill (#402) already recommends this command for CI secrets.

## Change type
Feature -- new CLI commands + a small credential-dispatch extension. No breaking changes; version bump 0.80.0 -> 0.81.0.

## Impact analysis
- `services/base.py`: `make_client_factory` gains one new branch (PAT-prefix detection) before the existing session-sentinel branch; a plain static token or session-sentinel token takes the byte-identical path as before.
- `auth/auth_client.py` / `auth/models.py` / `auth/token_provider.py`: three new client methods (`sudo_totp`, `create_pat`, `revoke_pat`), three new wire models (`SudoResult`, `PatItem`, `PatCreateResult`), one new lightweight `StaticBearerAuth` (no refresh/retry -- a PAT doesn't rotate).
- `services/auth_service.py`, `commands/auth.py`, `permissions.py`: two new commands, classified `admin` risk (same class as `project add`'s pasted static token / `project remove`).
- Fully backwards-compatible; no changes to existing command behavior.

## Test plan
- New unit tests: `test_auth_client.py` (sudo/create/revoke against a mocked HTTP transport, including the 404 feature-flag mapping and the sudo-not-active 403), `test_auth_service.py` (sudo-before-create ordering, sudo-failure aborts before create, no-session error, ttl/project-id forwarding), `test_cli_auth.py` (argument wiring, `--json`-without-`--totp-code` fails fast, admin permission-class gating, revoke confirm/abort), `test_base_service.py` (PAT-prefix dispatch to `Authorization: Bearer`, plain static token unaffected).
- `make check` green end-to-end: ruff, ty (0 new diagnostics vs. main), `make skill-check`/`command-sync-check`/`changelog-check`/`check-error-codes`/`check-sentinel-guards`/`loc-check`, full test suite (5436 passed, 11 skipped, 0 failed).

## Deployment
Merge & automatic deploy. No migration.

## Rollback plan
Revert of this PR.


Matovidlo:
@claude review

Matovidlo:
Closing: abandoning the PAT-for-CI approach. WebAuthn step-up (#562) can't work until the platform hosts the ceremony page, and that's the step-up factor actually available on the accounts we tested with -- TOTP-only is left unvalidated against a real account/MFA setup, so the whole 'mint a PAT via sudo step-up' flow isn't usable end-to-end today. Keeping the raw Storage token as the only CI credential path for now.

PR #565, opened 2026-08-10
feat(auth): add kbagent auth login-password for unattended CI login

Matovidlo:
## Summary

Adds `kbagent auth login-password` -- an unattended, CI-safe login path using email + password (+ optional TOTP code, computed by kbagent itself from a stored TOTP seed) to obtain a full session token via the Auth Service's password grant (`POST /v1/auth/login`, resolving MFA via `POST /v1/auth/mfa`).

## Why

`kbagent auth login` (PKCE / device-code) deliberately requires a human at a browser -- correct for interactive use, but a hard blocker for non-SAML CI/CD service accounts that only have email+password (and optionally a TOTP MFA seed) as credentials. This command is the explicit, narrow exception: same session-token machinery (`auth.json`, sentinel-token registration, introspect, etc.) as `auth login`, but reached without any browser step.

- `KBC_LOGIN_EMAIL` / `KBC_LOGIN_PASSWORD` / `KBC_LOGIN_TOTP_SECRET` env vars mirror the existing `KBC_TOKEN`/`KBC_STORAGE_API_URL` headless-injection convention.
- TOTP is computed locally via a ~30-line stdlib-only RFC 6238 implementation (`auth/totp.py`) -- no new dependency, and the *seed* is what's stored/passed, never a pre-computed code.
- Non-TOTP MFA factors (e.g. WebAuthn) fail fast with a new `AUTH_MFA_INVALID` error code pointing back at `kbagent auth login`, since those genuinely require a browser ceremony.
- `_finalize_login()` was extracted out of `login()` so both flows share the exact same persist/revoke/introspect/register-projects tail; existing `login()` tests all pass unchanged, confirming the extraction is behavior-preserving.

## Review pass (code quality + OWASP/security, independent agents)

Security review found **no material issues**: password/TOTP seed are never logged, written to disk, or included in exception messages; session storage correctly reuses the existing 0600-permissioned `auth.json` path rather than adding a parallel one; no injection risk (all values go through `httpx` JSON bodies, never string-formatted into URLs/shell). One non-blocking suggestion (add `--password-stdin` as a future enhancement to close the residual `ps`/shell-history exposure window) -- deferred, not required for this PR since the env-var path is already the documented primary usage.

Quality review found **one real bug**, now fixed in this PR: `compute_totp_code(totp_secret)` ran inline as a call argument, outside the command's `try/except (ConfigError, KeboolaApiError)` block -- a malformed or blank `--totp-secret` crashed with a raw `binascii.Error` traceback instead of the structured JSON error every other input parser in this codebase produces. Fixed by validating in `totp.py` (raises `ValueError` on empty/malformed base32) and mapping that to `ConfigError` at the CLI layer. Added test coverage for: TOTP code rotation across the 30s period boundary, empty/whitespace-only secret, malformed base32, and the CLI-level `ConfigError` mapping.

## Change type

- [x] New feature (new command)
- [ ] Bug fix
- [ ] Breaking change

## Impact analysis

- New CLI command `auth login-password`; new error code `AUTH_MFA_INVALID`; new env vars `KBC_LOGIN_EMAIL`/`KBC_LOGIN_PASSWORD`/`KBC_LOGIN_TOTP_SECRET`.
- No changes to existing `auth login`/`auth status`/`auth logout`/`auth register-projects` behavior beyond the internal `_finalize_login()` extraction (behavior-preserving, verified by unchanged existing tests).
- `permissions.py` `OPERATION_REGISTRY` gains `"auth.login-password": "write"`.
- Version bump 0.80.x -> 0.81.0 with a changelog entry.

## Test plan

- [x] `tests/test_auth_totp.py` -- RFC 6238 test vector, default digit count, determinism/rotation across the period boundary, empty/whitespace/malformed-secret error paths, seed formatting tolerance
- [x] `tests/test_auth_client.py::TestLoginPassword` / `TestVerifyMfaTotp` -- request bodies, MFA-required branch, 404 mapping
- [x] `tests/test_auth_service.py::TestLoginPassword` -- no-MFA path, TOTP-resolution path (call order), missing-code error, non-TOTP-factor error
- [x] `tests/test_cli_auth.py::TestLoginPassword` -- arg/env-var wiring, computed TOTP forwarded, malformed-secret maps to `ConfigError` (not a traceback), `AUTH_MFA_INVALID` surfaces in `--json`, password never appears in output
- [x] Full `make check` (ruff, format, ty, skill-check, version-check, command-sync-check, changelog-check, check-error-codes, check-sentinel-guards, loc-check, full pytest suite): **5428 passed, 11 skipped, 0 failed**

## Deployment / Rollback plan

Standard: merge to `main`, ships in the next `kbagent` release. No migrations, no state changes to existing installs. Rollback is a plain revert -- the new command/env vars/error code are additive only.


Matovidlo:
@claude review

Matovidlo:
@claude review

tests/test_auth_totp.py
```diff
@@ -0,0 +1,45 @@
...
+
+from keboola_agent_cli.auth.totp import compute_totp_code
+
+
+class TestComputeTotpCode:
+    def test_rfc6238_test_vector(self) -> None:
+        """RFC 6238 appendix B: seed 'GEZDGNBVGY3TQOJQGEZDGNBVGY3TQOJQ' (the base32
+        encoding of the ASCII string '12345678901234567890') at T=59s -> '94287082'
+        for SHA1/8-digit. This module hardcodes 6 digits, so check the low-order
+        6 digits of the same well-known reference value instead."""
```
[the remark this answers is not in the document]
Matovidlo: Fixed -- corrected the docstring to explain that digits defaults to 6 but the RFC's published vector is 8-digit, so the test passes digits=8 to check against the exact reference value.

src/keboola_agent_cli/commands/auth.py
```diff
@@ -382,6 +385,76 @@ def _run_post_login_hook(
...
+        ...,
+        "--email",
+        envvar=ENV_KBC_LOGIN_EMAIL,
+        help="Account email. Also settable via KBC_LOGIN_EMAIL.",
+    ),
+    password: str = typer.Option(
+        ...,
+        "--password",
+        envvar=ENV_KBC_LOGIN_PASSWORD,
+        help="Account password. Prefer KBC_LOGIN_PASSWORD (a CI secret in the step's "
```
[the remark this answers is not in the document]
Matovidlo: Fixed -- --password is now optional and a new --password-stdin flag was added (mirrors dev-portal identity add's existing pattern: hidden getpass prompt on a TTY, read-to-EOF on a pipe). login-password now fails with a ConfigError if neither --password, --password-stdin, nor KBC_LOGIN_PASSWORD supplied a value.

src/keboola_agent_cli/auth/auth_client.py
```diff
@@ -258,6 +261,33 @@ def __exit__(self, *args: Any) -> None:
     # PKCE
```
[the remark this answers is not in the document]
Matovidlo: Fixed -- split the section header: login_password/verify_mfa_totp now have their own "Password grant + MFA" header, and the PKCE header now only covers the actual PKCE methods below them.

src/keboola_agent_cli/services/auth_service.py
```diff
@@ -270,92 +270,165 @@ def login(
...
+                        f"This account requires MFA type {result.mfa_type!r}, which "
+                        "password-grant login cannot resolve without a browser -- use "
+                        "`kbagent auth login` instead.",
+                        error_code=ErrorCode.AUTH_MFA_INVALID,
+                        retryable=False,
+                    )
+                if not totp_code:
+                    raise ConfigError(
+                        "This account requires a TOTP code to sign in -- pass totp_code."
                     )
```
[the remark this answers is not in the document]
Matovidlo: Fixed -- the message now says "pass --totp-secret" instead of "pass totp_code", matching the actual public CLI flag.

zajca (commented):
## Review: NEEDS-WORK

Reviewed against the real Connection implementation on `master` (`d6bb758342`), `connection/docs/rfc/programmatic-auth`, and this repo's binding conventions. The wire contract is essentially right; the problems are in error handling, TOTP timing, four unupdated agent-facing doc surfaces, and one factual claim in `docs/auth.md`.

---

### A. Confirmed correct against Connection master

Worth recording, because most of the contract checks out — in particular the load-bearing assumption.

| CLI assumption | Verdict | Evidence |
|---|---|---|
| `POST /v1/auth/login`, body `{grantType,email,password}` | correct | `AuthLoginAction.php:67,117`; `LoginRequest.php:30,72-95` |
| `allowExtraFields:false` — no stray keys sent | correct | `LoginRequest.php:91` vs `auth_client.py:275` |
| **MFA challenge arrives as HTTP 200 with `mfaRequired` in the body** | correct — this is the load-bearing one | `AuthLoginAction.php:171` returns `HTTP_OK` for both `LoginResponse` and `MfaChallengeResponse` |
| `mfaRequired/mfaType/mfaToken/expiresIn/allowedMethods` | correct | `MfaChallengeResponse.php:16-22,84-97` |
| `POST /v1/auth/mfa` with `{mfaToken,type:"totp",code}` | correct | `MfaAction.php:49`; `MfaRequest.php:29-33`; `MfaType.php:16` |
| 6-digit code required | correct | server enforces `/^\d{6}$/`; `totp.py:37` does `zfill(6)` |
| Reusing `CliTokenResponse` for the response | correct — **byte-identical** shapes | `LoginResponse.php:59-74` vs `CliLogin/CliTokenResponse.php:59-75` |
| 404 ⇒ feature flag off, same flag as PKCE/device | correct | both controllers gate on `AuthFeatures::STACK_PROGRAMMATIC_AUTH` |
| `mfa_type != "totp"` ⇒ cannot proceed | correct, **not** the bug it looks like | `ProgrammaticSessionService.php:113-116` — an admin has exactly one primary factor, so `allowedMethods` is `[totp,recovery_code]` **or** `[webauthn,recovery_code]`, never both |

On the RFC: the password grant is **not** in conflict with it. `programmatic-auth.md:56` lists it in scope ("Password + TOTP/recovery-code MFA flow (`grantType: password`)") and use case 5 is explicitly *"E2E tests — existing test suites need a non-browser path to obtain user-scoped tokens"* (`programmatic-auth.md:40`). The endpoint is a sanctioned programmatic path, not UI-only.

Also fine: `loc-check` passes with room (`commands/auth.py` 458 code lines vs 800/1200; `auth_service.py` 564 vs 1000/1500; `auth_client.py` 410 vs 1000/1500), `check_error_codes.py` and `check_command_sync.py` pass, the 0.80.0 → 0.81.0 bump is right (v0.80.0 released 2026-08-03), and neither `SESSION_UNSUPPORTED_FEATURES` nor the sentinel guards need an entry.

---

### B. Blockers (repo conventions)

**B1 — No E2E test, and unlike `auth login` there is no exemption available here.**

CLAUDE.md convention #16: *"Every new CLI command MUST have a corresponding E2E test."* `tests/test_e2e_auth.py` exists and was not touched. The 0.80.0 flows were legitimately exempt (`test_e2e_auth.py:31,479-506` — real browser / documented manual verification). `login-password` is the **one** auth command that is fully non-interactive, so it is also the one that can and must be covered end to end. Suggest new env gates `E2E_LOGIN_EMAIL` / `E2E_LOGIN_PASSWORD` / `E2E_LOGIN_TOTP_SECRET`, skipping when absent, mirroring the existing `E2E_API_TOKEN` / `E2E_URL` pattern in that file. Cover: no-MFA login, TOTP login, wrong password, `--register-projects`.

**B2 — Four mandated doc surfaces not updated; three of them now state something false.**

`scripts/check_command_sync.py:31-38` explicitly does not gate these, which is exactly why they were missed:

- **`plugins/kbagent/agents/keboola-expert.md`** (CONTRIBUTING.md calls it *"highest silent-drift risk"*) — line 155: *"needs a human at the keyboard, no headless path"*; lines 326-334 document programmatic auth as browser-only; the Rule 6 VERSION GATE (line 76) still says 0.80.0+ with no 0.81.0 entry. **Net effect: the `keboola-expert` subagent will refuse a command that now exists.**
- **`references/gotchas.md`** — heading line 14 and bullet lines 16-19: *"there is no headless/unattended path."* Line 167: *"any headless/unattended runner should keep using a static Storage token."* No new gotcha carrying the mandatory `(since v0.81.0)` tag. The error-code → exit-code table (line 1764) has no `AUTH_MFA_INVALID` row.
- **`plugins/kbagent/.claude-plugin/CLAUDE.md`** — lines 92-95: *"for unattended contexts point them at a static Storage token."*
- **`references/auth-workflow.md`** (14.7 KB) — untouched.

**B3 — Undocumented REST-route skip.**

CONTRIBUTING.md line 350 requires a 1:1 route per command or *"Document any skip in the PR description with a one-line reason."* There is no `server/routers/auth.py` and the PR description does not mention the skip. The skip is almost certainly correct here — exposing a password grant over `serve` would let whoever holds `KBAGENT_SERVE_TOKEN` submit arbitrary credentials — but it needs saying, ideally in `docs/web-server.md` too.

---

### C. Code defects

**C1 — `except ValueError` swallows pydantic validation errors and blames `--totp-secret`.**

In `commands/auth.py` the `try` block wraps `service.login_password(...)` as well as the TOTP computation. `pydantic_core.ValidationError` **is** a `ValueError` subclass, so every `model_validate` failure inside `login_password` / `verify_mfa_totp` / `introspect` surfaces as `CONFIG_ERROR: --totp-secret: N validation errors for CliTokenResponse …`, exit code 5, for a problem that has nothing to do with the flag.

```python
if totp_secret:
    try:
        totp_code = compute_totp_code(totp_secret)
    except ValueError as exc:
        _handle_errors(formatter, ConfigError(f"--totp-secret: {exc}"))
else:
    totp_code = None
try:
    result = service.login_password(...)
except (ConfigError, KeboolaApiError) as exc:
    _handle_errors(formatter, exc)
```

**C2 — A wrong password reports `Invalid or expired token (token: ****)`.**

`login_password` adds no status-specific mapping, so failures fall through to `http_base.py:306-320`:

- 401 (`InvalidCredentialsException`) → `INVALID_TOKEN`, with a message about a token that does not exist in this flow (`mask_token("")` gets interpolated into it).
- 403 → `ACCESS_DENIED`, flattening two distinct actionable cases: `SsoRequiredException` (SAML-enforced account, carries a recovery URL — `PasswordGrantProcessor.php:103-108`) and `SuperAdminMfaRequiredException` (same file, 130-137).
- 404 → the inherited message *"**Browser login** is not enabled on this Keboola stack yet"* (`auth_client.py:_map_auth_error`), for a command whose entire point is that it is not browser login.

This is the single most likely failure mode of the command and it currently misdiagnoses it. Suggest explicit mapping: 401 → "invalid email or password" (`AUTH_FLOW_DENIED` already exists); 403 → surface the server message and name `kbagent auth login` for the SSO case; generalize the 404 wording to "programmatic auth is not enabled on this stack".

**C3 — The TOTP code is computed too early.**

It is computed in the CLI layer *before* the `/v1/auth/login` round trip and carried into `/v1/auth/mfa`. `_do_request` retries 429/5xx and timeouts (3 attempts, `read=30s`, backoff 1s/2s — `http_base.py:172-218`, `constants.py:53-55,58`), so the gap can reach ~60-90 s. Server tolerance is ~90 s (`TotpVerifier.php:12-13`), so this is narrow rather than theoretical: two read timeouts on the login call (30 + 1 + 30 = 61 s) already push a slice-N code out of the accepted window.

Fix: pass the *seed* (or a `Callable[[], str]`) into `AuthService.login_password` and compute the code immediately before `verify_mfa_totp`. That also resolves C1 naturally.

**C4 — Retrying `/v1/auth/mfa` replays an already-consumed TOTP slice.**

`TotpVerifier.verifyAndConsume` records `(adminId, timeSlice)` via `INSERT IGNORE` and rejects any later submission of the same slice (`AdminTotpUsedTimeSliceRepository.php:34-42`, `TotpVerifier.php:9-19,33-42`). Two consequences:

1. A 429/5xx retry inside `_do_request` resubmits the **same** code — guaranteed rejection, and it burns one of `MAX_CHALLENGE_ATTEMPTS` (`MfaChallengeRepository.php:66-72`).
2. **Two CI jobs calling `login-password` within the same 30 s slice: the second one fails.** A matrix build sharing one MFA-enabled service account will fail intermittently with an opaque "invalid MFA code". Nothing in the docs warns about this.

Fix: exclude `verify_mfa_totp` from the retry loop — the same deliberate exception `poll_device_token` and `refresh` already make (`auth_client.py:14-19`) — or retry with a freshly computed code. Document the concurrency constraint in `docs/auth.md`.

**C5 — The rate-limit signal is discarded.**

`/v1/auth/login` allows **5 failed attempts per email and 20 per IP per 15 minutes** (`RateLimitListener.php:37-40`), and the server emits `X-RateLimit-Limit` / `-Remaining` / `-Reset` on **every** response from these endpoints specifically *"so clients can self-throttle before tripping the lockout"*, reporting the most restrictive bucket because *"this is what a CLI or UI countdown cares about"* (same file, 114-122).

The CLI ignores all three headers and clamps `Retry-After` to 60 s (`constants.py:105`), so a 15-minute lockout becomes three pointless retries and a generic `API error 429`. On a 429 from login/MFA, read `X-RateLimit-Reset` and report when the account unlocks.

---

### D. Security and design

**D1 — The `docs/auth.md:152` claim that the session behaves "identically" to a browser-login session is false.**

For an **MFA-enabled** account the password flow goes through `createSessionAfterMfa`, which sets `sudoVerifiedAt: new DateTimeImmutable()` unconditionally (`ProgrammaticSessionService.php:136-150`). PKCE/device instead inherit `$context->sudoVerifiedAt()`, which is *"fresh only for MFA admins whose sudo window has not expired; NULL otherwise"* (same file, 162, 175, 208-209).

So `login-password` reliably produces a session with a **live 3-hour sudo window** (`SudoService::SUDO_TIMEOUT_SECONDS = 10800`), which `auth login` usually does not. Sudo gates exactly the account-takeover-shaped operations: PAT create/revoke, TOTP delete, WebAuthn delete/register, recovery-code regeneration, revoke-all-sessions. Connection's own code flags the risk: *"a login MFA code re-submitted to the sudo step-up on the freshly issued session … unlocks 1-year PAT creation"* (`TotpVerifier.php:14-17`).

The security note in the docs discusses blast radius but omits this escalation. Related: `IntrospectResponse` (`auth/models.py:154-161`) drops the `sudoVerified` field the API returns (`TokenIntrospectProcessor.php:51`) — surfacing it in `auth status` would make this visible.

**D2 — `KBC_LOGIN_PASSWORD` cuts against two established rules in this repo.**

Not necessarily wrong, but it should be a stated decision rather than an implicit reversal:

- CLAUDE.md convention #12 makes the manage token **default-deny from env** (`--allow-env-manage-token` to opt in) precisely because *"any subprocess (including the AI agent itself) inherits the manage token via env."* An account password is a longer-lived credential than a manage token — it cannot be revoked without a password change — yet it is env-readable by default here.
- `docs/programmatic-auth-login-plan.md:735-737` (§ risk 4) states the guardrail: credentials *"never logged, printed, **put on the command line**, or exported to subprocess environments."* `--password` puts one on the command line; `KBC_LOGIN_PASSWORD` exports one to the environment.

`--password-stdin` is the right primitive and fixes the command-line half. Consider deprecating `--password` in its favour — the pattern already exists in this repo (`dev-portal identity add --password-stdin`).

**D3 — The RFC nominates PATs for CI/CD, and kbagent has no PAT support at all.**

`programmatic-auth.md:326`, use-case table: *"User in Settings UI | `POST /v1/auth/pat` | … | Long-lived, broad-scope tokens for **CI/CD, scripts**."* PATs are shipped (*"Phase 2: Personal Access Tokens — shipped"*, line 282) and `PatCreateAction` / `PatExchangeAction` are live on master. The key operational difference (`programmatic-auth.md:429`): a password change cascade-revokes sessions, but *"user-created PATs survive — they are the credential the user explicitly persisted for automation, and revoking them on password change would silently break unrelated CI/CD pipelines."* A pipeline built on `login-password` therefore breaks on every password rotation.

Meanwhile kbagent has **zero** handling for `kbc_pat_*` or `/v1/auth/pat` (grep on this branch is empty). This is not a defect in the code as written — the password grant is legitimate for RFC use case 5 — but the PR should say why the password grant was chosen over the PAT path that already exists and that the RFC designates for CI/CD. Otherwise someone asks in three months.

**D4 — `docs/programmatic-auth-login-plan.md` (49.6 KB, the design authority CLAUDE.md points at) was not updated** and contains no mention of a password grant.

---

### E. Nits

- `totp.py:26` normalizes whitespace and case but does not pad base32. `b32decode` requires a length that is a multiple of 8, so a legitimate unpadded seed (e.g. 26 chars) is rejected as *"not valid base32"*. Add `cleaned += "=" * (-len(cleaned) % 8)`.
- `totp.py:26` strips `" "` but not `"-"`; some enrollment UIs hyphenate seeds.
- `MfaChallengeResult` does not model `webauthn`, which the server includes for U2F admins (`MfaChallengeResponse.php:20`). Harmless under `extra="allow"`; worth a docstring line.
- `allowedMethods` always includes `recovery_code`, including for WebAuthn-only accounts. A future `--recovery-code` would let those accounts use this command too — enhancement, not a gap.

---

### Suggested landing order

1. C1, C2 — smallest diffs, worst user-facing symptoms.
2. C3 + C4 — one refactor (seed pushed into the service, no retry on `verify_mfa_totp`).
3. B2 — the four doc surfaces; the false statements are the real damage.
4. B1 — E2E test.
5. D1 — fix the "identically" claim.
6. B3, D2, D3, D4 — PR description / design-doc write-ups.
7. C5, E — polish.

Matovidlo:
Thanks for the depth here — especially section A verifying the wire contract against Connection master line by line. Addressed everything below, following your suggested landing order, in 4 commits: a48a321 (C1/C2/C3/C4/C5/E), 2aa7e7d (B2/D1), 8bbe323 (B1), c8d2a54 (B3/D3/D4).

**C1 (`except ValueError` misattributing pydantic failures)** — Fixed. The seed now flows into `AuthService.login_password` as `totp_secret` and is computed just-in-time (see C3), so the command layer no longer wraps a bare `ValueError`; `--totp-secret` validation failures surface as `ConfigError` from the service itself. Covered by `test_auth_service.py::TestLoginPassword::test_malformed_totp_secret_raises_config_error_not_value_error`.

**C2 (wrong password → "Invalid or expired token")** — Fixed. `AuthClient._map_auth_error` now dispatches `/v1/auth/login` and `/v1/auth/mfa` to a dedicated `_raise_password_login_error`: 401 → `AUTH_FLOW_DENIED` with "Invalid email or password" (or "Invalid or expired TOTP code" on the MFA path), 403 → server message + points at `auth login`, 404 → "Programmatic auth" wording instead of "Browser login". Covered in `test_auth_client.py`.

**C3 + C4 (TOTP computed too early / retry replays a consumed slice)** — Fixed as one refactor. `login_password` now takes `totp_secret`, computes the code immediately before the MFA call, and `verify_mfa_totp` bypasses `_do_request` entirely (same deliberate exception `poll_device_token`/`refresh` already make) so a 429/5xx retry can never resubmit an already-consumed code. Covered by a new no-retry test in `test_auth_client.py` and the RFC-vector-seeded test in `test_auth_service.py`.

**C5 (rate-limit signal discarded)** — Fixed. A 429 from login/MFA now reads `X-RateLimit-Reset` and reports when the account unlocks, inside the same `_raise_password_login_error`.

**E (TOTP seed parsing nits)** — Fixed. `totp.py` now strips hyphens and pads to a multiple of 8 before `b32decode`. Added a docstring note on `MfaChallengeResult` re: unmodelled `webauthn`.

**B2 (four doc surfaces, three now false)** — Fixed all four: `keboola-expert.md` (Rule 6 gate, matrix row, programmatic-auth bullet), `gotchas.md` (new `(since v0.81.0)` section + `AUTH_MFA_INVALID` exit-code row), `plugins/kbagent/.claude-plugin/CLAUDE.md` (Path A fallback), `auth-workflow.md` (new "Unattended login" section, CI step in the loop, troubleshooting bullets, Boundaries correction).

**B1 (no E2E test)** — Fixed. Added `TestLoginPasswordCommand` to `test_e2e_auth.py`, gated on `E2E_LOGIN_EMAIL`/`E2E_LOGIN_PASSWORD`/`E2E_LOGIN_TOTP_SECRET` (skips cleanly when absent, same two-tier pattern as the existing session gate): login→status→logout (exercising the TOTP path when the secret is set), wrong-password (regression guard for C2, asserts the new message), and `--register-projects`. Bonus: this also means the `E2E_SESSION_REFRESH_TOKEN` provisioning recipe at the top of that file no longer strictly needs a real browser — noted in the docstring.

**D1 ("identically" claim, sudo escalation)** — Fixed the claim in `docs/auth.md`: narrowed to the parts that are actually identical, added a dedicated bullet on the 3-hour sudo window `createSessionAfterMfa` stamps unconditionally for MFA accounts, and documented the C4 concurrency constraint (two CI jobs sharing one MFA account in the same 30s window). I did **not** wire `sudoVerified` through `IntrospectResponse`/`auth status --json` — that's a real improvement but a wider change (touches every `AuthStatusResult` construction site + the CLI formatter); flagging it as a follow-up rather than scope-creeping this PR. Let me know if you'd rather it land here.

**B3 (undocumented REST-route skip)** — Documented in `docs/web-server.md`: the whole `auth` group has no router (pre-existing, not new to this PR), and `login-password` specifically is a deliberate skip — exposing a password grant over `serve` would let whoever holds `KBAGENT_SERVE_TOKEN` submit arbitrary account credentials.

**D3 (PATs vs. password grant)** and **D4 (design doc)** — Addressed together in a new §12 addendum to `docs/programmatic-auth-login-plan.md`: explains why the password grant (RFC use case 5, E2E tests) rather than PATs, and flags PAT support as a natural follow-up given a password rotation cascade-revokes this grant's sessions.

**D2 (`--password` on the CLI / `KBC_LOGIN_PASSWORD` from env)** — Deliberately **not changed**. I want your read on how far to take this before touching it: keep both and document it as a stated tradeoff (closest to what's there now), soft-deprecate `--password` in favour of `--password-stdin` the way `dev-portal identity add` does, or go all the way to default-deny-from-env like the manage token (`--allow-env-manage-token`) — which would break the simplest CI setup this command exists for. Wrote the tension up as its own bullet in the §12 addendum so it doesn't get lost either way.

Ready for another pass whenever you have time.

src/keboola_agent_cli/auth/auth_client.py
```diff
@@ -254,6 +260,48 @@ def __enter__(self) -> AuthClient:
...
+
+    def login_password(self, email: str, password: str) -> CliTokenResponse | MfaChallengeResult:
+        """Password-grant login (``POST /v1/auth/login``).
+
+        The unattended-capable login path -- unlike PKCE/device, this never
+        opens a browser. Returns a token pair directly, or an
+        `MfaChallengeResult` if the account has MFA configured (resolve it
+        with `verify_mfa_totp`).
+        """
+        response = self._do_request(
```
zajca: Reviewed by Opus.

`login_password()` uses `_do_request`, so a 429/5xx/timeout is retried up to `MAX_RETRIES` times (the PR's own `test_429_reports_rate_limit_reset` registers three 429 responses to reach the mapped error, confirming three POSTs). This is the one token-minting call in this client still inside the loop — `refresh`, `poll_device_token` and the new `verify_mfa_totp` all deliberately bypass it, and this file's module docstring states the rationale.

Two consequences:

1. **Rate limiting.** `_raise_password_login_error` documents that `/v1/auth/login` throttles per email and per IP over a 15-minute window (docs/programmatic-auth-login-plan.md: 5/20 per 15 min) and surfaces `X-RateLimit-Reset` "specifically so a client can self-throttle". The client instead fires two more login POSTs 1s and 2s later before printing that message — the server sends `X-RateLimit-Reset`, not `Retry-After`, so `_do_request` falls back to plain exponential backoff and cannot reach a 15-minute reset.

2. **Duplicate sessions.** Unlike `exchange_pkce_code` (single-use code, a retry just fails), the password grant is repeatable. A read timeout *after* the server created the session makes the retry mint a second session; only the last token pair reaches `_finalize_login`, so the first is a live session with no local record — `auth logout` revokes only the stored refresh token plus recorded orphans, so nothing can ever reach it for the refresh token's full lifetime.

Failure scenario: a CI job runs `auth login-password` against a stack that is briefly 5xx-ing or read-times-out after processing the login. The same `{grantType: password, email, password}` body is resubmitted up to 3 times: against a 429 it burns two extra requests of the account's 5-per-15-min bucket before printing "Try again after <reset>"; against a post-processing read timeout it creates a second server-side session whose tokens are discarded locally.

Giving `login_password` the same bypass as `verify_mfa_totp` (with `_post_refresh`-style transport-error mapping) resolves both.

src/keboola_agent_cli/auth/auth_client.py
```diff
@@ -254,6 +260,48 @@ def __enter__(self) -> AuthClient:
...
+
+        Bypasses `_do_request`/the shared retry loop -- the same deliberate
+        exception `poll_device_token` and `refresh` already make (see the
+        module docstring). The server consumes a TOTP time-slice on its
+        first submission and rejects any resubmission of it, so a 429/5xx
+        retry here would resend the same code and be rejected by
+        construction, burning one of the account's limited MFA attempts for
+        a failure that was never the credential's fault.
+        """
+        response = self._client.request(
```
zajca: Reviewed by Opus.

`verify_mfa_totp` calls `self._client.request(...)` directly and maps only responses (`status_code >= 400`); transport failures are not mapped. The other credential-bearing method that bypasses `_do_request`, `refresh`, routes through `_post_refresh` (auth_client.py:542-570), which converts `httpx.TimeoutException` to `ErrorCode.TIMEOUT` and `httpx.TransportError` to `ErrorCode.CONNECTION_ERROR`.

`commands/auth.py:475` catches only `(ConfigError, KeboolaApiError)` and there is no global `httpx` handler in `cli.py`, so an `httpx.ConnectError`/`ReadTimeout` on the MFA hop escapes as an unhandled traceback with no `--json` error envelope — the same structured-error contract violation this PR already fixed once for the raw `binascii.Error`. The first hop is unaffected: `_do_request` maps timeouts and connect errors for `login_password`, so only the MFA step is inconsistent.

Failure scenario: an MFA-enabled account signs in from CI, the password POST succeeds, and the network drops during `POST /v1/auth/mfa`. Instead of exit code 4 with `{"error": {"code": "CONNECTION_ERROR"}}`, the process prints a raw `httpx.ConnectError` traceback and the workflow step's JSON parsing of the output fails.

src/keboola_agent_cli/auth/auth_client.py
```diff
@@ -694,16 +742,28 @@ def _raise_api_error(self, response: httpx.Response, base_url: str | None = None
...
+        specific flow) is not enabled on this stack -- a fail-closed feature
+        flag -- not "wrong URL". Surfacing the generic `NOT_FOUND` code would
+        send the user chasing a routing bug that does not exist; this maps it
+        to `AUTH_NOT_SUPPORTED_ON_STACK` with a message naming the
+        static-token fallback instead. Every other status delegates to the
+        shared `BaseHttpClient` mapping -- no retry loop is added around the
+        404 case, since a disabled feature flag will not become enabled by
         retrying.
         """
+        if response.request is not None and response.request.url.path in (
```
zajca: Reviewed by Opus.

`httpx.Response.request` is a property that raises `RuntimeError("The request instance has not been set on this response.")` when `_request` is unset — it never returns `None` (httpx `_models.py:596-604`). So this guard, and the identical one in `_raise_password_login_error` (line 814), cannot do what they were written to do: on the exact input they defend against they raise `RuntimeError` rather than falling through to the generic mapping.

Every response reaching these methods today originates from `self._client.request(...)`, so production is unaffected — but the guard is dead code that gives false assurance, and this repo already constructs bare responses in tests (`tests/test_client.py:3865-3885` passes `httpx.Response(403, text=...)` straight into error mapping).

Failure scenario: a unit test (or any future non-client-originated path) calls `AuthClient._map_auth_error(httpx.Response(401, json={"error": "..."}))` following the pattern already used at tests/test_client.py:3865; instead of the intended `KeboolaApiError` the call raises `RuntimeError: The request instance has not been set on this response.`

`getattr(response, "_request", None) is not None`, or reading `response.request.url.path` inside a `try`, would actually hold.

src/keboola_agent_cli/auth/auth_client.py
```diff
@@ -721,6 +781,98 @@ def _map_auth_error(self, response: httpx.Response) -> NoReturn:
...
+            raise KeboolaApiError(
+                message=f"Too many failed login attempts.{when}",
+                status_code=429,
+                error_code=ErrorCode.API_ERROR,
+                retryable=False,
+            )
+        super()._raise_api_error(response, self._base_url)
+        raise AssertionError("unreachable: BaseHttpClient._raise_api_error always raises")
+
+    def _extract_api_message(self, response: httpx.Response) -> str:
```
zajca: Reviewed by Opus.

`_extract_api_message` re-implements body-message extraction that `AuthClient._extract_error_message` (line 698, same class, ~150 lines above) already provides. Its docstring justifies duplicating `BaseHttpClient._raise_api_error`'s private extraction but does not account for the sibling that already exists.

The two also diverge: `_extract_error_message` checks `error` (non-empty string only) → `exception` → `message` → `description` → `detail` → `errors` → `json.dumps(body)`, while the new one stops after `message`. A 403 whose body carries only `description`/`detail`/`errors` therefore renders as a raw JSON blob instead of the server's prose. The existing call site shows the intended composition: `revoke()` at line 695 does `self._truncate(self._extract_error_message(response))`, which is exactly what this method needs.

Failure scenario: the auth service returns 403 with `{"detail": "SSO is enforced for this organization"}`. `_extract_error_message` would yield that sentence; `_extract_api_message` falls through to `json.dumps(body)`, so the user reads ``{"detail": "SSO is enforced for this organization"} If this account requires SSO or its MFA cannot be resolved without a browser, use `kbagent auth login` instead.``

src/keboola_agent_cli/commands/context.py
```diff
@@ -94,6 +97,27 @@
     A 404 from any auth endpoint means browser login is not enabled on that
     stack yet (per-stack feature flag); use a static token instead.
 
+  kbagent auth login-password --email EMAIL --password PASSWORD [--totp-secret SECRET] [--stack URL|alias] [--register-projects]
```
zajca: Reviewed by Opus.

`--password-stdin` was added in the third commit of this PR but never propagated to the runtime agent surface. `AGENT_CONTEXT` here shows the synopsis as `--email EMAIL --password PASSWORD [--totp-secret SECRET] ...` and the following prose (lines 101-118) never mentions the flag; the same omission exists in `CLAUDE.md`'s "All CLI Commands" block and in the `login-password` bullet in `plugins/kbagent/skills/kbagent/references/commands-reference.md`. It *is* documented in `keboola-expert.md`, `auth-workflow.md` and `gotchas.md`, so the drift is partial.

Both omitted files are named in CLAUDE.md convention #17 as mandatory, non-CI-enforced surfaces. This matters more than ordinary doc drift because `kbagent context` is the first thing the skill tells an agent to run, and the flag it omits is precisely the one the command's own help text recommends to keep the password out of shell history and process listings. Compare the existing `dev-portal identity add` entry at CLAUDE.md:601, which does document its `--password-stdin`.

Failure scenario: an agent runs `kbagent context`, reads the synopsis at commands/context.py:100, and constructs `kbagent auth login-password --email svc@example.com --password "$PASS"` — putting the service account's password into the process argument list and shell history, the exact exposure `--password-stdin` was added to close.

src/keboola_agent_cli/auth/auth_client.py
zajca: Reviewed by Sonnet.

`verify_mfa_totp` deliberately bypasses `_do_request` (to avoid replaying a single-use TOTP code on retry) and calls `self._client.request(...)` directly, but unlike the other two methods that make the exact same bypass -- `_post_refresh` (used by `refresh()`) and `revoke`/`delete_session` -- it wraps nothing in try/except. `_post_refresh` explicitly catches `httpx.TimeoutException`/`httpx.TransportError` and maps them to structured `KeboolaApiError`s (auth_client.py ~544-570); `verify_mfa_totp` does not, and there is no test for this path analogous to `TestRefresh`'s connect-failure/timeout tests, or `TestVerifyMfaTotp`'s existing 429-no-retry test.

Concretely: `kbagent auth login-password --totp-secret ...` in a CI job where `POST /v1/auth/login` succeeds and returns an MFA challenge, but the follow-up `POST /v1/auth/mfa` hits a transient network blip (`httpx.ConnectError`/`ReadTimeout`) -- the exception propagates unmodified out of `AuthService.login_password` and past `commands/auth.py:auth_login_password`'s `except (ConfigError, KeboolaApiError)` (auth.py:479), which does not match it. The CLI then crashes with a raw Python traceback instead of the documented structured `--json` error envelope / one of the 0-5 exit codes -- exactly the class of bug this PR's own description calls out fixing for the malformed `--totp-secret` case (finding C1), reintroduced here on the network-failure path.

src/keboola_agent_cli/commands/auth.py
```diff
@@ -382,6 +385,102 @@ def _run_post_login_hook(
...
+    email: str = typer.Option(
+        ...,
+        "--email",
+        envvar=ENV_KBC_LOGIN_EMAIL,
+        help="Account email. Also settable via KBC_LOGIN_EMAIL.",
+    ),
+    password: str | None = typer.Option(
+        None,
+        "--password",
+        envvar=ENV_KBC_LOGIN_PASSWORD,
```
zajca, line 400: Reviewed by Sonnet.

`--password`/`KBC_LOGIN_PASSWORD` are read unconditionally (no `--allow-env-*`-style opt-in), while `KBC_MANAGE_API_TOKEN` is ignored by convention #12 unless `--allow-env-manage-token` is explicitly passed, specifically to close the AI-exfiltration risk where any subprocess (including the agent itself) inherits a powerful secret via env. The PR's own addendum (docs/programmatic-auth-login-plan.md, finding D2) flags this as 'a real, unresolved tension ... left as an open question for the reviewer rather than resolved unilaterally in this PR' -- and per docs/auth.md, a successful `login-password` on an MFA account grants a live 3-hour 'sudo' window, arguably a higher-privilege artifact than a single scoped Storage token. Shipping it without the same env-opt-in gate (or an explicit accepted-risk rationale beyond 'CI already sets these once') is a real security-consistency gap the PR leaves open by design; worth a maintainer decision before merge rather than silently defaulting to 'allowed'.

zajca (commented):

src/keboola_agent_cli/commands/auth.py
```diff
@@ -382,6 +385,102 @@ def _run_post_login_hook(
...
+
+    Storing an account's password (and TOTP seed) as CI secrets is a bigger
+    blast radius than a single scoped credential: whoever holds them can do
+    anything that account can do, not just what one project's token allows.
+    Use a dedicated, least-privileged service account for this, never a
+    real human's own credentials.
+    """
+    formatter = get_formatter(ctx)
+    service: AuthService = get_service(ctx, "auth_service")
+    if password_stdin:
```
zajca, line 459: Reviewed by Sonnet.

When both `--password` (or `KBC_LOGIN_PASSWORD`) and `--password-stdin` are supplied, `password_stdin` silently wins with no error or warning (auth.py:465-466). Every other multi-source-input command in this codebase treats that combination as a hard error, e.g. `_metadata_input.py:41` ('--text, --file, and --stdin are mutually exclusive'), `config.py:1313` ('--configuration and --configuration-file are mutually exclusive'). A CI author who sets `KBC_LOGIN_PASSWORD` in the step env and also passes `--password-stdin` gets no feedback that the env value was ignored -- inconsistent with the codebase's established mutual-exclusion convention, though low-impact since the safer input source (stdin) is the one that wins.

Matovidlo:
Thanks — this round caught two real bugs the first round's fixes introduced (O001, O002/S001) plus a genuinely dead guard (O003). Cross-validated against a thermo-nuclear structural pass + ponytail-review before touching anything; both independently converged on O001–O004, and ponytail additionally caught `_read_password_stdin` duplicating `dev_portal.py`'s copy verbatim (not in your list, fixed alongside).

**O001 (login_password retries through the shared loop)** — Fixed. `login_password` now bypasses `_do_request` the same way `verify_mfa_totp` already did, via a new shared `_request_bypassing_retry`.

**O002 / S001 (no transport-error mapping on the bypassed calls — same bug, two reviewers)** — Fixed as part of the same refactor. `_request_bypassing_retry` extracts `_post_refresh`'s existing try/except shape (`TimeoutException`→`TIMEOUT`, `TransportError`→`CONNECTION_ERROR`) and is now shared by `login_password`, `verify_mfa_totp`, and `refresh()` itself. New tests mirror `TestRefresh`'s timeout/connect-error pair for both.

**O003 (`response.request is not None` guard is dead code)** — Fixed by deletion rather than patched. Went with the fuller move: `login_password` now calls `_raise_password_login_error` directly instead of routing through `_map_auth_error`'s path-based dispatch, and `is_mfa` is passed explicitly by the caller instead of sniffed from `response.request.url.path`. Both guards — and every `response.request` reference in the file — are gone; `_map_auth_error` is back to purely the generic mapping it was before this PR touched it. New regression test constructs a bare `httpx.Response(...)` (the exact shape that used to crash) and asserts it doesn't.

**O004 (`_extract_api_message` duplicates `_extract_error_message`)** — Fixed. Deleted; the one call site now reuses `_extract_error_message`, matching `revoke()`'s existing usage.

**O005 (`--password-stdin` doc drift)** — Fixed in the three surfaces that were missing it: `commands/context.py`'s `AGENT_CONTEXT`, `CLAUDE.md`'s "All CLI Commands" block, `commands-reference.md`'s `login-password` bullet.

**S003 (`--password-stdin` silently overrides `--password`/env)** — Fixed. Now raises `ConfigError` if both are supplied, matching `_metadata_input.py`'s `--text`/`--file`/`--stdin` convention. Left `dev-portal identity add`'s pre-existing identical silent-override behavior untouched — out of scope here, flagged as a latent inconsistency in the commit message for whoever picks it up.

**Bonus (ponytail-review, not in your findings)**: `commands/auth.py`'s `_read_password_stdin()` was byte-identical to the one already in `commands/dev_portal.py` — same docstring, same body. Extracted to `commands/_helpers.py::read_password_stdin`, both command modules now share it.

**S002 / D2 (env-var gating for `--password`/`KBC_LOGIN_PASSWORD`)** — Still open, same disposition as round 1: I want your read on how far to take this (keep as documented tradeoff / soft-deprecate `--password` / go full `--allow-env-*` like the manage token) before touching it myself.

One unrelated thing surfaced while getting `make check` green after these fixes: this branch had drifted 10 commits behind `main`, which had independently released its own (unrelated) `0.81.0` — a real version collision with this PR's own unreleased `0.81.0` bump. Rebased onto `main` and renumbered this PR's version to `0.83.0` (next free slot), same pattern as the existing `chore(release): renumber the unreleased 0.77.1 to 0.78.0` precedent in this repo's history. Everything force-pushed; CI green on all four checks.

Ready for another pass.

zajca (approved):
Approved as requested. The remaining low-severity finding was sent as a non-blocking review comment.

zajca (approved):
Approved as requested. The remaining low-severity finding could not yet be published because GitHub rejected the inline review request; it will need to be retried separately.

zajca:
Reviewed by Sonnet.

The new constant `JOB_IDEMPOTENCY_LOCK_FILENAME = "job_idempotency.json.lock"` is added here, but `JobIdempotencyStore.__init__` (`services/job_idempotency_store.py`) builds the actual lock path independently via `self._path.with_name(self._path.name + ".lock")` and never imports or references this constant. Today the two happen to produce the same string, but the constant is dead code that documents an assumption the implementation does not enforce — if `JOB_IDEMPOTENCY_FILENAME` or the lock-path derivation ever changes, this constant will silently drift out of sync with the real lock filename with nothing pointing at the mismatch. Either wire `_lock_path` through this constant or remove it.

PR #579, opened 2026-08-12
fix(lineage): warn when an unqualified table id resolves to several projects (#568)

### LINKED ISSUE #568: lineage show silently drops results for tables shared across multiple projects instead of warning as documented
**Summary**

`kbagent lineage show --downstream/--upstream <table_id>` (without an explicit `project:` prefix) silently resolves to only ONE project's copy of the table when the same table_id exists as a node in more than one project's graph (e.g. via bucket sharing/linking across layer projects). No warning is printed, and the command exits 0 with output that looks complete.

This contradicts the command's own `--help` text:

> Table only: `bucket_id.table_name` (auto-resolves, **warns if ambiguous**)

**Steps to reproduce**

1. Have two projects (A, B) where a bucket is shared from A into B, so the same `bucket_id.table_name` exists as a lineage node in both `A:` and `B:` namespaces, and a config in project B has an actual input-mapping dependency on it that doesn't exist in project A.
2. `kbagent lineage build -d <dir> -o lineage.json` (with both projects synced under <dir>).
3. `kbagent lineage show -l lineage.json --downstream "bucket_id.table_name"` (no project prefix).

**Expected:** either a warning that the table id is ambiguous across N projects, or results merged/listed per project.

**Actual:** returns only project A's downstream edges, silently. Project B's real dependency (confirmed to exist in the graph's `edges` array and reproducible when queried as `--downstream "B:bucket_id.t

padak:
## What

`kbagent lineage show --upstream/--downstream <table_id>` silently answered for a single project when the same unqualified `bucket_id.table_name` existed in several project namespaces (a bucket shared from one project and linked into another produces one node per project). `_find_node()` did `return sorted(matches)[0]` and threw the other candidates away, so the command exited 0 with output that looked complete — while contradicting its own `--help` text: *"Table only: `bucket_id.table_name` (auto-resolves, **warns if ambiguous**)"*.

## Why

Real dependencies were invisible: a config in project B reading the shared table did not show up at all, and nothing told the user to retry with `B:bucket_id.table_name`. A lineage answer that is quietly partial is worse than one that is loudly incomplete — "is anything downstream of this table" is exactly the question people ask before dropping it.

## How

- `DeepLineageService._find_node_candidates()` (new) returns **every** FQN an identifier could refer to, best match first. `_find_node()` stays a thin wrapper returning the first, so node selection is unchanged and deterministic.
- `query_upstream()` / `query_downstream()` now share one `_query()` body and, when more than one candidate matched, attach two additive keys to the result: `ambiguous_matches` (the full candidate list) and `warnings` (a human-readable line naming the projects and how to disambiguate).
- `commands/lineage.py` prints those warnings via `formatter.warning()` — stderr in human mode, no-op in JSON mode where they already ride along in the payload. stdout stays byte-clean for `--format mermaid/html/er` piping.
- An explicit `project:table` FQN or `--project ALIAS` is unambiguous by construction and stays silent.

The two near-identical `if upstream:` / `if downstream:` blocks in the command are folded into one loop — behaviour identical (upstream first, then downstream), and it pays for the added lines so `commands/lineage.py` stays inside its grandfathered file-size budget.

## Testing

- New `TestAmbiguousNodeResolution` (6 cases): warns and lists both candidates for downstream and upstream; explicit FQN, `--project` scope, and a genuinely unique id stay silent; an unknown id still errors.
- New CLI cases: human mode puts `Warning: ... exists in 2 projects` on **stderr**; `--json` carries `ambiguous_matches` + `warnings` inside the envelope and stdout still parses.
- `make check` green: 5455 passed, 12 skipped; lint/format/ty/loc-check clean.

No CLI surface change (no new command, flag or error code), so no changelog entry and no plugin/agent doc sync required.

Fixes #568

### DIFF
--- a/src/keboola_agent_cli/commands/lineage.py
+++ b/src/keboola_agent_cli/commands/lineage.py
@@ -340,8 +340,13 @@ def lineage_show(
 
     display_opts = {"show_columns": columns, "filter_column": column}
 
-    if upstream:
-        query_result = service.query_upstream(graph, upstream, project or "", depth)
+    # Both directions render identically; only the service call and the label
+    # differ. Passing both --upstream and --downstream emits both, in order.
+    for direction, identifier in (("upstream", upstream), ("downstream", downstream)):
+        if not identifier:
+            continue
+        query = service.query_upstream if direction == "upstream" else service.query_downstream
+        query_result = query(graph, identifier, project or "", depth)
         if "error" in query_result:
             suggestions = query_result.get("suggestions", [])
             msg = query_result["error"]
@@ -350,36 +355,30 @@ def lineage_show(
             formatter.error(message=msg, error_code=ErrorCode.NODE_NOT_FOUND)
             raise typer.Exit(code=1)
 
-        if formatter.json_mode:
-            if column:
-                query_result = _filter_column_json(query_result, column)
-            formatter.output(query_result)
-        elif format in ("mermaid", "html", "er"):
-            _output_mermaid_or_html(formatter, service, graph, query_result, "upstream", format)
-        else:
-            _format_lineage_tree(formatter, graph, query_result, "upstream", **display_opts)
-
-    if downstream:
-        query_result = service.query_downstream(graph, downstream, project or "", depth)
-        if "error" in query_result:
-            suggestions = query_result.get("suggestions", [])
-            msg = query_result["error"]
-            if suggestions:
-                msg += "\nDid you mean: " + ", ".join(suggestions[:5])
-            formatter.error(message=msg, error_code=ErrorCode.NODE_NOT_FOUND)
-            raise typer.Exit(code=1)
+        _emit_query_warnings(formatter, query_result)
 
         if formatter.json_mode:
             if column:
                 query_result = _filter_column_json(query_result, column)
             formatter.output(query_result)
         elif format in ("mermaid", "html", "er"):
-            _output_mermaid_or_html(formatter, service, graph, query_result, "downstream", format)
+            _output_mermaid_or_html(formatter, service, graph, query_result, direction, format)
         else:
-            _format_lineage_tree(formatter, graph, query_result, "downstream", **display_opts)
+            _format_lineage_tree(formatter, graph, query_result, direction, **display_opts)
 
 
 # -- Output formatting helpers ----------------------------------------------
+
+
+def _emit_query_warnings(formatter, query_result: dict) -> None:
+    """Print non-fatal warnings carried by a lineage query result.
+
+    Currently only the ambiguous-identifier warning (the same table id exists
+    in several projects). ``formatter.warning`` is a no-op in JSON mode --
+    there the warnings ride along in the emitted payload instead.
+    """
+    for warning in query_result.get("warnings", []):
+        formatter.warning(warning)
 
 
 def _output_mermaid_or_html(
--- a/src/keboola_agent_cli/commands/lineage.py
+++ b/src/keboola_agent_cli/commands/lineage.py
@@ -20,6 +20,7 @@ from urllib.parse import parse_qs, urlparse
 import typer
 
 from ..errors import ErrorCode
+from ..output import OutputFormatter
 from ..services.deep_lineage_service import DeepLineageService, LineageGraph
 from ._helpers import (
     check_cli_permission,
@@ -370,7 +371,7 @@ def lineage_show(
 # -- Output formatting helpers ----------------------------------------------
 
 
-def _emit_query_warnings(formatter, query_result: dict) -> None:
+def _emit_query_warnings(formatter: OutputFormatter, query_result: dict) -> None:
     """Print non-fatal warnings carried by a lineage query result.
 
     Currently only the ambiguous-identifier warning (the same table id exists
--- a/src/keboola_agent_cli/services/deep_lineage_service.py
+++ b/src/keboola_agent_cli/services/deep_lineage_service.py
@@ -371,14 +371,44 @@ class LineageGraph:
 # ---------------------------------------------------------------------------
 
 
+# Keep a warning readable when a bare table name matches across many buckets.
+_MAX_LISTED_CANDIDATES = 5
+
+
+def _format_candidates(items: list[str]) -> str:
+    """Join ids for display, trimming a long tail rather than printing all of it."""
+    head = items[:_MAX_LISTED_CANDIDATES]
+    tail = f", +{len(items) - len(head)} more" if len(items) > len(head) else ""
+    return ", ".join(head) + tail
+
+
 def _ambiguity_warning(identifier: str, candidates: list[str]) -> str:
-    """Build the warning shown when an identifier matches several projects."""
-    projects = [fqn.split(":", 1)[0] for fqn in candidates]
+    """Build the warning shown when an identifier matches more than one node.
+
+    Two different shapes reach this. The one #568 reports is the same
+    ``bucket_id.table_name`` living in several project namespaces (a bucket
+    shared from one project and linked into another) -- there every candidate
+    differs only by project, and ``<project>:<identifier>`` is a retry that
+    resolves. The other comes from the name-only fallback in
+    ``_find_node_candidates``, which matches a bare table name across buckets:
+    those candidates can share a single project, so counting them as projects
+    would tell the user a table "exists in 2 projects (alpha, alpha)", and
+    ``<project>:<identifier>`` would not resolve because the real node ids
+    carry a bucket. That case gets the full ids instead.
+    """
+    shown = candidates[0]
+    if all(fqn.partition(":")[2] == identifier for fqn in candidates):
+        projects = sorted({fqn.partition(":")[0] for fqn in candidates})
+        return (
+            f"'{identifier}' exists in {len(projects)} projects "
+            f"({_format_candidates(projects)}); showing '{shown}' only. "
+            f"Query a specific one with '--upstream/--downstream "
+            f"<project>:{identifier}' or scope with --project."
+        )
     return (
-        f"'{identifier}' exists in {len(candidates)} projects "
-        f"({', '.join(projects)}); showing '{candidates[0]}' only. "
-        f"Query a specific one with '--upstream/--downstream "
-        f"<project>:{identifier}' or scope with --project."
+        f"'{identifier}' matches {len(candidates)} nodes "
+        f"({_format_candidates(candidates)}); showing '{shown}' only. "
+        f"Query a specific one by its full id, e.g. '--upstream/--downstream {shown}'."
     )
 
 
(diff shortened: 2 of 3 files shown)


padak (commented):
# Review of #579 — fix(lineage): warn when an unqualified table id resolves to several projects (#568)

## Summary

This PR fixes issue #568: `lineage show --upstream/--downstream <table>` used to silently pick the alphabetically-first project when an unqualified `bucket_id.table_name` matched nodes in several projects, contradicting the command's own `--help` text ("auto-resolves, warns if ambiguous"). The fix unifies node resolution behind a new `_find_node_candidates()` that returns every match, and `query_upstream`/`query_downstream` (now sharing one `_query()` body) attach additive `ambiguous_matches` + `warnings` keys when more than one candidate is found; `commands/lineage.py` prints the warning via `formatter.warning()` (stderr in human mode, no-op in JSON — where it rides along in the payload). **Verdict: does resolve issue #568 as described** — reproduced live (see Verification log): the reporter's exact scenario (shared table across two projects, one has a real dependency the other lacks) now returns the same first-candidate answer as before, but with `ambiguous_matches` listing both projects and a `warnings` line telling the caller how to disambiguate. I found no BLOCKING issues; a handful of NON-BLOCKING gaps (a REST sub-endpoint that silently drops the new warning, dead code left behind by the refactor, and thin doc-sync) are worth the author's attention but do not block merge. **Overall verdict: COMMENT.**

## Verdict

- **Verdict**: COMMENT
- **Blocking findings**: 0
- **Non-blocking findings**: 5
- **Nits**: 2

## Blocking findings

(none)

## Non-blocking findings

### `[NB-1]` `src/keboola_agent_cli/server/routers/lineage.py:193-226` — `GET /lineage/mermaid` silently drops the new ambiguity warning

I verified empirically (see Verification log) that `POST /lineage/show` and `GET /lineage/walk` pass `ambiguous_matches`/`warnings` straight through to `kbagent serve` clients — the router returns the service dict verbatim with no `response_model` filtering, so this PR's fix is **not** CLI-only as originally suspected. However, `GET /lineage/mermaid` (line 211-214) calls `query_upstream`/`query_downstream` and then only reads `result.get("edges", [])` to render the diagram — the `warnings`/`ambiguous_matches` keys are computed and then discarded, and the endpoint returns `PlainTextResponse` with no side channel to carry them. A React-UI user viewing the Mermaid graph for an ambiguous table gets a silently-partial diagram with zero indication, the exact failure mode #568 describes, just on a different surface. This PR didn't touch `server/routers/lineage.py` at all (confirmed unchanged), so this gap pre-dates it structurally but is now live for a case that previously never warned anywhere. Consider at minimum a response header (e.g. `X-Lineage-Ambiguous: 2`) or documenting the gap.

### `[NB-2]` `src/keboola_agent_cli/services/deep_lineage_service.py:1036-1038` (PR branch) — `_find_node()` is now dead code

The PR's own description says `_find_node()` "stays a thin wrapper returning the first, so node selection is unchanged and deterministic." I verified this is not true: `query_upstream`/`query_downstream` now call `_query()`, which calls `self._find_node_candidates(...)` directly — `_find_node()` is never called anywhere in the diff or the rest of the tree (confirmed by grepping every `.py` file in the PR branch for `._find_node(`, zero hits outside its own definition). `ruff`/`ty` don't catch unused private methods, so `make check` stays green. Either remove `_find_node()` or restore a real caller; leaving it invites a future reader to "fix" a code path that no longer runs.

### `[NB-3]` `src/keboola_agent_cli/services/deep_lineage_service.py:213-220` (PR branch) — partial-match fallback branch (`f.split(":")[-1].endswith(...)`) has no dedicated ambiguity test

Per the focus question: I confirmed by direct reproduction (see Verification log) that the *partial* match branch (queries like `"orders"` with no bucket prefix, matched via `f.split(":")[-1].endswith(f".{identifier}")`) is **also** fixed — because `_find_node_candidates()` unifies both the exact-suffix and partial-suffix branches into one return path that `_query()` treats uniformly. This is a genuine strength of the refactor. However, `TestAmbiguousNodeResolution` (the new test class) only exercises the exact `bucket_id.table_name` branch (`SHARED_TABLE = "in.c-shared.orders"`, always matched via the `f.endswith(f":{identifier}")` path) — there is no test with two *different* buckets sharing only a bare table name across projects, which is the only way to exercise the partial-match branch. Add one case to lock in the behavior I verified manually.

### `[NB-4]` `plugins/kbagent/skills/kbagent/references/gotchas.md:1785` — "Response structure varies by command" table entry for `lineage show` is stale, and this PR adds to the staleness

The table's `lineage show` row reads `{"lineage_links": [...], "errors": [...]}`, which does not match the actual shape (`node`, `direction`, `node_info`, `edges`, and now additively `ambiguous_matches`/`warnings`) returned by `query_upstream`/`query_downstream` — this mismatch pre-dates the PR, but the new additive keys make it further out of sync. The PR description states "No CLI surface change ... no plugin/agent doc sync required," which is true in the narrow sense (no new command/flag/error code), but the new JSON keys are exactly the kind of thing AI-agent consumers benefit from knowing about per CONTRIBUTING.md's Plugin synchronization map (gotchas.md, "always tag with `(since vX.Y.Z)`"). I did not find a version number assigned to this PR yet (no `pyproject.toml` bump in the diff), so a precise `(since vX.Y.Z)` tag isn't possible until release time — I found (during verification, outside this PR's diff) that `src/keboola_agent_cli/changelog.py` on the target branch already carries a drafted "Fix (#568)" bullet under a pending `0.84.0` entry, suggesting doc-sync for this fix is already being tracked at the release step per CONTRIBUTING.md's checklist. Flagging this NON-BLOCKING rather than BLOCKING for that reason — but the gotchas.md table row itself should be corrected (or removed) whenever that release-time pass happens.

### `[NB-5]` `tests/test_deep_lineage_service.py` — no E2E case for the ambiguous-table scenario

`tests/test_e2e_lineage_deep.py` has existing `lineage show` coverage but nothing that builds two projects with a shared bucket to exercise the new `ambiguous_matches`/`warnings` path end-to-end. Per CONTRIBUTING.md, "every CLI command must have E2E coverage" — this is a behavior fix on an existing command rather than a new command, and standing up a real cross-project bucket-share fixture is a genuinely heavier E2E lift than most fixes, so I'm not blocking on it. The unit (`TestAmbiguousNodeResolution`, 6 cases) and CLI-layer (`CliRunner`, stderr + JSON) coverage in this PR is solid and gives good confidence regardless.

## Nits

- `[NIT-1]` `src/keboola_agent_cli/services/deep_lineage_service.py` — this file was already over its 1000-code-line soft ceiling on `main` (1017 lines) before this PR; the PR's net addition pushes it to 1031. Not a blocker (soft ceiling only warns, `make loc-check` stays green), but worth keeping in mind for the next PR that touches this file.
- `[NIT-2]` `src/keboola_agent_cli/commands/lineage.py:60-68` — `_emit_query_warnings(formatter, query_result: dict)` leaves `formatter` untyped, matching the pre-existing (also untyped) pattern in sibling helpers `_output_mermaid_or_html`/`_format_lineage_tree` in the same file, so not new drift introduced by this PR — just noting the file as a whole could use an `OutputFormatter` type hint on all three someday.

## Verification log

- Read `CONTRIBUTING.md` "Plugin synchronization map", "Checklist: Adding a New CLI Command", "Releasing a new version"; `CLAUDE.md` convention #17 and `## All CLI Commands`; `plugins/kbagent/agents/keboola-expert.md` §1-3 — no `lineage` write/destructive group changes needed (all 4 `lineage.*` ops are `read` in `OPERATION_REGISTRY`).
- `gh issue view 568 --repo keboola/cli` → confirmed the exact reported scenario (bucket shared A→B, unqualified id resolves to A only, B's real dependency invisible, no warning, exit 0).
- `gh pr view 579 --json title,body,files,additions,deletions,...` → 3 files, +263/-53, conventional `fix(lineage):` ✓ matches change type (bug fix, no new surface).
- `gh pr diff 579` → read in full; confirmed CLI + service + test changes only, `server/routers/lineage.py` untouched.
- **Isolated worktree** (`git worktree add --detach ... origin/claude/issue-568-lineage-ambiguous-warning`, own `.venv` via `uv sync --extra server`), removed after use — shared working tree at `.claude/worktrees/issues-568-569-resolution-33f4e9` was never checked out or mutated (`git status`/`git rev-parse --abbrev-ref HEAD` confirmed clean and on its own branch before and after).
- `uv run pytest tests/test_deep_lineage_service.py -q` → 53 passed.
- `make check` (background, full run) → exit 0; `5455 passed, 12 skipped, 147 deselected` — matches the PR description's own claimed numbers exactly.
- `ruff check`, `ruff format --check`, `ty check` (targeted at the two changed source files), `make loc-check`, `make command-sync-check`, `make check-error-codes`, `make check-sentinel-guards`, `make version-check` → all clean/OK.
- **Focus point 1 (REST parity)** — built a FastAPI `TestClient` against `create_app()` with a synthetic two-project shared-table lineage cache and hit the real endpoints:
  - `POST /lineage/show {"downstream": "in.c-shared.orders"}` → response body **includes** `"ambiguous_matches": ["alpha:...", "beta:..."]` and `"warnings": [...]`. ✓ reaches REST.
  - `GET /lineage/walk?node=...&direction=downstream` → same, keys present. ✓ reaches REST.
  - `GET /lineage/mermaid?node=...&direction=downstream` → 200, `text/plain` Mermaid source only, no warning anywhere in body or headers. ✗ does NOT reach this one endpoint (see NB-1).
- **Focus point 2 (`--help`/doc promise)** — read `commands/lineage.py:292` (unchanged docstring: "Table only: bucket_id.table_name (auto-resolves, warns if ambiguous)") and `lineage-deep-workflow.md:110` (same claim). Confirmed via `output.py:197-207` that `formatter.warning()` writes to a `stderr=True` Rich console and is a no-op in JSON mode — matches the PR's stated stdout-clean-for-mermaid-piping design. **The documented promise is now true** where it was false before this PR; no doc text change needed on those two files (see NB-4 for the separate stale-schema-table issue in `gotchas.md`).
- **Focus point 3 (partial-match branch)** — wrote a standalone repro script building two projects with *different* buckets (`in.c-sales.orders`, `in.c-crm.orders`) sharing only the bare table name `orders`, then called `service.query_downstream(graph, "orders")` directly against the PR branch's service. Result: `ambiguous_matches: ['alpha:in.c-sales.orders', 'beta:in.c-crm.orders']` and a populated `warnings` list — confirms the partial-match fallback is fixed too, not just the exact-suffix branch (see NB-3 for the test-coverage gap this uncovered).
- `grep -rn "._find_node("` across every `.py` file in the PR branch tree → only the definition itself, no callers (see NB-2).
- Confirmed `OPERATION_REGISTRY` already has `lineage.build/info/show/server` all as `"read"` — no registry change needed since no new command/flag was added.

- The changelog draft I found (outside this PR's diff, in the local repo state) already stages a "Fix (#568)" bullet under a pending future version. Worth confirming with whoever owns that release-prep step that `gotchas.md`'s stale `lineage show` response-shape row (NB-4) gets corrected in the same pass, since it's adjacent housekeeping the release checklist already touches.

PR #580, opened 2026-08-12
fix(search): make --search-type config-based case-insensitive (#569)

### LINKED ISSUE #569: search --search-type config-based is case-sensitive, causing false negatives on "is this referenced anywhere" queries
**Summary**

`kbagent search "<term>" --search-type config-based` does a literal, case-sensitive substring match against stringified configuration JSON. Because Keboola configs routinely reference the same logical table under different casing (e.g. a component config's row/display name in mixed case vs. the physical/output table name in the resulting `storage.input.tables[].source` string, which is often all-caps for DB-sourced tables), a case-sensitive search silently misses real references.

**Steps to reproduce**

1. Have an extractor config whose row/display name is mixed-case (e.g. `DCFAmount`) but whose resulting Storage table name is uppercase (`DCFAMOUNT`, common for SQL Server sourced tables).
2. Have a second, unrelated config elsewhere in the org reference that table by its uppercase Storage name in `storage.input.tables[].source` (e.g. `in.c-bucket.DCFAMOUNT`).
3. Run `kbagent search "DCFAmount" --search-type config-based` (org-wide or scoped).
4. Run `kbagent search "DCFAMOUNT" --search-type config-based` with only the casing changed.

**Expected:** identical result sets -- a "does this appear anywhere in configuration JSON" search should not depend on casing, especially since the tool's own textual-search mode is documented as "case- and diacritics-insensitive."

**Actual:** the two queries returned different result counts (5 vs. 13 org-wide in our case), with the mixed-case query missing a real, actively-scheduled daily transformation dependency that the uppercase query found.

**Suggested fix:** make `--search-type config-based` case-insensitive by default (matching the behavior already documented for textual search), or at minimum document the case sensitivity prominently in `--help` so users know to try multiple casings before treating a "no results" as "not used anywhere."

---
**Environment**
- OS: Windows 11 Home 10.0.26200
- Shell: Git Bash / PowerShell
- kbagent version: v0.79.0 (auto-updating to v0.80.0 at time of testing; not yet confirmed whether v0.80.0 changes this behavior)

padak:
## What

`kbagent search "<term>" --search-type config-based` matched configuration JSON bodies **case-sensitively**, while the textual mode of the same command matches case-insensitively. `SearchService._search_config_based()` called `ConfigService.search_configs()` without `ignore_case`, so the parameter defaulted to `False` and there was no flag on `kbagent search` to override it.

## Why

The same logical table is routinely spelled two ways inside one project: an extractor row named `DCFAmount` produces a Storage table `DCFAMOUNT` (typical for SQL Server / Snowflake sources), and a downstream transformation references the upper-case id in `storage.input.tables[].source`. A case-sensitive scan therefore answers *"is this referenced anywhere?"* with a false **no** — the reporter saw 5 vs. 13 org-wide results for the same term differing only in casing, with the mixed-case query missing an actively-scheduled daily dependency. Silent under-reporting is the worst failure mode for a query people run before deleting or refactoring something.

## How

- `SearchService._search_config_based()` now passes `ignore_case=True`. One line of behaviour; the matcher itself (`ConfigService.search_configs`) already supported it.
- No new flag. `kbagent search` never exposed case control, and `kbagent config search --query` remains the case-sensitive-by-default body-scan surface (it has its own `--ignore-case` / `-i`). Adding a third knob to express what a sibling command already expresses did not seem worth the CLI surface.
- Direction of change is safe: case-insensitive matching can only return a superset of the previous results, never fewer.

**Alternative considered and rejected:** the issue offers "or at minimum document the case sensitivity in `--help`". That leaves every existing user's past "no results" conclusions wrong, so aligning the two modes seemed clearly better — happy to flip to docs-only if you disagree.

## Docs

Updated every silent-drift surface that describes this command: `CLAUDE.md` command list, `context.py` `AGENT_CONTEXT`, `commands-reference.md`, `gotchas.md`, and the `keboola-expert.md` tool-selection matrix (now 61623 bytes, within the 62000 budget).

## Testing

- `test_config_based_finds_differently_cased_table_reference` — the reporter's scenario against the **real** matcher (mocked client, not a mocked `ConfigService`): a config referencing `in.c-finance.DCFAMOUNT` is found by `DCFAmount`, `DCFAMOUNT` and `dcfamount` alike. Verified this test fails on the pre-fix code.
- `test_config_based_asks_config_service_for_case_insensitive_match` — pins the delegation contract so a future refactor cannot silently drop the flag.
- `make check` green: 5449 passed, 12 skipped; lint/format/ty/loc-check/command-sync clean.

No CLI surface change (no command, flag or error code added), so no changelog entry — the release-time backfill covers it.

Fixes #569


padak (commented):
# Review of #580 — fix(search): make --search-type config-based case-insensitive (#569)

## Summary

This is a small, well-tested, surgically-scoped fix: `SearchService._search_config_based()` now passes `ignore_case=True` to `ConfigService.search_configs()`, making `kbagent search --search-type config-based` match case-insensitively, mirroring the already-case-insensitive `textual` mode. **This does resolve issue #569 as reported**: the reporter's acceptance criterion was "identical result sets for the same term differing only in casing" — verified below with the reporter's exact scenario (`DCFAmount` row name vs. `DCFAMOUNT` physical table id), and confirmed the added regression test fails on pre-fix code and passes after. The design choice to hardcode the behavior (no new `--ignore-case` flag on `kbagent search`) rather than add a flag is reasonable and adequately mitigated by the pre-existing `kbagent config search --query -i` surface, which intentionally stays case-sensitive by default. However, the PR has one concrete, independently-verified BLOCKING problem: its own "within the 62000-byte budget" claim is now stale relative to current `main` — three other PRs landed on `keboola-expert.md` since this branch diverged, and a real `git merge-tree` simulation shows the merged file lands at **62119 bytes**, over budget. Overall verdict: **REQUEST CHANGES**, purely on the stale byte-budget artifact of merge timing, not on the core fix.

## Verdict

- **Blocking findings**: 2
- **Nits**: 1

## Blocking findings

### `[B-1]` `plugins/kbagent/agents/keboola-expert.md:100` — merging this PR as-is pushes the file over its 62000-byte hard budget

The PR description claims the file is "now 61623 bytes, within the 62000 budget" — true against the PR's own stale base (merge-base `73ded7a`, 61494 bytes), but **three PRs have landed on `main`'s copy of this file since this branch diverged** (`5f5fbae` docs(mcp) no-migration-command, `8ee25c9` docs(auth) login-password, and `7445d76` — a dedicated *"trim keboola-expert.md back under the 62000-byte budget"* commit). Current `main`'s copy is already at 61990 bytes, i.e. **10 bytes of headroom**. I ran an actual `git merge-tree --write-tree origin/main origin/claude/issue-569-config-search-case-insensitive` (a real 3-way merge, not arithmetic) and extracted the resulting blob directly: `git show <merged-tree>:plugins/kbagent/agents/keboola-expert.md | wc -c` → **62119 bytes**. The merge is textually clean (`mergeable: MERGEABLE`), so nothing will warn the author or a merger — it will silently exceed budget. Per CONTRIBUTING.md ("do **not** raise the cap... trim stale content first") this is exactly the highest-value silent-drift class `keboola-expert.md` is called out for. Fix: before merging, rebase onto current `main` and trim ~120+ bytes elsewhere in the file (or shorten the new "Search config JSON bodies" row text) to stay under budget, then re-verify with `wc -c`.

### `[B-2]` `plugins/kbagent/skills/kbagent/references/gotchas.md:2629` — new case-insensitivity behavior is tagged with an issue number, not a `(since vX.Y.Z)` version

The amended sentence reads "It matches **case-insensitively** (issue #569) while `config search --query` stays case-sensitive..." — no version tag. Per CONTRIBUTING.md ("gotchas.md — always tag with `(since vX.Y.Z)`... non-optional") and CLAUDE.md convention #17, this is mandatory precisely because an AI agent reading this file cannot tell whether a given user's installed `kbagent` (pinned via `kbagent --json version`, per keboola-expert.md Rule 6 VERSION GATE) already has this fix. The enclosing `## ... (since v0.30.0)` heading refers to when `search` became a top-level command, not to this behavior change, so it doesn't cover the gap. This PR contains no version bump (current `pyproject.toml` is `0.84.0`, unreleased) — other recent same-situation PRs in this repo's history (e.g. the `login-password` docs commit `8ee25c9`) speculatively tag with the anticipated next version (e.g. `0.81.0+`) rather than omitting the tag. Fix: replace "(issue #569)" with a concrete "(since v0.85.0+)" (or whatever the next release turns out to be) — same gap exists more mildly in the `keboola-expert.md` §2 matrix row and the `CLAUDE.md`/`context.py` comments, which are NON-BLOCKING per the reviewer severity table but worth fixing in the same pass.

## Non-blocking findings

### `[NB-1]` `src/keboola_agent_cli/services/search_service.py:274` — case-sensitivity on `search --search-type config-based` is removed with no escape hatch, worth an explicit maintainer sign-off

Verified (focus point 1): `ignore_case=True` is hardcoded with no CLI flag to override it, so `kbagent search --search-type config-based` can no longer do a case-SENSITIVE body scan at all. I judge the tradeoff acceptable — the issue explicitly asked for case-insensitive-by-default, the change is a strict superset of prior results (never fewer matches), and `kbagent config search --query -i` remains the case-sensitive-capable sibling surface for callers who need exact matching. This is not a functional regression, but it is a one-way door: if a future caller genuinely needs case-sensitive matching through the unified `search` command (e.g. distinguishing `Orders` from `orders` configs), there is no way back short of a new flag. Flagging for explicit maintainer agreement rather than blocking, since the PR description already surfaces this tradeoff and explicitly invites disagreement ("happy to flip to docs-only if you disagree").

### `[NB-2]` `plugins/kbagent/skills/kbagent/references/gotchas.md:2629` — missing context that `--regex` was the only prior case-insensitivity workaround, and it's blocked on this surface

Verified (focus point, extra check): `commands/search.py:131-134` still rejects `--regex` combined with `--search-type config-based` (exit 2, unchanged by this PR) — confirmed via `gh pr diff` and a direct read of the PR-branch file. Before this fix, users had **no** workaround at all for case-insensitive config-body search (no `--ignore-case` flag, and `(?i)` regex explicitly blocked on this surface) — worth one added clause in the gotchas.md entry so future readers understand why this was a real gap and not just a minor inconvenience. Not blocking; purely additive context.

## Nits

- `[NIT-1]` PR description — "No CLI surface change... so no changelog entry" is correct per this repo's actual practice (changelog entries are added at release-bump time, not per-PR, confirmed by `CONTRIBUTING.md` "Releasing a new version" step 2 and multiple prior release commits) — flagging only so the reviewer/release-owner remembers to backfill `changelog.py` at the next version bump; this PR's behavior change is exactly the kind of thing CONTRIBUTING.md's step 6 wants captured then.

## Verification log

- `gh issue view 569 --repo keboola/cli` → confirmed reporter's exact scenario (mixed-case row name `DCFAmount` vs. upper-case Storage table `DCFAMOUNT`) and the acceptance bar ("identical result sets... case-insensitive by default, matching the documented textual-search behavior").
- `gh pr view 580 --repo keboola/cli --json title,body,files,additions,deletions,...` → 8 files, +88/-5, `state: OPEN`, conventional `fix(search):` prefix matches a bug fix ✓.
- `gh pr diff 580` (202 lines) read in full; confirmed every hunk (CLAUDE.md, keboola-expert.md, commands-reference.md, gotchas.md, context.py, commands/search.py, services/search_service.py, tests/test_search_service.py).
- Layer-violation / convention greps (typer-in-services, httpx-in-commands, formatter-in-clients, magic numbers, raw error-code strings, bare except, `print()`, token leaks, new bare `tuple[...]` returns) against the full diff → all empty ✓.
- `src/keboola_agent_cli/server/routers/search.py` (PR branch) read in full → unchanged by this PR; confirmed it calls `registry.search.search(...)` directly (the same `SearchService` instance the fix lives in), so `kbagent serve`'s REST `/search` endpoint inherits the case-insensitivity fix automatically with zero code change needed — focus point 2 resolved, no gap.
- `src/keboola_agent_cli/commands/search.py:131-134` (PR branch) read → `--regex` + `--search-type config-based` still raises `ErrorCode.INVALID_ARGUMENT` / exit 2, unchanged by this PR ✓ (focus point verification).
- `src/keboola_agent_cli/commands/config.py:494-498` (PR branch) read → `config search --ignore-case`/`-i` still defaults to `False` (case-sensitive by default), confirming the PR description's claim about the sibling surface ✓.
- Isolated `git worktree add .../pr580-wt origin/claude/issue-569-config-search-case-insensitive` (did NOT touch the shared review worktree or its HEAD) → `uv sync --extra server` → `make check` → **exit 0**, `5449 passed, 12 skipped, 147 deselected` (matches the PR description's claimed numbers exactly) ✓. Full `make check` target confirmed to run lint + format-check + typecheck + skill-check + version-check + command-sync-check + changelog-check + check-error-codes + check-sentinel-guards + loc-check + tests.
- `pytest tests/test_search_service.py -k config_based -v` → 5 passed ✓.
- **Reproduced the pre-fix failure directly**: temporarily removed the `ignore_case=True` line from `search_service.py` in the isolated worktree and re-ran the two new tests → both FAIL (`AssertionError: missed match for 'DCFAmount'`), then restored the file (confirmed `git status` clean after) → proves the regression test is real and would have caught the bug, matching the PR description's "verified this test fails on the pre-fix code" claim ✓.
- Byte-budget check (focus point 3): `git show origin/main:plugins/kbagent/agents/keboola-expert.md | wc -c` → 61990 (current main, NOT what the PR was authored against). `git merge-tree --write-tree origin/main origin/claude/issue-569-config-search-case-insensitive` → clean merge, no conflicts → `git show <result>:plugins/kbagent/agents/keboola-expert.md | wc -c` → **62119 bytes**, over the 62000 hard budget. See `[B-1]`.
- `plugins/kbagent/skills/kbagent/references/gotchas.md` diff hunk read → confirmed the new sentence uses "(issue #569)" not a version tag. See `[B-2]`.
- `src/keboola_agent_cli/permissions.py` → `"search": "read"` entry already existed pre-PR (no new command added, so no `OPERATION_REGISTRY` gap) ✓.
- `gh pr view 580 --json mergeable,mergeStateStatus` → `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (review-gate, not a merge conflict) — confirms the byte-budget overrun in `[B-1]` would NOT be surfaced by GitHub's own conflict UI.

(none)

PR #582, opened 2026-08-12
chore(release): renumber the unreleased 0.83.0 to 0.82.1

padak:
v0.82.0 shipped while #581 was in review, and it carried 0.81.0 with it via the
catch-up mechanism. That leaves exactly one unreleased change: the #581
documentation note recording that no mcp_tool migration command will be built.

A minor bump for a docs-only note is inflation -- nothing in it changes
behaviour, adds a command or moves an interface. 0.82.1 says what actually
happened.

Renumbered in pyproject.toml, the changelog key, and the `(0.83.0+)` since-tag
in keboola-expert.md; plugin.json / marketplace.json / uv.lock follow via
`make version-sync`. No content change.

---

<a href="https://app.devin.ai/review/keboola/cli/pull/582" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
   "plugins": [
     {
       "name": "kbagent",
-      "version": "0.83.0",
+      "version": "0.82.1",
       "source": "./plugins/kbagent",
       "description": "AI-friendly interface to Keboola Connection projects — explore configs, jobs, lineage, call MCP tools, manage dev branches, and debug SQL in workspaces",
       "category": "development"
--- a/plugins/kbagent/.claude-plugin/plugin.json
+++ b/plugins/kbagent/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kbagent",
-  "version": "0.83.0",
+  "version": "0.82.1",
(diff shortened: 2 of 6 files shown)


padak:
Closing as obsolete -- the premise no longer holds.

This PR assumed 0.83.0 was the only unreleased change, and that a minor bump for
a docs-only note was inflation. Since then #565 (`auth login-password`) merged and
took main to 0.84.0, so there are now two unreleased changelog keys, 0.83.0 and
0.84.0, and the version files this PR rewrites are already at 0.84.0 -- hence the
conflicts.

The inflation concern is also moot: 0.83.0 will never be its own release tag.
`gh release list` shows the last tag is v0.82.0; 0.81.0 was carried by it via the
catch-up mechanism, and 0.83.0 will be carried by v0.84.0 the same way. Renumbering
it to 0.82.1 would only place a patch of 0.82 *below* an already-written 0.84.0 entry
in the changelog, which reads worse than leaving it alone.

If the empty-looking 0.83.0 key ever becomes a nuisance, the cheaper fix is to fold
its note into the 0.84.0 entry and retag the since-marker in
`plugins/kbagent/agents/keboola-expert.md` from `(0.83.0+)` to `(0.84.0+)` -- the
version users actually receive it in. No rebase of five version-synced files needed.

PR #601, opened 2026-08-18
fix(http): never replay a credential-minting call on an ambiguous failure (#599)

### LINKED ISSUE #599: token create: persistent upstream 500 on POST /v2/storage/tokens surfaces as a blind retry + generic error, no actionable guidance
## Summary

`kbagent token create` against a project on the `europe-west3.gcp` (EU GCP) stack fails every time with a raw `500` from the Keboola Storage API. Investigation shows kbagent itself is behaving correctly (correct request, retries per its documented policy, faithfully surfaces the upstream error) — this looks like an upstream Storage API defect, not a kbagent bug. Filing here anyway because (a) kbagent is the only way I have to observe/reproduce it, and (b) there's a concrete CLI-side hardening gap exposed by it (see "Proposal").

## Environment

- `kbagent` v0.84.2 (`keboola-cli` PyPI package)
- Stack: `connection.europe-west3.gcp.keboola.com`

## Repro

```
kbagent -v token create -p <project> -d "<description>" --expires-in 60
```

Response: `500 {"error": "Application error."}` from `POST /v2/storage/tokens`, retried 3x by kbagent's existing backoff policy (500 is in `RETRYABLE_STATUS_CODES`), all 3 attempts fail identically.

Ruled out as the cause, so this doesn't look like a request-shape bug on the client side:
- Reproduces on 2 different projects on the same stack, with 2 different acting tokens.
- Reproduces with no scope, and with `--bucket-read` on a real bucket.
- Reproduces with `--expires-in 60`, `--expires-in 3600`, and omitted entirely (never expires).
- Acting token confirmed to have `canManageTokens` — rules out the missing-permission theory (which per `http_base.py`'s own `_raise_api_error` should surface as a clean `403` → `ErrorCode.ACCESS_DENIED`, not a `500`).
- No incident reported for the `europe-west3` stack on status.keboola.com / keboolastatus.com at the time of testing.

## Proposal

Not asking for a fix to the upstream API from this repo — that's a Keboola Connection backend issue I'll raise with Keboola support separately. But two things are actionable in kbagent itself:

1. **Non-idempotent retry risk**: `POST /v2/storage/tokens` is a create/mint call, not idempotent, yet it's retried on 500/502/503/504 like any other request. If the API ever creates the token server-side and then fails to respond cleanly (unlike this case, where all 3 attempts failed identically with no sign of partial success), a blind retry could silently mint a duplicate live credential the caller never sees. Worth excluding mutating/non-idempotent endpoints from the generic retry policy, or gating the retry on an idempotency key.
2. **Generic 500 message gives no next step**: today this surfaces as `API error 500 ... Application error.` with no hint that a *persistent* 500 (survives all retries, stable across projects/tokens/payloads) is very likely an upstream incident rather than something the caller can fix. A short, specific hint in the `ErrorCode.API_ERROR` 500 path (e.g. "if this persists across retries/projects, it's likely an upstream Keboola incident — contact Keboola support") would save the next person the same investigation.

## Open questions

- Is there a better internal channel (Keboola support / a status page not covered by status.keboola.com or keboolastatus.com) that kbagent maintainers already use to escalate confirmed upstream 500s like this one? If so, pointing to it from the error message would be more useful than a generic hint.

padak:
## What

Two CLI-side hardening asks from #599, whose reporter correctly concluded the underlying `500` on `europe-west3.gcp` is upstream and not a kbagent bug.

**1. Non-idempotent calls are no longer replayed on an ambiguous failure.**

`POST /v2/storage/tokens` mints a credential, but the shared retry loop treated it like any other request and replayed it up to 3 times on `500/502/503/504` — and on read timeouts. If the server ever mints the token and then fails to answer, the replay leaves a second live credential behind that the caller never sees a value for, and therefore can never revoke.

`BaseHttpClient._do_request` now takes `idempotent` (default `True`, so every existing call site is byte-identical). Four call sites pass `False`:

- `create_scoped_token` and `create_short_lived_token` (`POST /v2/storage/tokens`)
- `create_git_credential` (data-app managed repo — mints a one-time `http_token` secret)

The distinction is not "is it a POST" but "could the request have reached the handler":

| Failure | Non-idempotent | Why |
|---|---|---|
| `429` | still retried | rejected by the rate limiter before the handler runs |
| connect error / connect timeout | still retried | connection never established |
| `5xx` | not retried | server may have completed the write and failed to answer |
| read/write timeout | not retried | request may have been delivered and applied |

So the change costs no resilience anywhere replay is provably safe.

`token refresh` is **deliberately left retryable**: a replayed rotation overwrites the value the lost response carried rather than leaving a second credential behind, so the caller still ends up holding the only token that authenticates. Documented in the method docstring so it does not read as an oversight.

**2. A 5xx now says it is upstream, and carries the support id.**

`API error 500 ... Application error.` gave the caller no next step — which is exactly why the reporter spent an investigation across two projects and two tokens before concluding it was not theirs to fix. Server errors now append the status-page/support hint, plus `(exceptionId: ...)` whenever the response body carries one (the first thing Keboola support asks for; previously parsed and discarded).

4xx messages are untouched — a rejected request is the caller's to fix and must not blame upstream. A non-idempotent 5xx additionally reports `retryable: false` in the `--json` error envelope, so an automated caller does not replay a mint the server may already have honoured. Nothing in this repo branches on `retryable`; it is advice carried to the caller.

## Not in this PR

- The upstream `500` itself — a Keboola Connection backend issue, as the reporter says.
- `kbagent token list`, asked for in a follow-up comment on the same issue. Separate scope (new command + all doc surfaces), tracked separately.

## How it was tested

`make check` green: lint, format, typecheck, skill, version, command-sync, changelog, error-codes, sentinel-guards, file-size, **5758 tests passed**.

New tests:
- `tests/test_http_base.py::TestNonIdempotentRetryPolicy` — 7 cases pinning each row of the table above, plus a case proving the default (idempotent) path still makes its 3 attempts.
- `tests/test_http_base.py::TestUpstreamErrorGuidance` — hint on 5xx, `exceptionId` surfaced when present, no hint on 4xx.
- `tests/test_client_device_enrollment.py::TestCreateScopedTokenIsNotReplayed` — the reported scenario at the client layer: a persistent 500 on `create_scoped_token` costs exactly one attempt.

No live-API testing: reproducing this needs the upstream outage.

## Note for the maintainer

The changelog entry sits under `0.84.2`, the current top key, since `main` is exactly `v0.84.2` and there is no unreleased key. Renumber at release time. No command was added, renamed or removed, so no command doc surface needed syncing; `docs/sdk.md` gained one line because `create_scoped_token` is public SDK API and its retry behaviour changed.

Fixes #599

### DIFF
--- a/docs/sdk.md
+++ b/docs/sdk.md
@@ -29,7 +29,7 @@ with Client(url=os.environ["KBC_URL"], token=os.environ["KBC_TOKEN"]) as kbc:
 
 **It is:**
 
-- **Stateless.** A `Client` holds the stack URL, the token, one pooled HTTP client (with the shared retry/backoff), and an optional idempotency store. Nothing else.
+- **Stateless.** A `Client` holds the stack URL, the token, one pooled HTTP client (with the shared retry/backoff), and an optional idempotency store. Nothing else. The retry loop deliberately skips the credential-minting calls (`create_scoped_token` and friends) on an ambiguous failure -- a replayed 5xx or read timeout could mint a second live token you never see a value for (#599). A 429, connect error or connect timeout never reached the handler, so those stay retryable.
 - **Config-dir-free.** No `~/.config/keboola-agent-cli/`, no `kbagent project add`, no `config.json`. Auth is the token you pass in (12-factor: you read `KBC_TOKEN` yourself).
 - **In-process.** No CLI subprocess, no `kbagent serve` daemon. Just function calls.
 - **Typed.** `py.typed` ships in the wheel; the high-traffic operations return pydantic models (`JobResult`, `QueryResult`, ...). Your `mypy`/`ty`/IDE sees the shapes.
--- a/src/keboola_agent_cli/auth/auth_client.py
+++ b/src/keboola_agent_cli/auth/auth_client.py
@@ -770,7 +770,13 @@ class AuthClient(BaseHttpClient):
     # Shared error mapping
     # ------------------------------------------------------------------
 
-    def _raise_api_error(self, response: httpx.Response, base_url: str | None = None) -> None:
+    def _raise_api_error(
+        self,
+        response: httpx.Response,
+        base_url: str | None = None,
+        *,
+        idempotent: bool = True,
+    ) -> None:
         """Escalate a 404 before falling back to the shared error mapping.
 
         `BaseHttpClient._do_request` calls this method for every
@@ -778,6 +784,10 @@ class AuthClient(BaseHttpClient):
         (rather than adding a check in each method) covers all of them at
         once. `poll_device_token` bypasses `_do_request` entirely and calls
         `_map_auth_error` directly for the same 404 case.
+
+        `idempotent` is accepted to keep the override signature-compatible
+        with the base class; no auth call passes it, and `_map_auth_error`
+        raises before the base mapping (which reads it) is ever reached.
         """
         self._map_auth_error(response)
 
--- a/src/keboola_agent_cli/changelog.py
+++ b/src/keboola_agent_cli/changelog.py
@@ -25,6 +25,32 @@ from .constants import CHANGELOG_HEADLINE_MAX_CHARS
 # Ordered newest-first.  Each value is a list of brief one-line descriptions.
 CHANGELOG: dict[str, list[str]] = {
     "0.84.2": [
+        "Fix: credential-minting API calls are no longer replayed on an ambiguous failure "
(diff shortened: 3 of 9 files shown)


padak:
Closing as superseded by #616 (merged, shipped in 0.88.0), which addressed the same root cause with a broader scope: retries are now restricted to RFC 9110 idempotent methods across all POST call sites (not just token minting), the error message surfaces `exceptionId` and whether the call was retried, and `kbagent token list` was added.

PR #602, opened 2026-08-18
feat(notification): audit Flow Notification subscriptions fleet-wide (#600)

### LINKED ISSUE #600: kbagent notification: fleet-wide audit of Flow Notification subscriptions (Notification Service API)
## Problem

Auditing a Keboola org's Flows for "who gets notified when something breaks" is only half-possible today.

`kbagent flow list` / `config detail --component-id keboola.flow` fully expose:
- flow descriptions
- in-flow **Notification tasks** (a task node of `type: "notification"` with `recipients: [{channel, address}]`, added inside a phase)

They do **not** expose the Flow Builder's own **Notifications tab** — the per-flow Success / Error / Processing-delay / Warning cards with their own recipient lists, reachable via the bell icon in the UI. That's a genuinely different mechanism: it's backed by a separate **Notification service**, not stored inside the flow's `configuration` JSON.

I ran a fleet-wide audit for a customer (20 projects, 276 `keboola.flow` + `keboola.orchestrator` configs) checking every email address referenced anywhere in flow descriptions, in-flow notification tasks, and email-sending component configs (`kds-team.app-email-smtp-sender`, `kds-team.app-mailgun-v2`) — and found real issues (placeholder addresses, typo'd recipients, missing owners). But the Notifications-tab recipients — arguably the most important ones, since they're what actually pages someone when a production flow fails — were completely unreachable from `kbagent`, the MCP server, or any CLI. The only fallback was "open each flow in the UI by hand."

## What exists today (confirmed via public discovery, no auth needed)

- `GET https://connection.<region>.keboola.com/v2/storage` lists platform services, including:
  ```json
  {"id": "notification", "url": "https://notification.eu-central-1.keboola.com"}
  ```
- Its swagger is public: `https://notification.eu-central-1.keboola.com/docs/swagger.yaml`. Relevant paths:
  - `GET /project-subscriptions` — list all subscriptions for the project resolved from the auth token. Optional `?event=` filter.
  - `POST /project-subscriptions` / `DELETE /project-subscriptions/{id}` — create/remove.
  - Auth: `security: [StorageApiTokenAuth, StorageApiBearerToken, ...]` — **a plain Storage API token**, the same kind every `KeboolaClient` in kbagent already holds per project. No elevated scope (e.g. `canManageTokens`) needed for the read path — I originally assumed it would and burned a while confirming otherwise.
  - Schema: `event` (`jobFailed`, `jobSucceeded`, `jobSucceededWithWarning`, `jobProcessingLong`, plus `phaseJob*` variants), `filters` (`field`/`value`/`operator`, e.g. matching `configurationId`/`component`), `recipient` (`channel: email|webhook` + `address`), optional `expiresAt`.

So the missing piece is purely a client + CLI surface in `kbagent` — the API itself is a normal sibling service exactly like Queue/Encryption/Sync-Actions, which kbagent already talks to.

## Proposed solution

Follow the existing `queue` / `encryption` / `sync-actions` pattern in `client/_core.py` + the existing `schedule` command as the structural template (it already solves the identical shape of problem — "audit a per-project sibling-service concept across every registered project, joined against `keboola.flow`/`keboola.orchestrator` config names").

1. **`client/_core.py`**: add
   ```python
   @property
   def _notification_base_url(self) -> str:
       return self._derive_service_url(self._stack_url, "notification")

   def _notification_request(self, method: str, path: str, **kwargs: Any) -> httpx.Response:
       client = self._get_or_create_sub_client("_notification_client", self._notification_base_url)
       return self._do_request(method, path, client=client, base_url=self._notification_base_url, **kwargs)
   ```
   plus the matching `self._notification_client: httpx.Client | None = None` init and `close()` wiring — copy-paste of the `_queue_*` / `_encrypt_*` trio.

2. **`client/notifications.py`** (new mixin, modeled on `client/queue.py`):
   ```python
   class _NotificationMixin(_CoreClient):
       def list_project_subscriptions(self, event: str | None = None) -> list[dict[str, Any]]:
           params = {"event": event} if event else {}
           return self._notification_request("GET", "/project-subscriptions", params=params).json()
   ```
   (Write methods — `create_project_subscription` / `delete_project_subscription` — are trivial to add too, but out of scope for this ask; read-only is enough to close the audit gap.)

3. **`services/notification_service.py`** (new, modeled on `services/schedule_service.py::ScheduleService`):
   - `list_subscriptions(aliases: list[str] | None, event: str | None, component_id: str | None, config_id: str | None) -> dict` — fans out `self._run_parallel` over `self.resolve_projects(aliases)`, calls `list_project_subscriptions(event)` per project.
   - Joins each subscription's `filters` (`configurationId` / `component`) against `list_components_with_configs()` from the same project — exactly the `_partition_components` trick `ScheduleService` uses to resolve `parent_name` — so the output row can show *which flow* a subscription belongs to, not just a bare `configurationId`.
   - Same error-accumulation contract as `ScheduleService`: per-project failures land in `errors`, never abort the whole fan-out.

4. **`commands/notification.py`** (new, modeled on `commands/schedule.py`):
   ```
   kbagent notification list [--project ALIAS...] [--event jobFailed|jobSucceeded|jobSucceededWithWarning|jobProcessingLong] [--component-id keboola.flow] [--config-id ID]
   ```
   Output columns: `project_alias`, `subscription_id`, `event`, `flow_name` (resolved), `component_id`/`config_id`, `channel`, `address`, `expires_at`. Register the command group next to `schedule_app` in the root Typer app and in `check_cli_permission` the same way (read-only, safe under `--deny-writes`).

5. **Docs**: mention `notification list` from `flow list`'s docstring as the companion command for "what does the Notifications tab say", the way `flow list` already flags the `legacy_orchestrator_count` caveat for orchestrator flows. Add a changelog entry.

## Open questions / things to verify while implementing

- Whether subscriptions are branch-aware (dev branch flows). The swagger doesn't show a branch filter — worth checking against a real dev-branch flow before assuming production-only scope, similar to the existing branch caveat on `schedule find --not-run-since` (Queue API isn't branch-aware either).
- Whether `filters` always contains a `configurationId` for flow-level subscriptions, or whether some events are project-wide with no filters at all (e.g. a catch-all "notify me on any job failure" subscription) — the join logic needs to degrade gracefully to "project-wide, no specific flow" rather than erroring.

## Non-goals for this issue

- Write path (create/delete subscriptions) — natural follow-up, not needed to close the audit gap.
- Anything about the in-flow `type: "notification"` task (already fully supported today via `flow detail`/`config detail`) — this issue is only about the Notifications-tab / Notification-service side.

## Why it matters

This was the one blind spot left after an otherwise complete fleet-wide "are our flow notification emails valid" audit across 20 projects — every other notification surface (owner emails in descriptions, in-flow notification tasks, email-sending component recipients) was fully auditable via existing `kbagent` commands in a few minutes; this one required opening every flow by hand in the UI. Given `kbagent`'s stated goal of wrapping "everything ... into workflow-oriented commands where dev branches propagate automatically, multi-project operations run in parallel" — this is one of the last multi-project operations that still can't be done from the CLI.

padak:
## What

```
kbagent notification list [--project ALIAS ...] [--event NAME] [--component-id ID] [--config-id ID] [--branch ID]
```

Closes the last flow-notification blind spot described in #600. The Flow Builder **Notifications tab** — the bell icon's Success / Error / Processing-delay cards — is backed by the Notification Service, not by the flow's `configuration`, so `flow detail` and `config detail` never could show it. On the 20-project / 276-flow fleet in the issue, every other notification surface (owner emails in descriptions, in-flow `type: "notification"` tasks, email-sender component configs) was auditable in minutes; these recipients — the ones that actually page someone when production breaks — required opening each flow in the UI by hand.

Reads `GET /project-subscriptions` on the derived `notification.{stack}` host with a plain Storage token (no elevated scope), fans out over every registered project in parallel, and collects per-project failures in `errors` rather than aborting the run. Mirrored as `GET /notifications` on `kbagent serve`, per the 1:1 convention.

Follows the structure the issue proposed and the `billing credits` PR (#597) established: `client/notifications.py` mixin → `services/notification_service.py` → `commands/notification.py`, plus `permissions.py` (`notification.list = read`, safe under `--deny-writes`).

**Read-only by construction.** The service also exposes create/delete for subscriptions — changing who gets paged when production breaks. `_notification_get` hardcodes the verb like `_billing_get` does against a real-money top-up, so no future caller can reach the write path through the dispatcher at all.

## Contract details that differ from the issue's proposal

I fetched the service's public OpenAPI rather than coding to the shapes in the issue text, and three of them differ. Each one would have produced a command that silently returns nothing:

| Issue proposed | Actual (per the service's own schema) |
|---|---|
| `jobFailed`, `jobSucceeded`, … | kebab-case: `job-failed`, `job-succeeded`, `job-succeeded-with-warning`, `job-processing-long`, `phase-job-*` |
| filters on `configurationId` / `component` | dotted: `job.component.id`, `job.configuration.id`, `branch.id` |
| recipient `{channel, address}` | discriminated union — `email` carries `address`, `webhook` carries `url` |

`event` is typed as a free-form string, not an enum, so no client-side allow-list is imposed: an unknown value gets the API's own 400, and a newly shipped event type works the day Keboola ships it.

## The issue's two open questions, answered

**Are subscriptions branch-aware?** Yes, but as a *filter*, not a scope: a dev-branch subscription carries a `branch.id` filter, a production one carries none, and the list endpoint returns both together. So `--branch` filters client-side and — unlike every other branch-aware command — is **never** inferred from the project's active branch. Inheriting it would silently hide the production recipients an audit exists to check; there is a test pinning exactly that. A branch id is meaningful in one project only, hence `--branch` requires exactly one `--project` (exit 2 otherwise).

**Do project-wide subscriptions exist?** Yes — a subscription with no config filter fires for every job in the project. It degrades to `scope: "project-wide"` with an empty `config_id` rather than erroring, and is often the most important row in an audit. Likewise a subscription pointing at a deleted configuration keeps its id with an empty `config_name` — a finding, not an error.

## Name resolution cost

`config_name` is joined with one `list_component_configs` call per distinct **(branch, component)** pair actually referenced — in practice one, and none at all when every subscription is project-wide. `ScheduleService`'s `list_components_with_configs` was deliberately not reused: it downloads every configuration *body* in the project (megabytes on a 276-flow project) to recover a handful of names. Grouping by branch matters because a dev-branch config is invisible from production, and looking it up in the wrong branch would report a live flow as deleted.

## How it was tested

`make check` green: lint, format, typecheck, skill, version, command-sync, changelog, error-codes, sentinel-guards, file-size, **5786 tests passed**.

- `tests/test_notification_client.py` (7) — derived host, `X-StorageApi-Token` on the wire, `?event=` passthrough, clean URL without it, wrapped/garbage payload tolerance, and a test asserting the dispatcher exposes no `method` parameter.
- `tests/test_notification_service.py` (16) — every row shape above, both recipient shapes, threshold filters surviving verbatim, per-branch name resolution, client-side filters, fan-out with a failing project, client cleanup, unknown alias.
- `tests/test_notification_cli.py` (13) — JSON envelope, human table, project-wide labelling, dangling-config fallback, argument forwarding, `--branch` guard, per-project warnings at exit 0, `--deny-writes`.
- `tests/test_server_router_calls.py` (3) — query params → service kwargs, envelope returned verbatim.
- `tests/test_e2e.py::TestE2ENotificationList` (3) — envelope shape and invariants against a live project; row assertions run only if the project has subscriptions, so it starts covering the populated path without a rewrite.

No live-API verification of a *populated* list: that needs a project with Notifications-tab recipients, which I do not have. The wire contract comes from the service's published OpenAPI, and every shape it documents is covered by a test.

## Doc surfaces (convention #17)

Updated: `context.py` AGENT_CONTEXT, `CLAUDE.md`, `SKILL.md` (via `make skill-gen`), `commands-reference.md`, `gotchas.md`, new `notification-workflow.md`, `changelog.py`.

**Not updated — needs a maintainer decision: `plugins/kbagent/agents/keboola-expert.md`.** It is at 61,991 bytes against the 62,000-byte budget enforced by `tests/test_agent_prompt.py` — **9 bytes free**, so a new command-group row does not fit without trimming content I would only be guessing is stale. CONTRIBUTING's guidance is to trim rather than raise the cap, or split into per-domain specialists. The practical gap is small: the skill runs `kbagent context` as its first step, and `AGENT_CONTEXT` does carry the new group in full.

## Note for the maintainer

Changelog entries sit under `0.84.2`, the current top key, since `main` is exactly `v0.84.2` and there is no unreleased key. Renumber at release time. The version is not bumped.

Fixes #600


padak:
Closing as superseded by #615 (+ follow-up #618), which shipped `kbagent notification list` / `notification detail` in 0.86.0 with a narrower, verified-live `--event` client-side filtering semantics. The feature this draft implements is already in production and documented.

PR #626, opened 2026-08-21
feat(data-app): AI-3753 grant Storage access by default on `data-app create`

MiroCillik:
Closes [AJDA-3183](https://linear.app/keboola/issue/AJDA-3183/runtimeworkspaceenabledtrue-is-undocumented-and-unset-by-kbagent-apps).

`kbagent data-app create` gains `--workspace / --no-workspace`, **defaulting ON**. It writes `runtime.workspace.enabled: true` — the single switch that makes the platform provision a data app's ephemeral workspace and inject `WORKSPACE_ID`, `QUERY_SERVICE_URL` and `KBC_WORKSPACE_MANIFEST_PATH`.

## Validation of the report

Every claim in the issue checks out:

| Claim | Verified |
| -- | -- |
| kbagent never sets the flag | `data_app_service.py` wrote `"runtime": {"backend": {"size": size}}` and nothing else; `grep` for `workspace.enabled` across the repo returned 0 hits |
| MCP does set it | `keboola-mcp-server` `tools/data_apps.py:1242`, gated on `has_feature('data-apps-storage-workspace')` |
| MCP has a legacy fallback | `:1113-1117` injects `WORKSPACE_ID` via `parameters.dataApp.secrets` from a find-or-create RO workspace |
| The config shape | live `Direct Access Beers JS` config in project 40: `runtime: {backend: {size, type}, workspace: {enabled: true}}` |

## Two decisions worth reviewing

**1. Default ON.** Omitting the flag fails *silently* — the app deploys, reports `state=running`, passes its health probe, and reads nothing. The opposite mistake (an unused workspace on an app that never reads Storage) is the cheap one, so `--no-workspace` is the opt-out rather than `--workspace` the opt-in.

`--no-workspace` **omits the key** rather than writing `enabled: false`, so that request body is byte-identical to 0.86.0.

**2. No feature gate, and no secrets fallback.** The `data-apps-storage-workspace` project feature is being enabled everywhere and then removed, so the config option is the sole control. This deliberately does **not** copy MCP's `parameters.dataApp.secrets.WORKSPACE_ID` fallback — replicating it would mean creating a long-lived workspace that the app then depends on forever, and which `kbagent workspace gc` could delete out from under it.

## Verified live

Created a real app on project 40 and read the config back:

```json
"runtime": { "backend": { "size": "tiny" }, "workspace": { "enabled": true } }
```

...matching the known-working app byte for byte. (App deleted afterwards.) I verified the **config shape**, not a live data read — the issue already establishes causality: the reporter set this flag on their broken app and it immediately worked (`[duck] cached 2412 beers in 7253ms`).

## Also in this PR

* `data_app_service.py` **was over its grandfathered size budget**, so `make loc-check` blocked the addition. Split the pure body-builders and redaction helpers into `services/_data_app_bodies.py` (re-exported, so existing import sites are untouched); baseline ratcheted 1726 → 1648 code lines.
* **Fixed a latent trap in the e2e data-app** `cleanup` **fixture** — it touched `self._created_app_ids` on teardown even when `setup` had skipped before assigning it, so any test gated only on credentials errored. Existing tests dodged this by carrying a skip *marker*, which fires before fixtures.
* `keboola-expert.md` **was at 61999/62000 bytes**, so any addition broke its prompt budget. Trimmed back under by collapsing duplication with the §2 matrix row and honouring §3's stated "one-line triggers only" contract, rather than raising the ceiling (which the test comment explicitly discourages).

## Out of scope for this repo

* Issue item 3 (a `duckdb-caching.md` prerequisite line) lives in `keboola/ai-kit`, not here.
* Issue item 5 (have the platform fail a deploy loudly when an app requests Storage env vars it was never granted) is a platform change. Still worth doing — it is the only fix that helps apps created by paths other than kbagent.

## Test plan

* `make check` — green (5699 passed)
* Full suite with `--extra server` — green; confirmed the 2 new REST tests and 88 server-router tests actually executed rather than skipping
* New coverage: 3 service tests (default on, opt-out omits the block, dry-run preview does not lie), 3 CLI tests (flag pass-through both ways, human output on/off), 2 REST tests, 2 e2e tests (live config assertion + dry-run opt-out)

## Release Notes

* **Justification**
  * A data app that could not read a single row reported itself healthy at every level a user can see — `state=running`, `desiredState=running`, health probe green. The one flag that grants Storage access was undocumented and unset by kbagent, and there was no CLI option to set it. Users found it by dumping the config of an unrelated working app and diffing. This makes the safe shape the default shape.
  * `data-app create` now grants Storage access unless you explicitly opt out with `--no-workspace`.
* **Plans for Customer Communication**
  * Changelog entry included (0.87.0). Worth flagging to whoever owns the `dataapp-developer` skill: its DuckDB-cached read-only dashboard pattern is exactly the one that needs this flag, and apps created with kbagent ≤ 0.86.0 need the retrofit (`config update --merge --set 'runtime.workspace.enabled=true'` then `data-app deploy`).
  * Not a BC break: existing configs are untouched, and `--no-workspace` reproduces the old body exactly.
* **Impact Analysis**
  * Newly created apps get an ephemeral workspace provisioned at start even when they never read Storage. That is the intended trade (cheap mistake vs. silent dead data path) and `--no-workspace` avoids it.
  * Not behind a feature flag. Not gated on `data-apps-storage-workspace` — per platform direction, that flag is being enabled everywhere and then removed.
  * No single-tenant-specific behavior. Any project where the platform ignores the flag lands on the pre-existing behavior, not worse.
* **Deployment Plan**
  * Continuous — normal `kbagent` release. Not critical.
* **Rollback Plan**
  * Two-way. Revert the commit, or pass `--no-workspace`. No migration, no data change, no config rewritten in place.
* **Post-Release Support Plan**
  * No new support burden. If a user reports "my data app shows no data", the gotchas entry and `data-app-workflow.md` now name the cause and the retrofit; `data-app logs` shows `Missing env vars: WORKSPACE_ID`.
  * Support does not need a heads-up beyond the changelog.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


src/keboola_agent_cli/services/_data_app_bodies.py
```diff
@@ -0,0 +1,217 @@
...
+    "KBC::ProjectSecure::",
+    "KBC::ProjectSecureGKMS::",
+    "KBC::ProjectSecureKV::",
+)
+
+# Defence-in-depth caps for free-form user input. The platform may accept
+# longer values, but kbagent refuses anything beyond these bounds at the
+# service boundary so an external caller using the service directly cannot
+# exfiltrate giant payloads or smuggle control characters into audit logs.
+
```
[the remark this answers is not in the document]
MiroCillik: Good catch, and it was worse than described — the split swapped *both* halves of a comment pair.

The mechanical extraction cut on a line boundary that fell between a comment block and the constant it documented, in both directions:

- the `MAX_*` "defence-in-depth caps" comment came here, where no `MAX_*` constant exists (what you spotted);
- the 16-line block documenting `ENCRYPTED_PASSWORD_PREFIXES` (the per-cloud cipher registry, and why the 4th entry was dropped in 0.86.0) stayed behind in `data_app_service.py`, where it ended up sitting directly above `MAX_NAME_LENGTH = 255` and describing nothing.

Both are now attached to their own constants. Nothing was dropped, just misfiled.

I also verified the split is content-complete rather than eyeballing it: diffed every non-blank, non-import line of `main`'s `data_app_service.py` against the union of the two files as a multiset. The only lines absent from the union are the two `"runtime": {"backend": {"size": size}}` literals and one docstring line — exactly the three the feature intentionally replaced.

CLAUDE.md
```diff
@@ -568,7 +568,19 @@ kbagent workspace from-transformation --project ALIAS --component-id ID --config
...
+kbagent data-app create --project ALIAS --name NAME --slug SLUG (--git-repo URL | --use-managed-git-repo) [--description STR | --description-file PATH] [--git-branch main] [--git-public/--no-git-public] [--git-username USER] [--git-pat-env VAR | --git-pat-file PATH | --git-pat-encrypted KBC::Project...] [--auth password|public] [--size tiny|small|medium|large] [--auto-suspend SECONDS] [--type python-js|python|streamlit|r|...] [--workspace/--no-workspace] [--branch ID] [--no-deploy] [--wait] [--timeout SECONDS] [--keep-on-failure] [--dry-run]
+# --workspace / --no-workspace (0.87.0+): DEFAULT ON. Writes runtime.workspace.enabled=true --
+#   the ONLY switch that makes the platform provision the ephemeral workspace and inject WORKSPACE_ID,
+#   QUERY_SERVICE_URL and KBC_WORKSPACE_MANIFEST_PATH. Every app that reads Storage needs it. Before
+#   0.87.0 kbagent never wrote it and offered no way to: such an app deploys, reports state=running and
```
Matovidlo, line 575: Do we need to reference this version ? Just out of cursioty, the user should be always at start of work prompted to update the CLI and then it should work in general. Maybe its benefitial to know what is the plugin version vs CLI version (some matrix)

MiroCillik: Two things worth separating here.

**The version tag is load-bearing for the agent, not decoration.** `keboola-expert.md` Rule 6 is a hard VERSION GATE: the agent runs `kbagent context` first, reads the installed version, and treats each inline `(0.X.Y+)` tag as the version floor for the task — refusing rather than improvising when it is below. This section already carries ~20 such tags for that reason.

The startup auto-update prompt does not close the gap: it is skippable, it never lands on a pre-release, and since 0.79.0 a standalone binary (choco/winget/brew/apt) **refuses** `kbagent update` outright and reports its own channel's command instead. So an agent really can be on 0.86.0 while reading docs written for 0.87.0.

It matters more than usual for this particular flag. On <= 0.86.0 it does not exist, and an agent that assumes it does will report a clean `data-app create` for an app with a dead data path — the exact silent failure this PR exists to fix. Without the tag the agent has no way to know it owes the user a `config update --merge --set 'runtime.workspace.enabled=true'` + redeploy.

**On a plugin-vs-CLI matrix:** there is nothing to tabulate, because they are pinned 1:1 by construction. `plugins/kbagent/.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` are generated from `pyproject.toml` by `make version-sync`, and `make version-check` (in `make check`, so CI) fails the build if any of them drift. A matrix would only ever have a diagonal.

That said, your underlying point stands — that invariant is currently implicit. If it is not obvious that plugin version == CLI version always, that is worth one line in CLAUDE.md. Happy to add it.

Matovidlo (approved):

PR #628, opened 2026-08-21
fix(storage): write column descriptions where the MCP server reads them (#624)

### LINKED ISSUE #624: `storage describe-column` writes column descriptions where the MCP server never reads them
# `storage describe-column` writes column descriptions where the MCP server never reads them

**kbagent:** v0.86.0
**keboola-mcp-server:** 1.74.6
**Impact:** column descriptions written by kbagent are invisible to every AI client using the Keboola MCP server, while `kbagent storage table-detail` reports them as present.

## Summary

There are two incompatible conventions for storing column descriptions:

| | Key | Location |
|---|---|---|
| **kbagent writes** | `KBC.column.{name}.description` | flat entry in **table** metadata |
| **MCP server reads** | `KBC.description` | inside `columnMetadata[{name}]` |

They never meet. Descriptions written via `kbagent storage describe-column` (or the `columns:` section of `storage describe-batch`) are silently invisible to `get_table_detail` / `search` in the MCP server — which is precisely the audience those descriptions are written for.

The failure is silent in both directions: `kbagent storage table-detail` reads back its own convention, so the descriptions look correctly applied.

## Root cause

`client/storage_tables.py:126-129` documents the assumption behind the current behaviour:

> Column-level descriptions use the namespaced key convention `KBC.column.{colname}.description` stored at table-metadata level (Keboola Storage API does not expose a user-writable column-metadata endpoint; `columnMetadata` is populated exclusively by components).

**That assumption is incorrect.** The same endpoint kbagent already uses, `POST /v2/storage/tables/{id}/metadata`, accepts a `columnsMetadata` payload with `provider: user`. The MCP server does exactly this in `clients/storage.py:944-953`.

The mirrored read path has the same problem — `services/storage_service.py:664-668` populates `col_descriptions` only from flat `KBC.column.*` keys, ignoring `KBC.description` inside `columnMetadata`, even though `columnMetadata` is already loaded on line 654.

## Reproduction

Verified live on project 9432 (`out.c-out-reporting.customer`).

Writing through the `columnsMetadata` payload:

```python
payload = {
    "provider": "user",
    "columnsMetadata": {
        "customer_id": [{"key": "KBC.description", "value": "Customer ID from the ERP system."}]
    },
}
client._request("POST", "/v2/storage/tables/out.c-out-reporting.customer/metadata", json=payload)
```

`GET /v2/storage/tables/{id}` afterwards:

```
customer_id  [... , ('KBC.description', 'Customer ID from the ERP system.', 'user')]
```

The write succeeds with `provider: user`, contradicting the docstring.

Reading the same table back through kbagent after 70 columns were described this way:

```
$ kbagent --json storage table-detail --project <alias> --table-id out.c-out-reporting.price_history_today
description:      "Today's price snapshot: ..."   # table description: read back fine
column_metadata:  0                               # <- the 70 column descriptions are not seen
column_details with a description: 0 of 70
```

## Where the MCP server reads from

- `tools/storage/tools.py:793-799` — column description resolved from `columnMetadata[col]` via `MetadataField.DESCRIPTION`, with a fallback to `sourceTable.columnMetadata`
- `tools/storage/search.py:107-110` — same lookup for search matching

Neither path inspects `KBC.column.*` keys.

## Suggested fix

1. `set_table_metadata` (or a sibling method): send column descriptions as a `columnsMetadata` JSON payload with `provider: user` instead of flat `metadata[i][key]` entries.
2. `storage_service.py:664-668`: read `KBC.description` from `columnMetadata[col]` as the primary source, keeping the `KBC.column.*` lookup as a fallback for already-written data.
3. Correct the docstrings in `client/storage_tables.py:126-129` and `services/storage_service.py:2534-2538`, plus `skills/kbagent/references/storage-describe-workflow.md`, which currently states that descriptions written this way will be visible in MCP `get_tables`.

A migration for existing `KBC.column.*` entries would be useful — any project already documented this way is silently invisible to its AI clients.

## Workaround

Bypass the CLI and post `columnsMetadata` through `keboola_agent_cli.lib.Client(...).raw`:

```python
payload = {"provider": "user", "columnsMetadata": {col: [{"key": "KBC.description", "value": desc}] for col, desc in cols.items()}}
client._request("POST", f"/v2/storage/tables/{quote(table_id, safe='')}/metadata", json=payload)
```

Applied to 529 columns across 20 tables in project 9432; all became visible to the MCP server.

padak:
## What

`storage describe-column` / `describe-batch` now write column descriptions as
`KBC.description` inside the table's native `columnMetadata` store, instead of as
flat `KBC.column.{name}.description` entries on the **table's** metadata.

## Why

Two incompatible conventions never met:

| | Key | Location |
|---|---|---|
| kbagent wrote | `KBC.column.{name}.description` | flat entry in **table** metadata |
| MCP server reads | `KBC.description` | inside `columnMetadata[{name}]` |

The `client/storage_tables.py` docstring justified the flat convention with "Keboola
Storage API does not expose a user-writable column-metadata endpoint". That is not
correct. The same `POST /v2/storage/tables/{id}/metadata` kbagent already calls accepts
a `columnsMetadata` payload with `provider: user`; `keboola/mcp-server`'s
`StorageClient.table_metadata_update()` does exactly that, and its read paths
(`tools/storage/tools.py`, `tools/storage/search.py`) resolve column descriptions only
from `columnMetadata`.

So every column description kbagent ever wrote was invisible to every AI client using
the Keboola MCP server — precisely the audience those descriptions are written for.

The failure was **silent in both directions**: `storage table-detail` read back
kbagent's own convention, so the descriptions were reported as correctly applied.

## Changes

- **client** (`client/storage_tables.py`) — new `set_table_column_metadata()` posting a
  JSON `{provider, columnsMetadata}` payload. It sends JSON rather than form data
  because the nested per-column shape has no form-encoded equivalent; the table-level
  `set_table_metadata()` is untouched and still form-encoded. Entries carry the
  redundant `columnName` key, mirroring what the platform's own clients send.
- **service — write** (`services/storage_service.py::describe_columns`) — writes
  `KBC.description` per column through the new client method. `describe_batch` picks
  this up for free via its `columns:` section.
- **service — read** (`services/storage_service.py::get_table_detail`) — reads
  `KBC.description` from `columnMetadata` as the primary source, keeping the flat
  `KBC.column.*` keys as a fallback so descriptions written by an older kbagent stay
  visible here. When both carry a value, `columnMetadata` wins.
- **line budget** — `storage_service.py` is a grandfathered file that may not grow, so
  the repetitive five-branch per-column `elif` chain became a `_COLUMN_META_FIELDS`
  mapping. Adding a pass-through field is now a one-line data change, and the file
  stays at its 1733 ceiling.
- **docs** — corrected the incorrect claim in the client and service docstrings, the
  `describe-column` command help, `AGENT_CONTEXT`, `storage-describe-workflow.md`,
  `commands-reference.md`, and a rewritten `gotchas.md` entry tagged `(since v0.87.1)`.

## Migration

There is **no bulk migration command**. Re-running `describe-column` (or
`describe-batch`) on an affected table rewrites the description into the right place;
until then those tables stay invisible to MCP clients. The read fallback means nothing
disappears from kbagent's own output in the meantime.

A dedicated migration command was left out deliberately — it would be a new CLI surface
(with the full set of drift surfaces to update) for a one-shot operation that re-running
an existing command already performs. Happy to add one if you would rather have it.

## Version

Bumped to **0.87.1** with a changelog entry. This changes *where* data is written and
carries a migration note, so shipping it without a changelog entry would leave users
unaware they need to re-run `describe-column`. Renumber freely if it should land under a
different version.

## Testing

- `make check` green: lint, format, typecheck, skill-check, version-check,
  command-sync-check, changelog-check, error-codes, sentinel-guards, loc-check, and
  **5703 tests passed**.
- New service tests: `KBC.description` read from `columnMetadata`; `columnMetadata`
  beating the legacy flat key when both are present.
- Rewritten client tests: asserts the exact JSON `columnsMetadata` body and the
  `application/json` content type, plus the branch-scoped endpoint.
- Updated write-path tests assert the legacy `set_table_metadata` call no longer happens
  for columns; the pre-existing flat-key read tests now stand as legacy-fallback
  regression cover.
- Not verified against a live project — no live-project write access in this run. The
  endpoint contract is corroborated by the issue author's live reproduction on project
  9432 and by `keboola/mcp-server`'s implementation.

Fixes #624


padak:
Superseded by #631, which merged the same issue (#624) with a different — and better — write path.

**#628 writes** to `columnMetadata` via `POST /v2/storage/tables/{id}/metadata` with a `columnsMetadata` payload.
**#631 writes** through the native `PUT /v2/storage/branch/{branch}/tables/{id}/definition` endpoint — the one the web UI uses — and the backend *mirrors* the value into `columnMetadata` `KBC.description` on its own.

Two things #631 gets that this approach cannot:

1. `isDescriptionSystemManaged: false`, which is what stops the next component run's Output Mapping from overwriting a hand-authored description (DMD-1662). The metadata POST has no equivalent.
2. The mirroring only runs one way. The merged gotchas entry records it explicitly: *a `POST .../metadata` write never reaches the native field.* So this PR's descriptions would be visible to the MCP server but still absent from the native definition — and from the Snowflake `COMMENT` / BigQuery column description.

Merging this on top of #631 would therefore regress the fix rather than extend it. The branch is also `DIRTY` against main and carries a `pyproject`/`plugin.json`/`uv.lock` version bump that conflicts with the 0.88.0 renumber in #638.

Closing as superseded. The analysis here — in particular that `columnsMetadata` with `provider: user` is accepted, contradicting the old `client/storage_tables.py` docstring — is what pointed at the native endpoint in the first place.

Issue #655, opened 2026-08-23
serve: PermissionEngine firewall is not enforced on REST routes


padak:
A 0.89.0 serve audit found that the permissions firewall protects only the CLI process. `PermissionEngine` is built exclusively in the Typer callback (`cli.py:374`) and carried in `ctx.obj`; there are zero `check_cli_permission` / `permission_engine` references anywhere under `src/keboola_agent_cli/server/` (grep count 0). A persisted `permissions set --mode deny` policy, `--deny-writes`, and `--deny-destructive` therefore have no effect on `kbagent serve`, which exposes all 226 routes — including `DELETE /storage/buckets`, `POST /token/{p}/delete`, `DELETE /configs/...` — behind a single all-or-nothing bearer token (`server/auth.py:42`).

Aggravating detail: `permissions.py:320` contains a load-bearing comment asserting the opposite — *"The serve's own routes enforce their own permissions on top."* — which is false today and misleads the `http.*` risk classification directly beneath it.

Proposed direction:
1. Build the `PermissionEngine` in `create_app()` from the same config dir and enforce it in a FastAPI dependency (map route → `cli:read|write|destructive|admin` category, mirroring the CLI risk registry), OR
2. at minimum document loudly (docs/web-server.md + `permissions` command help) that serve is outside the firewall, fix the false comment, and add a read-only `GET /permissions/show` so clients can discover the policy.

Given the guard-rails-not-sandbox philosophy (#271), option 1 with a doc caveat seems right: the firewall guards against agent mistakes, and agents are exactly what `serve` hosts (scheduled `ai_agent` tasks calling back via `kbagent http`).

PR #658, opened 2026-08-23
feat(ui): 0.89.0 catch-up — jobs/config actions, trash & restore, tokens, palette

padak:
Frontend-only catch-up for the NERD web UI (`web/frontend`). Every item below is wired to a **serve route that already exists** — no Python was touched. Verified against `src/keboola_agent_cli/server/routers/` before wiring.

## What shipped

| # | Item | Routes wired |
|---|---|---|
| 1 | **Jobs actions** — per-row + in-drawer `re-run` and `terminate`. Terminate is offered only for `created` / `waiting` / `processing` (a terminal job has nothing to stop) and goes through `ConfirmModal` with an explicit `job_ids` list. Re-run starts a fresh job from the config **as it stands now** — the Queue API offers no replay of the historical `configData`, and the code says so. The SSE log stream is unchanged. | `POST /jobs/{p}/run`, `POST /jobs/{p}/terminate` |
| 2 | **Run from config** — the config detail view became a right-hand `Drawer` (it was an inline card, against the design contract) and gained a **Run job** action. Fire-and-return (`wait=false`), then a success line with an `open Jobs →` jump. | `POST /jobs/{p}/run` |
| 3 | **Config delete + Trash & restore (#643)** — **Delete** behind `ConfirmModal` (soft-delete; the modal says so), plus a **Trash** tab listing `deleted_at` + `version` with per-row **Restore**. Empty state: *“Trash is empty — deletes are reversible here.”* | `DELETE /configs/{p}/{component}/{id}`, `GET /configs/trash/{p}`, `POST /configs/{p}/{component}/{id}/restore` |
| 4 | **New Tokens page** (new `PageId` + sidebar entry under MANAGE) — fast list by default; a **derive last-used** toggle re-fetches with `with_last_used=true`. `never` / `unknown` / `error` render as *distinct* pills with tooltips and are never collapsed (they lead to opposite decisions), and no client-side re-sort happens because the server already returns dormant-first. Create / rotate / delete; **the secret is shown once** in a copy-to-clipboard `nerd-code` block with a “shown once” warning, held in React state only and dropped on close. The clipboard call degrades to a manual-copy hint on a non-secure origin. | `GET /token/{p}/list`, `POST /token/{p}/create\|delete\|refresh` |
| 5 | **Storage: table definition (#621)** — `TableDetail` gained `definition`; the Info tab renders Time partitioning / Range partitioning / Clustering / Partition filter required / Partitions (a **count** — `partitions[]` is unbounded). Renders nothing at all when there is no layout. | `GET /storage/table-detail/{p}/{id}` |
| 6 | **Storage: editable column descriptions (#624)** — the schema tab’s Description cell is click-to-edit (pencil on row hover, Enter saves, Esc cancels), optimistic with rollback + `ErrorBox` on failure. A non-empty `legacy_column_descriptions` surfaces a one-line `describe-migrate` hint. | `POST /storage/columns/{p}/{table_id}/describe` |
| 7 | **Dashboard: billing credits tile** — fifth `StatTile` scoped to the active project, showing remaining credits and derived minutes. `PAYG_NOT_AVAILABLE` renders as a muted `n/a` pill, not an error — it is the normal state on most stacks. | `GET /billing/credits` |
| 8 | **Flows: Notifications tab** — read-only. Fetches the project’s subscriptions **unfiltered** and splits client-side: passing `config_id` to the API drops the filter-less catch-alls *server-side*, and those fire for every job in the project, so a filtered fetch would silently under-report who gets paged. Project-wide subscriptions get their own group with a warning pill, and a note records that a `branch.id` value alone does not mean “dev branch” (production writes the default branch’s numeric id). | `GET /notifications` |
| 9 | **Command palette** — `Ctrl+K` / `Cmd+K` opens a centered overlay with subsequence fuzzy matching over all pages, all registered projects, and a couple of actions (toggle theme, open Swagger `/docs`). Arrows + enter, esc closes, cyan match highlighting, green selection bar. The page list is the sidebar’s now-exported `SECTIONS`, so a new page can never appear in one surface and not the other. Footer hint added to the `StatusBar`. | — (local) |
| 10 | **Cleanups** — `window.confirm` at `pages/Agents.tsx:566` replaced with `ConfirmModal` (portaled to `<body>`: the drawer’s `backdrop-blur` makes it a containing block, so a nested fixed modal would be clipped). Dead `useManageTokenPrompt` deleted from `state.tsx` — verified zero consumers first. | — |

## Verification

- `npm ci && npx tsc --noEmit && npm run build` — all clean.
- No eslint/prettier/biome config exists in `web/frontend`; `tsc -b && vite build` is the only gate, and it passes.
- `web/frontend/dist` and `src/keboola_agent_cli/_ui_dist` are gitignored and untracked — nothing built was committed.
- No version bump, no changelog entry (per the #648 process, the release PR handles those).
- `docs/web-server.md` “Web UI” section updated with the Tokens page, the command palette, and one line per new capability.

Part of the serve/UI audit follow-up (#655/#656/#657 context).

## Scope note

Item 10's cleanup also **deletes `useManageTokenPrompt` from `state.tsx`** — a `window.prompt`-based manage-token helper with zero remaining call sites (verified by grep before removal; `tsc` clean after). It is unrelated to the features above and is called out here so a future `git blame` on that deletion does not have to reconstruct the "why" from an 18-file diff. The `ManageTokenModal` component that superseded it is untouched.

## Review follow-ups (commits 2-4)

- `e65fcc6` — Devin: the optimistic column-description override was never cleared on success (masked later server values while the drawer stayed mounted); `ConfirmModal` now portals to `<body>` itself like `Drawer`, so confirms raised inside a drawer no longer depend on where they were declared.
- `c44a4f9` — live browser verification: the Queue API job row names the config id `config`, not `configId`. `types.ts` had the wrong name, so the Jobs table's Config column had always rendered empty **and** the new re-run button's `canRerun` gate was permanently false — the button never rendered at all.
- `ed66cc2` — `kbagent-pr-reviewer` B-1: re-run dropped `branch_id` and silently retargeted the default branch.

### DIFF
--- a/docs/web-server.md
+++ b/docs/web-server.md
@@ -120,18 +120,40 @@ and `next_run_at` so re-runs after restarts pick up where they left off.
 
 A NERD-themed React SPA that drives the API:
 
+- **Command palette** — `Ctrl+K` / `Cmd+K` anywhere: fuzzy jump to any
+  page, switch the active project, toggle the theme, open Swagger `/docs`.
+  Arrows + enter, esc closes.
 - **Dashboard** — greeting, big Kai chat input, stat tiles (projects /
-  agents / doctor / recent jobs), scheduled-agent activity, suggested
-  next steps, recent jobs panel.
+  agents / doctor / recent jobs / PAYG credits), scheduled-agent
+  activity, suggested next steps, recent jobs panel. The credits tile
+  reads `GET /billing/credits` for the active project and shows a muted
+  `n/a` on a non-PAYG project (`PAYG_NOT_AVAILABLE`).
 - **Projects, Branches, Doctor, Changelog** — manage local config and
   health.
+- **Tokens** — scoped Storage tokens for the selected project
+  (`/token/{p}/list`): create / rotate / revoke, with the secret revealed
+  ONCE in a copy-to-clipboard block, and an opt-in "derive last-used"
+  toggle (`with_last_used=true`) that sorts dormant tokens first and
+  renders `never` / `unknown` / `error` as distinct pills.
 - **Configs, Components (AI search), Storage (with per-column data
   preview), Jobs (cards layout + SSE log stream), Search** — browse a
   selected project.
+  - Configs: detail Drawer with **Run job** (`POST /jobs/{p}/run`) and
+    **Delete** (soft-delete via `DELETE /configs/…`), plus a **Trash**
+    tab (`GET /configs/trash/{p}`) with per-row Restore.
+  - Storage: the table drawer's Info tab renders the raw `definition`
+    (time / range partitioning, clustering, partition filter, partition
+    count) when the stack reports one, and the Schema tab's Description
+    cell is click-to-edit through `POST /storage/columns/{p}/{id}/describe`.
+  - Jobs: per-row **re-run** and **terminate** (`POST /jobs/{p}/run` /
+    `…/terminate`), the latter behind a confirm modal.
 - **SQL Workspaces** — info Drawer with credentials + actions sidebar;
   Open SQL Editor opens a Monaco editor with a clickable Storage
   Explorer tree on the left.
-- **Flows** — visual Mermaid builder of phase DAG + per-phase task list.
+- **Flows** — visual Mermaid builder of phase DAG + per-phase task list,
+  plus a read-only **Notifications** tab (`GET /notifications`) listing
+  who gets paged for the flow, with filter-less project-wide
+  subscriptions shown in their own warning-pilled group.
 - **Schedules** — cross-project cron list + find-by-window query.
 - **Data Apps** — list, start/stop, secrets, validate-repo.
 - **Lineage** — Sharing graph (live, cross-project) + Deep lineage (UI
--- a/web/frontend/src/App.tsx
+++ b/web/frontend/src/App.tsx
@@ -21,6 +21,7 @@ import { SearchPage } from "./pages/Search";
 import { SharingPage } from "./pages/Sharing";
 import { StoragePage } from "./pages/Storage";
 import { StreamsPage } from "./pages/Streams";
+import { TokensPage } from "./pages/Tokens";
 import { WorkspacesPage } from "./pages/Workspaces";
 import { UIStateProvider, useUIState } from "./state";
 import { ThemeProvider } from "./theme";
@@ -70,6 +71,8 @@ function Router() {
       return <OrgPage />;
     case "members":
       return <MembersPage />;
+    case "tokens":
+      return <TokensPage />;
     case "doctor":
       return <DoctorPage />;
     case "changelog":
--- a/web/frontend/src/components/CommandPalette.tsx
+++ b/web/frontend/src/components/CommandPalette.tsx
@@ -1,0 +1,326 @@
+/**
+ * Ctrl+K / Cmd+K command palette.
+ *
+ * One keystroke to reach anything the shell can already do: jump to a page,
+ * switch the active project, or fire a small action (theme, Swagger docs).
+ * Deliberately NOT a search over Keboola data -- that is the Search page's
+ * job, and mixing "navigate the app" with "query the project" makes both
+ * slower. Everything here resolves locally, so the list never waits on a
+ * network round-trip.
+ *
+ * The page list comes from the sidebar's exported SECTIONS, so a page can
+ * never be reachable from one surface and invisible in the other.
+ */
+import { useQuery } from "@tanstack/react-query";
+import { ArrowRight, Boxes, CornerDownLeft, Search, Sparkles } from "lucide-react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { createPortal } from "react-dom";
+import { api } from "../api/client";
+import { SECTIONS } from "../layout/Sidebar";
+import { type PageId, useUIState } from "../state";
+import { useTheme } from "../theme";
+import type { Project } from "../types";
+
+type CommandKind = "page" | "project" | "action";
+
+interface Command {
+  id: string;
+  kind: CommandKind;
+  /** Matched + rendered label. */
+  label: string;
+  /** Muted right-hand context (section name, project name, ...). */
+  hint?: string;
+  /** Extra text folded into the match haystack but not rendered. */
+  keywords?: string;
+  icon: React.ComponentType<{ className?: string }>;
+  run: () => void;
+}
+
+/**
+ * Subsequence ("fuzzy") match, case-insensitive. Returns the matched indices
+ * so the caller can highlight them, or null when the query does not match.
+ *
+ * Scoring favours matches that start earlier and stay contiguous, which is
+ * what makes "sto" rank Storage above "Semantic Layer" even though both
+ * contain the letters.
+ */
+function fuzzyMatch(query: string, text: string): { score: number; indices: number[] } | null {
+  if (!query) return { score: 0, indices: [] };
+  const q = query.toLowerCase();
+  const t = text.toLowerCase();
+  const indices: number[] = [];
+  let ti = 0;
+  let score = 0;
+  let lastHit = -2;
+  for (let qi = 0; qi < q.length; qi++) {
+    const ch = q[qi];
+    const found = t.indexOf(ch, ti);
+    if (found === -1) return null;
+    // Contiguous runs and hits at the very start of the string are cheaper.
+    score += found - ti;
+    if (found !== lastHit + 1) score += 3;
+    if (found === 0) score -= 2;
+    indices.push(found);
+    lastHit = found;
+    ti = found + 1;
+  }
+  return { score, indices };
+}
+
+/** Render `text` with the fuzzy-matched characters tinted cyan. */
+function Highlight({ text, indices }: { text: string; indices: number[] }) {
+  if (indices.length === 0) return <>{text}</>;
+  const hit = new Set(indices);
+  return (
+    <>
+      {text.split("").map((ch, i) =>
+        hit.has(i) ? (
+          <span key={i} className="text-accent">
+            {ch}
+          </span>
+        ) : (
+          <span key={i}>{ch}</span>
+        ),
+      )}
+    </>
+  );
+}
+
+const KIND_LABEL: Record<CommandKind, string> = {
+  page: "page",
+  project: "project",
+  action: "action",
+};
+
+export function CommandPalette() {
+  const [open, setOpen] = useState(false);
+  const [query, setQuery] = useState("");
+  const [cursor, setCursor] = useState(0);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const listRef = useRef<HTMLDivElement>(null);
+
+  const { project, setProject, setBranchId, setPage } = useUIState();
+  const { theme, toggle } = useTheme();
+
+  // Projects are already cached by the top bar under this exact key, so
+  // opening the palette normally costs no request.
+  const projectsQ = useQuery<{ projects: Project[] }>({
+    queryKey: ["projects"],
+    queryFn: () => api.get("/projects"),
+    enabled: open,
+  });
+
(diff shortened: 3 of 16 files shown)


padak (commented):
# Review of #658 — feat(ui): 0.89.0 catch-up — jobs/config actions, trash & restore, tokens, palette

> (lint, format, tests) are confirmed via `make check` / `tsc` / `vite build`,
> not duplicated here.

## Summary

This is a large, frontend-only catch-up PR (`web/frontend`, 2305/-398, no Python touched) adding job re-run/terminate, a config Drawer with Run-job/Delete + a Trash&Restore tab, a new Tokens page, table `definition` rendering, click-to-edit column descriptions, a billing-credits dashboard tile, a Flows Notifications tab, and a Ctrl+K command palette. The PR description is unusually precise about which server route backs each feature, and every claim I checked against the actual FastAPI router/Pydantic body in `src/keboola_agent_cli/server/routers/*.py` held up — including the `job.configId` → `job.config` field-name fix that the PR itself calls out. `tsc --noEmit` and `vite build` both pass clean. I found one genuine payload-correctness bug that fits squarely in this review's focus area: the per-job "re-run" action drops the job's `branch_id`, so re-running a job that originally executed on a dev branch silently re-runs it against the default/production branch instead. Verdict is REQUEST CHANGES for that one item; everything else (destructive-confirm coverage, secret handling on Tokens, design-language consistency) checked out clean.

## Verdict

- **Blocking findings**: 1
- **Nits**: 1

## Blocking findings

### `[B-1]` `web/frontend/src/pages/Jobs.tsx:169-174` — per-job "re-run" drops `branch_id`, silently retargets the default branch

`JobActions`'s `rerun` mutation posts only `component_id` and `config_id` to `POST /jobs/{p}/run`:

```ts
mutationFn: () =>
  api.post(`/jobs/${encodeURIComponent(job.project_alias)}/run`, {
    component_id: job.component,
    config_id: job.config,
  }),
```

The Queue API job resource the row is built from does carry `branchId` (confirmed live: `job_service.py:718` filters jobs by `j.get("branchId")`), but the `Job` TS interface (`types.ts`) never declares it, so it's silently unavailable to this call site. `JobRun.branch_id` on the server (`server/routers/jobs.py`) defaults to `None` when omitted, which resolves to the **default/production** branch. Concretely: a job originally run against a dev-branch config, re-run via this button, executes against the *production* config instead — a materially different configuration, potentially writing to the wrong tables/branch. This is a real behavior gap, not a display bug — the docstring above `JobActions` only calls out the `configData`-replay limitation, not this one.

Contrast with `ConfigDetail`'s "Run job" action a few files over (`Configs.tsx`), which correctly threads `branch_id: branchId ?? undefined` from `useUIState()` because that Drawer is opened from a branch-scoped `configs` query. `JobActions` has no equivalent context to reuse, which is exactly why the job's own `branchId` needs to be surfaced and sent.

Fix: add `branchId?: number | null` to the `Job` interface (it's already on the wire, just untyped) and send `branch_id: job.branchId ?? undefined` in the `rerun` mutation body.

## Non-blocking findings

### `[NB-1]` No CI coverage for `web/frontend` (tsc / build)

`.github/workflows/ci.yml` has no step that runs `tsc --noEmit` or `vite build` against `web/frontend` (grepped for `frontend`/`web/` — only a comment mentions it). This PR passes both when run manually, but it means a future frontend PR's type errors ship straight to `main` undetected. Pre-existing gap, not introduced by this PR — flagging per the review's "verify, don't assume" mandate since I had to run the build myself to confirm this PR is clean. Worth a follow-up issue, not a blocker here.

### `[NB-2]` `web/frontend/src/state.tsx` — `useManageTokenPrompt` removed outside the PR's stated scope

The PR description doesn't mention removing the manage-token prompt helper. It's confirmed dead code (`grep` found zero remaining call sites, `tsc` is clean), so it's safe, but it's an unrelated cleanup riding along in a large feature PR. Worth a one-line mention in the PR description so a future `git blame` on this deletion doesn't have to reconstruct the "why" from an 18-file diff.

## Nits

- `[NIT-1]` `web/frontend/src/pages/Tokens.tsx` — the `never` / `unknown` / `error` status semantics (from `STATUS_TITLES`) are excellent and match the CLI's `--with-last-used` documentation precisely; no action needed, calling it out only because it's the kind of fidelity the rest of the PR should be held to.

## Verification log

- `gh auth status` → authenticated as `padak` ✓
- Read `CONTRIBUTING.md` (Checklist for new CLI commands, Plugin synchronization map, Releasing a new version) and `CLAUDE.md` §17/`## All CLI Commands` — this PR adds/removes/renames zero CLI commands (frontend-only), so `OPERATION_REGISTRY`, `AGENT_CONTEXT`, `keboola-expert.md`, `commands-reference.md`, `gotchas.md` are all correctly untouched. No silent-drift findings.
- `gh pr view 658 --json title,body,files,additions,deletions,baseRefName,headRefName,labels,state` → OPEN, `main`←`feat/ui-0890-catchup`, +2305/-398, 15 files, conventional `feat(ui):` prefix matches (new page + several new actions) ✓
- `gh pr diff 658` → 3199-line diff fetched to scratch, cross-checked file-by-file against `gh pr view --json files` ✓
- Isolated detached `git worktree add ... origin/feat/ui-0890-catchup --detach` — invoking worktree's HEAD (`claude/keboola-cli-issues-pr-review-eb4380`) confirmed unmoved before and after ✓
- `git log --oneline origin/main..HEAD` → 3 commits (`feat(ui): ... catch-up`, `fix(ui): clear optimistic column override; portal ConfirmModal itself`, `fix(ui): job rows carry config, not configId`) — all UI-scoped, consistent with the PR ✓
- Cross-checked every claimed route in the PR body against the live FastAPI router source: `jobs.py` (`JobRun`/`JobTerminate` bodies), `configs.py` (delete/restore/trash-list), `token.py` (`CreateTokenBody`/`TokenIdBody`), `storage.py` (`DescribeColumns`, `table_detail`), `billing.py` (`get_credits` response fields `remaining`/`consumed`/`total`/`remaining_minutes`/`error_code: PAYG_NOT_AVAILABLE"`), `notifications.py` (`list_subscriptions` unfiltered-by-design + `project_wide_excluded`) — all field names and shapes match the frontend types byte-for-byte ✓
- Confirmed the PR's headline claim: `services/job_service.py` raw job dicts use `config` (not `configId`) — `job_service.py:718` (`j.get("branchId")`) also confirms `branchId` IS present on the raw dict, which is what B-1 leans on ✓
- `cd web/frontend && npm ci` → 392 packages, 0 vulnerabilities ✓
- `npx tsc --noEmit` → clean, no output, exit 0 ✓
- `npm run build` (`tsc -b && vite build`) → built in 1.42s, only pre-existing chunk-size warning (mermaid/cytoscape vendor chunks, unrelated to this PR) ✓
- `npx vitest run` → "No test files found" — confirmed pre-existing (zero `*.test.*`/`*.spec.*` files anywhere in `web/frontend/src`), not a regression introduced by this PR
- Grepped the full diff for `console.log`/`debugger`/`TODO`/`FIXME` in added lines → none ✓
- Grepped the full diff for token/secret/password patterns → all matches are the intentional Tokens-page UI (`SecretPanel`, `RevealedSecret`, etc.); secret is held in React state only, cleared on `closeDrawer()`/`onClose()`, never sent to `console` or a query string ✓
- Manually traced `ConfirmModal` coverage: config delete (`Configs.tsx`), token delete + token refresh (`Tokens.tsx`, both `danger`), job terminate (`Jobs.tsx`, `danger`) — all four destructive/high-consequence actions gate through `ConfirmModal`; config restore and column-description edits (non-destructive writes) correctly do NOT require a confirm ✓
- Verified the `ConfirmModal` → `createPortal(..., document.body)` fix — a genuine correctness fix for confirms raised from inside a `Drawer` (stacking/backdrop-blur containing-block issue), applies to every consumer listed above ✓
- Grepped `Sidebar.tsx` `SECTIONS` (now exported, shared with `CommandPalette.tsx`) — `Tokens` entry present under "Manage" section, matches PR description "sidebar entry under MANAGE" ✓

- Is the missing `branch_id` on job re-run (B-1) intentional for some reason not captured in the docstring (e.g. "re-run always targets production by design")? If so, worth a one-line comment next to the `rerun` mutation saying so explicitly, since the current comment only addresses the `configData` limitation and reads as if branch fidelity were preserved.

padak:
Thanks — B-1 is a real bug and is fixed. Mapping each finding to its commit:

## `[B-1]` re-run drops `branch_id` — **fixed in `ed66cc2`**

Took the pass-through option, since both halves were already available:

- `branchId` is on the raw Queue API row — `JobService._fetch_project_jobs` (`job_service.py:717`) filters on `j.get("branchId")`, so it is definitely on the wire.
- The router **does** expose branch targeting: `JobRun.branch_id: int | None` (`server/routers/jobs.py:30`).

So no `canRerun` gating was needed — this is a pass-through, not a new capability.

Three details worth flagging beyond the literal fix:

1. **Typed loosely on purpose.** `branchId?: number | string | null`. The Queue API is inconsistent about the numeric-vs-string form — which is exactly why the service compares it as `str(j.get("branchId"))` rather than `==`. A new `jobBranchId()` helper coerces to the `int | None` the router declares and **drops anything non-numeric** rather than sending a value FastAPI would 422 on.
2. **The tooltip now names the target branch** (`…on branch #123` / `…on the default branch`). Threading `branch_id` through is only half the fix; the affordance also has to be honest *before* the click.
3. **Answering the open question** — no, it was not intentional. I updated the `JobActions` docstring, which previously called out only the `configData`-replay limitation and therefore read as if branch fidelity were already preserved. It now says explicitly that the branch *is* preserved and points at the helper.

## `[NB-2]` `useManageTokenPrompt` removal outside stated scope — **fixed (PR description)**

Fair. Added a **Scope note** section to the PR body recording that the deletion is an unrelated dead-code cleanup (zero call sites, verified by grep before removal) and that the `ManageTokenModal` that superseded it is untouched. No code change — the removal itself is still correct.

## `[NB-1]` no CI coverage for `web/frontend` — **acknowledged, deliberately not fixed here**

Agreed it is a real gap, and I hit it directly: this PR shipped three commits' worth of frontend bugs that only manual `tsc`/build runs and a live browser pass caught. But adding a workflow step means editing `.github/workflows/ci.yml`, which is outside this PR's frontend-only remit and would put a new required-check on `main` inside an already-large feature PR — a change that deserves its own review, not a rider on this one. Flagged for a follow-up rather than filed as an issue, since that is the maintainer's call.

Worth noting for whoever picks it up: `npx vitest run` currently reports "No test files found" (zero `*.test.*` anywhere in `web/frontend/src`), so the follow-up is really two things — a `tsc --noEmit && vite build` gate, and a decision about whether the frontend gets tests at all.

## `[NIT-1]` — no action, thank you.

---

Two other fixes landed on this branch since your review snapshot, both worth a second look:

- **`c44a4f9`** — the Queue API job row names the config id `config`, not `configId`. `types.ts` had the wrong name, so the Jobs table's Config column had **always** rendered empty, and the new re-run button's `canRerun` gate was permanently false — the button never rendered at all. (Your B-1 was found by reading the code; this one only surfaced in a live browser pass, which is a decent argument for NB-1.)
- **`e65fcc6`** — the optimistic column-description override was never cleared on success, so an edited column masked later server values for as long as the drawer stayed mounted; and `ConfirmModal` now portals to `<body>` itself.

**Note for a maintainer, out of scope here:** the same `configId` mistake exists on the Python side. `output.py:411` renders the CLI's `job list` Config column as `str(job.get("configId", job.get("config_id", "")))`, so `kbagent job list` has also always printed an empty Config column. It survived because the fixtures in `tests/test_output.py` and `tests/test_services.py` hand-write `"configId"` — they assert against the same invented key the renderer reads. Any fix there needs to correct the fixtures too, or it will pass tests that were never checking anything real.

`npx tsc --noEmit` and `npm run build` clean on `ed66cc2`.

PR #681, opened 2026-08-23
serve: honor the root-level --config-dir, and fix two stale permissions comments

### LINKED ISSUE #679: serve: root-level --config-dir is silently ignored (kbagent --config-dir X serve serves a different dir)
## Summary

`kbagent serve` has its own `--config-dir` option, separate from the root-level one. The root-level flag is **silently ignored** by `serve`: `kbagent --config-dir X serve` does not serve `X`, it serves whatever the default resolution chain finds (env var, local `.kbagent` walk-up, or the global directory). No warning, no error — the server starts and answers happily from a different set of projects than the one the caller named.

Every other command honours the root flag, so the flag is not inert in general — it is inert *for this one command*, which is exactly the shape that goes unnoticed.

## Reproduction (verified on `main` @ 763c303)

Fixture: a config dir containing exactly one project alias, `alpha-only`.

```bash
# the CLI itself honours the root flag
$ kbagent --config-dir /tmp/dirA --json project list
CLI sees in dirA: ['alpha-only']

# case 1 -- ROOT-level flag, serving /projects
$ KBAGENT_SERVE_TOKEN=probe1 kbagent --config-dir /tmp/dirA serve --port 8895 --no-banner
$ curl -s -H "Authorization: Bearer probe1" localhost:8895/projects
served: ['pabu', 'padak', 'conditional-flow-testing', 'kids-app-factory']   # <- the GLOBAL config

# case 2 -- SERVE-level flag, same fixture
$ KBAGENT_SERVE_TOKEN=probe2 kbagent serve --config-dir /tmp/dirA --port 8894 --no-banner
$ curl -s -H "Authorization: Bearer probe2" localhost:8894/projects
served: ['alpha-only']                                                     # <- correct
```

## Cause

`serve_command` forwards only its own option:

- `commands/serve.py:277` — `create_app(config_dir=config_dir, ...)` where `config_dir` is the **serve-level** `--config-dir` parameter.
- The root callback resolves its own directory into `ctx.obj["config_store"]` (`cli.py:275-276`, stored at `cli.py:337`), and `serve_command` never reads it.
- The root callback does not export `KBAGENT_CONFIG_DIR` either (grep for `environ[`/`setdefault` in `cli.py` returns nothing), so there is no indirect channel: when `serve --config-dir` is omitted, `create_app` falls through to `resolve_config_dir(None)` → env var → local walk-up → global (`config_store.py:186-204`).

## Why it matters

1. **Wrong projects served, silently.** The realistic invocation is a script or systemd unit that sets the root flag once for the whole command line. It gets a server bound to somebody else's project set, and nothing in the output says so.
2. **Security control does not apply.** Since #677 the serve permission engine is composed from the config dir `create_app` resolves, so the *served* directory's persisted `permissions` policy is the one that applies. With the root flag that directory is not the one the caller named — a deny policy stored next to the projects a user meant to serve is simply not read. (This is a direct consequence of the same bug, not a separate one; #677 deliberately made the engine follow the served dir precisely so policy and services can never disagree with each other.)

## Options

1. **Propagate**: have `serve_command` fall back to the root-resolved config dir when its own `--config-dir` is absent (read `ctx.obj["config_store"].config_dir`). Makes the root flag behave like it does for every other command; the serve-level flag stays as the explicit override.
2. **Reject the ambiguity**: fail fast when both flags are given with different values, and otherwise propagate as in (1).
3. **Document only**: keep the current behaviour and say loudly in `--help` + `docs/web-server.md` that `serve` reads only its own `--config-dir`. Cheapest, but leaves a silent-wrong-answer path in place.

Option 1 (optionally with 2's guard) looks right: the current behaviour has no upside — nobody deliberately passes a root `--config-dir` expecting it to be dropped.

## Notes

- Pre-existing behaviour, not a regression: it predates #677, which only surfaced it (that PR documents "pass `--config-dir` to `serve` itself" on three doc surfaces as a stopgap).
- Related: #655 (permission firewall not enforced on REST routes) — #677 enforces it for `/auth/*` only.

padak:
Closes #679.

## 1. `kbagent serve` ignored the root-level `--config-dir`

`serve` is the only subcommand in the repo that defines its own `--config-dir`, and it forwarded *only* that one into `create_app`. The directory the root callback had already resolved into `ctx.obj["config_store"]` was never read, and the callback exports no `KBAGENT_CONFIG_DIR`, so there was no indirect channel either. `kbagent --config-dir X serve` therefore served whatever the default chain found — silently, with no warning.

Two consequences, both reproduced live before the fix:

- **The wrong projects were served.** `kbagent --config-dir dirA serve` answered `/projects` with the 51 aliases of the *global* config instead of dirA's single `alpha-only`.
- **The wrong directory's permission policy applied.** Since #677 the serve permission engine is composed from the directory `create_app` resolves, so a persisted deny policy stored next to the projects the caller named was not read at all. In the end-to-end fixture, `POST /auth/register-projects` returned **401 SESSION_NOT_FOUND** (policy never loaded) where it should have returned **403 PERMISSION_DENIED**.

**Fix — most-specific-wins precedence**, matching the pattern `commands/repl.py` already uses when forwarding the root flag into a subcommand:

1. `serve --config-dir X` given → serve X *(unchanged; every existing invocation keeps working)*
2. otherwise, root `--config-dir Y` given explicitly → serve Y *(this is the fix)*
3. otherwise → today's chain: `KBAGENT_CONFIG_DIR` → `.kbagent` walk-up → global *(unchanged)*

"Given explicitly" is read off `ConfigStore.source == "cli-flag"` rather than by comparing or re-resolving paths. Only that source travels: for `env-var` / `local` / `global`, letting `create_app` resolve on its own yields the identical directory, and propagating them would pin a resolution made at a different moment. Deliberately **no** fail-fast and no warning when both flags are given and differ — that would break invocations that work today, and the precedence is unambiguous.

The serve `--config-dir` help text now states the precedence; its old wording ("matches kbagent --config-dir") is what made the bug plausible in the first place.

### Verification

Live A/B against the real CLI with two fixture config dirs, plus 6 new tests in `tests/test_server_permissions.py` — five precedence cases (root-only, serve-only, both, neither, env-var-set-without-root-flag) asserting the `config_dir` kwarg `create_app` receives, and one end-to-end test that persists a deny policy in the served directory and asserts `403 PERMISSION_DENIED` on `POST /auth/register-projects` while a sibling operation outside the policy is not denied. A negative control confirmed exactly the two new-behavior tests fail without the fix.

| Invocation | before | after |
|---|---|---|
| `--config-dir dirA serve` | global config (51 aliases) | `alpha-only` |
| `serve --config-dir dirA` | `alpha-only` | `alpha-only` |
| both flags | serve-level wins | serve-level wins |
| neither | fallback chain | fallback chain |

## 2. Two comment corrections in `permissions.py`

Found while filing #679; both are fallout of #677.

- **A `(since vNEXT)` placeholder shipped in 0.90.1.** `auth.projects` still carried it on `main`, because `scripts/check_version_gates.py` scanned `CLAUDE.md`, `docs/*.md` and `plugins/kbagent/**/*.md` only — a placeholder anywhere under `src/` was invisible to it and shipped unresolved. Pinned to `0.90.1`, and the cause is fixed too (below).

## 3. The version gate now scans `src/**/*.py`

Added on review: pinning the placeholder alone would let the next feature PR ship the same defect, since writing `(since vNEXT)` in a Python comment is a pattern `permissions.py` itself establishes as normal. Verified the widened scan would have caught the original — run against `git show 11724656:src/keboola_agent_cli/permissions.py`, it flags line 29. Scanned files 67 → 295, gates 476 → 529, `vNEXT` residue unchanged at 7.

`scripts/*.py` stays **out**, deliberately: the gate script must name the token it hunts (`VNEXT_TOKEN = "vNEXT"`, its own `--release` usage example), so scanning itself self-flags forever. The comment says so, to stop the next person retrying it.

`INLINE_CODE_RE` also learned RST double-backtick spans, since Python docstrings use them and a ``(since vNEXT)`` in one would otherwise false-positive. Measured: markdown residue identical at 7, so no live gate is lost. The double-backtick alternative must come first in the alternation — with the single-backtick one first, the pattern matches the empty span between the two opening backticks and leaves the token exposed.

**The wider scan immediately found a real pre-existing defect**: `commands/project.py:839` carried `(since v0.26.1)`, a version that was never released (`git tag -l "v0.26*"` → only `v0.26.0`; no such GitHub release; not a `CHANGELOG` key). Corrected to `v0.29.0` — `CHANGELOG["0.29.0"]` describes those seven member/invitation commands verbatim, and the commit that added the line is `c7d70788 feat(0.29.0): … + member lifecycle (#256)`, whose message still carries the pre-renumber `(since v0.26.1)` draft. Same failure mode as the 0.90.1 placeholder, three months older.
- **The `http.*` block claimed serve-side enforcement that did not exist.** It read *"The serve's own routes enforce their own permissions on top."* — #655 flagged this as load-bearing and false, and it was: before #677 no serve route consulted the engine, so an `http.post` classified `write` met no second check anywhere. Since 0.90.1 it is true for `/auth/*` and nothing else, so the comment now says exactly that and names #655.

## Docs

`docs/web-server.md`, `CLAUDE.md` and `gotchas.md` carried #677's stopgap advice ("pass `--config-dir` to `serve` itself"), which describes the bug rather than the intended behavior — rewritten to the fixed precedence, with the old trap kept intact for readers on 0.90.1. `commands-reference.md` and `context.py`'s `AGENT_CONTEXT` are on CONTRIBUTING's synchronization map for flag-behavior changes and were updated too. New behavior is gated with the literal `(since vNEXT)`; `context.py` deliberately carries no version tag, since a placeholder in a `.py` file is exactly what the gate cannot resolve.

## Testing

`make check` green: 6123 passed, 12 skipped. `ty` clean for every touched file (one pre-existing unrelated diagnostic in `scripts/hatch_build.py` remains). `check_command_sync.py`, `check_version_gates.py`, ruff check + format all OK. The version-gate tests grew by 9, mutation-checked to fail against the pre-change state. No version bump, no changelog entry.


src/keboola_agent_cli/permissions.py
```diff
@@ -26,7 +26,7 @@
     # tokens, never a real credential) -- same risk class as login/logout,
     # not the "admin" class `project add` uses for a pasted static token.
     "auth.register-projects": "write",
-    # Serve-only (since vNEXT): `GET /auth/projects` lists the session's
+    # Serve-only (since 0.90.1): `GET /auth/projects` lists the session's
```
zajca, line 29: Reviewed by Opus.

This hunk fixes a `(since vNEXT)` placeholder that survived the 0.90.1 release, but only the symptom. The release gate that is supposed to catch exactly this (`scripts/check_version_gates.py --release`, wired via `Makefile:124` and `.github/workflows/ci.yml:187`) resolves its file list from `SCANNED_GLOBS` at `scripts/check_version_gates.py:73-78`:

```
SCANNED_GLOBS = ("CLAUDE.md", "docs/*.md", "plugins/kbagent/**/*.md", "plugins/kbagent/.claude-plugin/CLAUDE.md")
```

`src/**/*.py` is not in that list, which is precisely why the placeholder on this line shipped in a released version undetected. As soon as the next feature PR writes `(since vNEXT)` in a Python docstring or comment — a pattern this file itself establishes as normal — the same defect ships again, and neither the per-PR gate nor the release gate will flag it.

Suggested fix: add `src/**/*.py` (and, if desired, `scripts/*.py`) to `SCANNED_GLOBS`. The existing inline-code-span exemption in `find_vnext_residue` already handles the prose-about-the-placeholder case, so genuine mentions such as the ones in `changelog.py` would not false-positive as long as they stay inside backticks. Without this, the PR leaves the detection gap it was created by fully open.

src/keboola_agent_cli/permissions.py
```diff
@@ -323,7 +323,10 @@
     "semantic-layer.reference-data.delete": "destructive",
     # Raw HTTP client against `kbagent serve` (used by AI subprocesses).
     # Categorised by the underlying HTTP method: GET = read, mutating verbs
-    # = write. The serve's own routes enforce their own permissions on top.
+    # = write. These keys are the ONLY firewall an `http.*` call meets for
```
zajca: Reviewed by Opus.

This line rewrites the `http.*` comment for accuracy but keeps the preceding claim intact: "Categorised by the underlying HTTP method: GET = read, mutating verbs = write." That does not match the registry three lines below — `"http.delete": "destructive"` (line 333). DELETE is a mutating verb, so a reader following the stated rule would expect `write` and conclude either that the comment or the classification is a mistake.

This matters more than a typo here because the same comment is the security rationale for these four keys, and the PR's whole purpose is to make that rationale trustworthy: the paragraph now correctly warns that these keys are the only firewall for most routes, and a reader auditing the `http.*` risk classification against the stated rule hits an immediate contradiction on the one verb whose classification is strictest.

Suggested fix: state the actual mapping, e.g. "GET = read, POST/PATCH = write, DELETE = destructive."

zajca (commented):

src/keboola_agent_cli/permissions.py
padak: Fixed the cause, not just the symptom: `src/**/*.py` is now in `SCANNED_GLOBS` (e3f650b8). Verified it would have caught the original defect — running the new scan against `git show 11724656:.../permissions.py` flags line 29. Scanned files went 67 → 295, gates 476 → 529, residue unchanged at 7. `scripts/*.py` is deliberately still excluded and the comment says why: the gate script must NAME the token it hunts (`VNEXT_TOKEN = "vNEXT"`, its own `--release` usage line), so scanning itself self-flags forever. I also widened `INLINE_CODE_RE` to understand RST double-backtick spans, since Python docstrings use them — measured: markdown residue identical at 7, so no live gate is lost. The wider scan immediately surfaced one real pre-existing defect: `commands/project.py:839` carried `(since v0.26.1)`, a version never released — corrected to `v0.29.0` (CHANGELOG["0.29.0"] describes those seven member/invitation commands verbatim; the commit that added the line is `c7d70788 feat(0.29.0)`). +9 tests, mutation-checked to fail against the pre-change state.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/keboola_agent_cli/permissions.py
padak: Fixed in b690c2c3. The opening line now states the mapping the registry actually implements — GET = read, POST/PATCH = write, DELETE = destructive — in the same taxonomy declared at the top of the file. I also recorded why the extra rung is load-bearing rather than cosmetic: `apply_firewall_flags` appends `cli:destructive`, which matches only the destructive class, so classifying `http.delete` as `write` would let a `--deny-destructive`-only session delete straight through the REST boundary. Checked the mapping covers every `http.*` key with nothing left over (four keys, four leaf commands in `commands/http_client.py`). The `/auth/*`-only scope sentence and the #655 reference are untouched.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Issue #119, opened 2026-04-11
sync push collapses transformation script[] array → runtime fails


frantisekrehor:
## Summary

`sync pull` → `sync push` roundtrip collapses `parameters.blocks[].codes[].script` from N elements to 1 whenever the original config has more than one element per code block. KBC API accepts the flattened config silently; runtime fails at job execution because each `script[]` element is supposed to map 1:1 to a separate executable unit.

Not specific to any particular SQL shape — triggers for any `code` containing multiple script elements (CREATE + INSERT + UPDATE, TRUNCATE + MERGE, multiple Python blocks, etc.).

## Impact

- **SQL transformations** (Snowflake / BigQuery / Redshift / Synapse): runtime error `Actual statement count N did not match the desired statement count 1`, SQL state `0A000` — driver default `MULTI_STATEMENT_COUNT=1`.
- **Python / R transformations**: named sub-blocks merge into one element; runtime may still succeed but UI structure is lost.
- No warning at push time. Failure only at job execution.
- `sync push` is effectively unusable as the bulk-edit path for any transformation with multi-statement code blocks.

## Reproduction (~2 minutes)

Needs any transformation with `len(script) > 1` in at least one code block.

```bash
# 1. Create test branch
kbagent --json branch create --project {alias} --name "repro"
# → e.g. branch_id=466659

# 2. Verify starting state
kbagent --json config detail --project {alias} --branch 466659 \
  --component-id keboola.snowflake-transformation --config-id {id} \
  | jq '.data.configuration.parameters.blocks[0].codes[0].script | length'
# → 6

# 3. Clean sync pull
mkdir /tmp/repro && cd /tmp/repro && git init -q && git commit --allow-empty -m init -q
kbagent sync init --project {alias} --directory /tmp/repro --git-branching
kbagent sync branch-link --project {alias} --branch-id 466659 --directory /tmp/repro
kbagent sync pull --project {alias} --directory /tmp/repro

# 4. Any edit (or none — even no-op push reproduces)
echo "-- test" >> /tmp/repro/{branch}/transformation/{component}/{config}/transform.sql

# 5. Push
kbagent sync push --project {alias} --directory /tmp/repro

# 6. Array has collapsed
kbagent --json config detail --project {alias} --branch 466659 \
  --component-id keboola.snowflake-transformation --config-id {id} \
  | jq '.data.configuration.parameters.blocks[0].codes[0].script | length'
# → 1   ← BUG

# 7. Run job → error
[...]

PR #694, opened 2026-08-25
fix(sync): push stamps API-derived baselines; one canonical script[] shape; statement-boundary markers (#686)

### LINKED ISSUE #686: sync push stamps pull_config_hash from disk while diff compares an API-derived hash — every pushed config stays "REMOTE MODIFIED" forever (mirror of #466)
## Summary

`sync push` stamps `pull_config_hash` with a hash computed **from the files on disk**, while `sync pull` and `sync diff` compute the remote-side hash **from the API response**. For any config where the local↔API round-trip is not hash-stable, those two producers disagree — so after a successful `sync push` the config is reported `~ REMOTE MODIFIED` by every subsequent `sync diff`, forever, even though the remote and the working tree are byte-identical. Only a real `sync pull` clears it.

The stored hash exists precisely to avoid that round-trip — the diff path says so itself:

```python
# For files unchanged since pull, use the stored pull_config_hash
# directly (avoids lossy code extraction roundtrip).
```

…but `push` fills that field *with* a value produced by the lossy round-trip.

## Relation to #466 (please read together)

This is the **mirror image** of #466 and it survived the v0.72.0 fixes. #466 was `sync diff/status` falsely reporting **"in sync"** (false negative) plus reconcile ergonomics; this is `sync diff` falsely reporting **drift** (false positive), and the root cause is on the *push* side — nothing in the #466 / #467 / #482 / #497 fixes touches it. The `--theirs` reconcile path shipped for #466 does clear the symptom (any real pull does), but it cannot prevent it: the next deploy re-creates it.

Net effect for us is the same conclusion as #466's impact section, from the opposite direction: **`sync diff` still cannot be trusted as a local-vs-production audit** — now because it reports drift that does not exist.

## Impact — why this is urgent for us

We keep a git repo in lock-step with production: each project is a `sync` working tree, and a production deploy is `sync push` after a merged PR (~20 projects on one stack).

1. **Every deploy now needs a second, content-free PR.** The documented "refresh the manifest baseline after deploy" chore cannot work, because committing the manifest that `push` just wrote stores the wrong hash form. To get a truthful baseline into git we have to open a *separate* PR whose entire diff is `manifest.json` — after every single deploy. That is what pushed us to file this.
2. **Real drift hides in the noise.** Our repo currently carries 18 phantom `REMOTE MODIFIED` configs against 2 real UI changes. We could only separate them by pulling all of production into a scratch tree and comparing file by file — exactly the kind of manual audit #466 was supposed to make unnecessary.
3. It is self-inflicted by the tool's own happy path: the more disciplined you are about deploying through `sync push`, the more phantom drift you accumulate.

## Root cause (paths on `main`, v0.90.1)

| step | file:line | `pull_config_hash` computed from |
|---|---|---|
| `sync pull` | `src/keboola_agent_cli/services/sync_service.py:787-793` | `config_hash(api_config_to_local(component_id, cfg, config_id))` — the **API response** |
| `sync push` (modified) | `sync_service.py:1710` → `_compute_config_hashes` (`:1908-1930`) | `self._read_config_file(config_dir)` + `merge_code_files(...)` — the **files on disk** |
| `sync diff` | `sync_service.py:1251` builds the remote side with `api_config_to_local(...)`, then compares it against the stored `pull_config_hash` | mixed producers ⇒ mismatch |

The same asymmetry is in the create path (`sync_service.py:1652`) and in the row push path (`services/_sync_push_ops.py:239`).

Every affected config in our data has at least one code whose API body holds `parameters.blocks[].codes[].script` as an **array of several statements**, which `transform.sql` cannot represent (there are no statement boundaries in the file, and `_config.yml` carries no `blocks`/`codes`/`script` at all), so `merge_code_files` cannot reconstruct a hash-identical body. We did not isolate the trigger further than that.

## Evidence

Observed on kbagent **v0.59.0**, one stack, 21 projects, 244 configs. (Still on 0.59.0 because auto-update cannot cross the #424 package rename, as you noted on #466 — we will reinstall and re-verify. All three code paths above are unchanged on `main` @ v0.90.1, which is why we are filing now rather than waiting.)

Committed baseline vs a fresh `sync pull --all-projects` of the same production state, compared field by field:

- **224** configs — baseline identical.
- **18** configs — **`pull_config_hash` differs and nothing else**: `pull_hash` and `pull_extra_hashes` (`transform.sql`, `_description.md`) match exactly, and the pulled files are byte-identical to the tracked ones (0 differing files across 17 project trees, 11–139 files each).
- **2** configs — genuinely changed in the UI (file hashes differ too). These are the true positives the noise was hiding.

The discriminator is **who wrote the remote's current version** (`config detail` → `currentVersion.creatorToken.description`), and it was exceptionless:

| remote version last written by | stored `pull_config_hash` |
|---|---|
| `kbagent-cli [...]` — i.e. a `sync push` | **mismatches** (18/18) |
| a plain user token — i.e. the UI | matches |

Ruled out as the discriminator: multi-statement codes *per se*. Several configs whose codes hold up to 39 statements each compare clean — because their baseline was written by a pull, not a push.

Finally, `sync diff` run against the freshly pulled tree reports **all 21 projects `in sync`**, confirming the API-derived form is the one `diff` accepts.

## Repro

1. In a `sync` working tree, pick a `keboola.snowflake-transformation` whose API body has a code with more than one statement in `script`.
2. Change anything in it locally and `kbagent sync push --project <alias>`. Push succeeds.
3. `kbagent sync diff --project <alias>` → `~ REMOTE MODIFIED … parameters.blocks[0].codes changed`, with no local edits and nothing changed on the remote since. It never goes away.
4. `kbagent sync pull --project <alias> --no-jobs --no-storage` → now `in sync`, and the only thing that changed on disk is `manifest.json` (`pull_config_hash`). No config file is touched, which is the proof that step 3 was phantom.

## Suggested fix

Make `push` stamp the baseline with the same producer `pull`/`diff` use: hash the config **as the API returns it** after the write (the update response body, or a re-fetch), via `config_hash(api_config_to_local(...))`, instead of `_compute_config_hashes(config_dir, ...)`. Applies to the config update path, the create path and the row path.

If that is not desirable, then a push should not stamp `pull_config_hash` at all — leaving it stale is at least *visibly* stale, whereas today it is written with a value that can never match.

A regression test would be: push a config with a multi-statement code, then assert `sync diff` reports it in sync.

padak:
Fixes #686.

## What this fixes

Three defects sharing one root cause — `sync push` stamped the manifest baseline (`pull_config_hash`) from a hash computed **from the files on disk**, while `sync pull` and `sync diff` compute it **from the API response**, and the two sides held contradictory definitions of what `script[]` looks like:

1. **Phantom drift** (the reported symptom): after a successful `sync push` of a multi-statement SQL transformation, every subsequent `sync diff` reported `~ REMOTE MODIFIED … codes changed` forever, even though remote and working tree were byte-identical. Only a real `sync pull` cleared it — until the next deploy re-created it.
2. **Silent production rewrite** (found during triage, worse than the report): when the API's `script[]` elements carry no trailing semicolons (`["SELECT 1", "SELECT 2"]`), `transform.sql` had no statement boundaries left to recover, so push collapsed N statements into ONE element — the `MULTI_STATEMENT_COUNT=1` crash shape of #119/#120/#274 — and `diff` then said *in sync*.
3. **`is_disabled` phantom** (same class, non-script instance): a config disabled in the UI whose local YAML has no `is_disabled` key drifted permanently after any push, because the disk-derived baseline lacked the key the API-derived comparison emits.

## How

**Part 1 — push stamps the baseline from the API response** (all six sites: config create/update, row create/update, Phase C/D binding backfills). The mutation response is used only when it carries a complete `configuration`; an id-only response triggers a `get_config_detail()` / `get_config_row()` read-back; if that fails, the baseline is left visibly stale (+ a `warnings[]` entry in the push envelope) — never silently stamped from disk. `pull_hash` / `pull_extra_hashes` stay disk-derived (they describe local files; that is correct).

**Part 2 — one canonical `script[]` shape.** Runtime semantics decide: one array element = one executable statement (the premise of #119/#120/#274). `_normalize_scripts` is now component-aware — SQL transformations split per-element via the new single producer `canonical_sql_script()` (never joined first), everything else keeps the join. This is also why existing *push*-stamped baselines start matching diff without any user action: they were already split-shaped.

**Part 3 — `transform.sql` carries statement boundaries when semicolons can't.** Extraction verifies the round-trip; only when `split_statements()` cannot reproduce the canonical array does it insert `/* ===== STATEMENT ===== */` between statements — `;`-terminated trees stay byte-identical. Markers are guaranteed boundaries on merge, and `split_statements()` still runs *within* each segment (a hand-typed `; SELECT …` inside a marked segment splits correctly). A statement whose own text contains a marker-identical line suppresses markers with a warning (documented limitation). Auto-appending missing `;` was rejected: it changes content and Oracle/ODBC rejects trailing semicolons (ORA-00911).

**Migration — per-entry `config_hash_version`.** Part 2 changes the computed hash for pull-stamped baselines of multi-statement SQL configs, so a naive release would hand every existing tree a *third* flavour of phantom drift. Instead:
- versioned entries compare strictly; unversioned (pre-upgrade) entries match leniently against the new shape OR the legacy collapsed shape (recomputed from the raw API config via the old path — never by re-collapsing already-split data);
- leniency applies everywhere a stored hash meets a fresh API hash: diff base, force-pull conflict detection, pull idempotency;
- one clean `sync pull` migrates: re-extracts (writing markers where needed) and stamps `config_hash_version: 2` — but only when `_config.yml` AND all companion files are unchanged; a locally-modified tree keeps its files, old hash and no version stamp;
- the version key is never stamped next to a legacy-produced hash;
- pushing a *modified* SQL config from an unversioned tree whose only difference from remote is statement boundaries aborts that change with the new `SYNC_LEGACY_BOUNDARY` error ("run `sync pull` first") instead of silently rewriting production; genuine edits proceed.

## Verification

The issue's repro table now reads (was `PHANTOM`/`CHANGED` before):

```
['SELECT 1;', 'SELECT 2;']  -> sent ['SELECT 1;', 'SELECT 2;']  | content SAME    | diff in sync
['SELECT 1', 'SELECT 2']    -> sent ['SELECT 1', 'SELECT 2']    | content SAME    | diff in sync
['SELECT 1;\nSELECT 2;']    -> sent ['SELECT 1;', 'SELECT 2;']  | content CHANGED | diff in sync
['SELECT 1']                -> sent ['SELECT 1']                 | content SAME    | diff in sync
```

Row 3 is deliberate: one element packing two statements is not a canonical array (it is the #274 crash shape); push normalizes it. No file format can express "do not split here", and push has always sent 2 elements for it — only the phantom is gone.

- `tests/test_sync_baseline_stamping.py` (17 tests): the reporter's regression (pull → edit → push → diff in sync), statement-count mirrors with/without semicolons, the `is_disabled` instance, partial-response read-back, failed-fetch leaves baseline unstamped + warning, versioned-vs-unversioned migration incl. real-drift discrimination, pull migration preserving an edited `transform.sql`, the `SYNC_LEGACY_BOUNDARY` abort vs genuine-edit passthrough, force-pull leniency (A/B-verified).
- `tests/test_sync_code_extraction.py`: marker round-trip, marker collision fallback, within-segment re-split, broad-predicate SQL variants, the full repro table pinned as a parametrized test.
- `make check` green: 6189 passed, 12 skipped; ruff/ty/loc-check/error-codes/sentinel-guards all clean (`sync_service.py` net −78 code lines).

## Deliberately updated legacy tests

- `tests/test_sync_config_format.py::TestNormalizeScripts` (7 tests) — `_normalize_scripts` now takes `component_id`; they pass a non-SQL component and keep asserting the join semantics for that branch.
- `tests/test_sync_reconcile.py::test_adopted_by_id_push_registers_manifest_entry` — the bare mock now serves `get_config_detail` from the remote fixture (an id-only response no longer counts as a stampable baseline).
- `tests/test_sync_service.py::TestFreshCreateVariableBinding::test_push_resolves_bindings_end_to_end` — create mocks gained `get_config_detail`/`get_config_row` read-backs.

## Notes for the release PR

- No version bump, no `changelog.py` entry (per the release process). Doc surfaces (`gotchas.md`, `sync-workflow.md`, `CLAUDE.md`, `commands-reference.md`, `keboola-expert.md`) are tagged `(since vNEXT)`.
- New error code: `SYNC_LEGACY_BOUNDARY` (type `conflict`).
- Release note should say: **run `sync pull` once per project after upgrading** to migrate baselines to `config_hash_version: 2`; until then unversioned entries are matched leniently (no worse than the pre-upgrade blindness, and only for the one collapsed-shape difference).

## Follow-up

A separate PR wires `normalize_blocks_codes_script` (the #274 runtime-safety guard) into the sync push path as a belt-and-braces backstop — after parts 2+3 it is a no-op on this path by construction.


padak:
## Live E2E verification (project 5946, e2e-snowflake)

Ran an A/B harness (`~/kbagent/e2e/sync_baseline_stamp_e2e.py`, follows the local e2e convention: token stays in-process via ConfigStore, sync operations shell out to the CLI under test). Each phase gets its own throwaway `keboola.snowflake-transformation` with two codes — `["SELECT 1;", "SELECT 2;"]` and `["SELECT 3", "SELECT 4"]` (no trailing semicolons) — then: pull → edit `transform.sql` → push → diff.

**Control = installed v0.90.1** (validates the instrument — both defects must reproduce):

```
PASS: control: diff before push sees local edit
PASS: control: PHANTOM drift after push (bug reproduced)
PASS: control: silent rewrite reproduced (NoSemi collapsed to 1 element)
```

**Fix = this branch:**

```
PASS: fix: pulled transform.sql has marker for NoSemi
PASS: fix: marker count is exactly 1 (Semi needs none)
PASS: fix: manifest entry stamped config_hash_version=2
PASS: fix: diff before push sees local edit
PASS: fix: NO phantom drift after push (the #686 regression)
PASS: fix: edited statement landed remotely
PASS: fix: NO silent rewrite (NoSemi keeps 2 elements)
PASS: fix: unversioned entry still in sync (lenient match)
```

**Result: 11 pass, 0 fail.** The last check strips `config_hash_version` from the manifest entry and re-diffs — the pre-upgrade leniency path works against live data too.

PR #703, opened 2026-08-26
feat(service): merge-request Layer 2 - lifecycle, derived status, conflict resolution (DMD-1899)

martinsifra:
Layer 2 of the merge-request stack ([DMD-1899](https://linear.app/keboola/issue/DMD-1899)): `services/merge_request_service.py` over the Layer 3 namespace shipped in #556. No Layer 1 commands yet — those are DMD-1900, so none of the convention-#17 doc surfaces or E2E tests apply to this PR; unit tests only (`tests/test_merge_request_service.py`, 94 tests).

The design record — the L2 RFC plus the wire-truth notes every decision below is written down in — lives on branch [`ms/merge-requests-rfcs`](https://github.com/keboola/cli/tree/ms/merge-requests-rfcs/docs) (RFCs deliberately do not merge to main). A full self-review pass (2026-08-27) was applied in the final commit; see "Review fixes" below.

## What's in

- **Status-derivation polyfill** (pure module-level functions): `derived_state` (the UI list badge's decision table incl. the rejected / closed-by-creator reviewer overrides), `merge_blockers` + `mergeable` (a list, so concurrent blockers don't mask each other), `allowed_actions` (state-machine table), `viewer` (`is_creator`/`has_approved`, anchored on `verify_token`'s new `admin_id`). Every function reads the future server-serialized field first ([DMD-1988](https://linear.app/keboola/issue/DMD-1988)) and falls back to the local table — the fallbacks get deleted when Connection serializes.
- **Reads**: `list_merge_requests` (client-side `state` filter over the closed derived + raw vocabulary, typos refused), `find_merge_request_for_branch` (the resolver behind L1's optional `--mr-id`; a branch has at most one MR ever), `get_merge_request` (detail with full derived status; conflicts fetched only for open MRs).
- **Writes**: `create_merge_request` (targets the default branch itself; refuses the default as source with a readable error instead of the backend's 404), `update_merge_request`, `request_review` / `approve` / `request_changes` — all behind the `branches-merge-requests` pre-flight raising `FEATURE_NOT_ENABLED` (a missing feature is otherwise a 403 byte-identical to a role denial; SOX-fence assumption in the comment).
- **`merge`**: 409 remapped onto its two wire shapes — `MR_NOT_READY_TO_MERGE` (retryable; carries `storage.mergeRequests.notReadyToMerge`) vs `MR_MERGE_CONFLICT` (not retryable, names `merge-request conflicts` as the next step); backend counterpart asking for codes on every MR error is [DMD-1984](https://linear.app/keboola/issue/DMD-1984). Post-merge cleanup mirrors `BranchService.delete_branch` (conditional `active_branch_id` reset, sync-mapping unlink, best-effort with warnings); output says the source branch "is being deleted", never that it's gone.
- **Conflict resolution**: `list_conflicts`, `get_config_diff` flattened to a per-path `changed_by: ours|theirs|both` classification (+ `agreed: true` when both sides made the identical change; `ours_deleted`/`theirs_deleted` flags for tombstoned sides), and `resolve_conflict` where every mode goes through the rebase endpoint: `take=ours|theirs` composes the full replace body from the side's `diff` envelope with `version=theirs.version`, a deleted side collapses to the delete resolution (both directions), a custom body must spell out `name`/`rows`/`configuration` (rebase REPLACES — a defaulted key would be silent data loss). The branch is derived from the MR itself, never caller-supplied. The UI's reset-to-default alternative for take=theirs is tracked as [DMD-1987](https://linear.app/keboola/issue/DMD-1987).
- **Infra**: `http_base` now surfaces a Keboola user error's machine string `code` as `KeboolaApiError.details["api_error_code"]` (additive, all four raise sites); `TokenVerifyResponse` gains `admin_id`/`admin_name`; `FEATURE_BRANCHES_MERGE_REQUESTS` renamed to `BRANCHES_MERGE_REQUESTS_FEATURE`; the `isDefault` scan hoisted to `services.base.find_default_branch_id` and the config/sync/workspace copies migrated; `json_utils.compute_diff` split into structured `compute_diff_entries` + a byte-identical formatter.

## Review fixes (final commit, `tasks/pr-703-review.md`)

The self-review found one **critical wire-shape error**: the diff sides were assumed flat, but the verified shape (connection `ConfigurationVersionResponse` + `ConfigurationDiffData`) nests all content under a `diff` envelope with `version`/`isDeleted` as side metadata — the original take/classify code would have been dead on arrival against the live API, invisibly, because the test fixtures encoded the same wrong assumption. Now recorded in `docs/merge-requests-notes.md`'s wire-truth table. Also fixed: `resolve_conflict` could rebase into an unrelated branch (branch now derived from the MR), deleted-side asymmetry, `_same_id` on branch ids, `FEATURE_NOT_ENABLED`, spec'd mocks at the L3 seam, and regression tests for every finding.

## Second review: wire truth vs. Connection (2026-08-27)

An adversarial Opus review verified every wire assumption against the Connection source (report: `tasks/pr-703-opus-wire-review.md`): 7 CONFIRMED, 3 MISMATCH, all fixed in the final commit:

- **The conflict 409 is not code-less** — `ExceptionConverter` serializes `storage.mergeRequests.validation` top-level plus the conflicting configs in `params.errors`. The remap now matches both codes explicitly (unknown 409 codes pass through unmapped), and `http_base` surfaces `params` as `details.api_error_params` — the conflict list travels with the error.
- **`approve` exists only in `in_review`** (from `approved` the backend answers 422 — the UI button offering it there is wrong); `allowed_actions` corrected, `update` added to `in_merge`. With the non-SOX default of 0 required approvals, `approve` is 422 everywhere and `in_review` is unreachable.
- **`derive_state`'s `rejected`/self-`closed` rows are best-effort by wire design**: `reviewers[].status` needs a `review_requested` anchor that `skip_review` never writes, and explicit reviewers shadow non-reviewer decisions — so on a default non-SOX project those states are underivable from `reviewers[]`. The UI badge has the identical blind spot (our table is its port). Documented, not re-derived client-side — the reliable fix is server-side and is now recorded on DMD-1988 (derive from the activity log).

## Layer 1 RFC findings applied (2026-08-28, commit 07daa50)

Writing the Layer 1 command RFC (DMD-1900, PR #708) surfaced seven Layer 2 findings (`tasks/dmd-1899-findings-from-layer1.md`); all verified against Connection and applied:

- **`get_config_diff` derives the branch from the MR** (signature now `alias, merge_request_id, component_id, config_id`) — the same make-the-wrong-call-unrepresentable reasoning `resolve_conflict` already had; the resolved `branch_id` is echoed, a published/canceled MR is refused readably. The old tests passed shifted positional args straight into MagicMock — rewritten to pin the wiring.
- **Every enriched return carries `allowed_actions`** (list rows, find, create, update, transitions, merge's post-merge state) — a `--json` consumer answers "what can I do next" without a second call.
- **Safety fact recorded**: `autoMergeStrategy=immediately` makes a background backend tick merge any *approved* MR through the same `MergeProcessor` under a system token (`AutoMergeCandidateRepository.php:38-47`, `AutoMergeTickHandler.php:86`) — `create` + `request-review` can end in a production merge with `merge()` never called. New notes section + both docstrings.
- **Docs**: full `MergeRequestResponse` wire shape in the notes table (`merge{}` is nested, `createdAt` top-level, `autoMerge*` are response fields); `MergeRequestVoter` cited (scoped token → 403 on detail/conflicts, a different axis than role whitelisting) + 403 added to the two read rows in the layer3 doc.
- **Empty `list` now carries `feature_enabled`** — 200 + `[]` on a project without the feature stops being indistinguishable from a genuinely empty project (the extra `verify_token` GET is spent only on the empty case).
- **`STATE_FILTER_VOCABULARY` and `TAKE_MODES` are public** so Layer 1 enumerates them in help text and pre-validates to exit 2 (precedent: `notification_service.KNOWN_EVENTS`).

## Known CI caveats

- Rebased onto current main (0.92.0) — `make changelog-check` is green again. No version bump here: this lands in the DMD-1899/1900 stack and the bump PR owns the changelog entry (the #616 precedent). **Note for the release PR:** the entry must also cover `details.api_error_params` (+ `api_error_params_truncated`) — a cross-cutting, bounded addition to the `--json` error envelope on every 4xx/5xx from all clients — not just the merge-request command group.
- `tests/test_changelog_render.py` fails (2 tests) whenever `FORCE_COLOR` is set in the environment (Warp exports `FORCE_COLOR=3`): Rich then emits ANSI inside the asserted substrings (`New: ` renders bold, splitting `"New: alpha thing."`). Pre-existing main-branch test fragility, fully reproducible and unrelated to this PR; worth a tiny follow-up fix on main.

## Review pointers

- The derivation tables are 1:1 with the RFC section "Derived status" and DMD-1988 — that's the contract; if a table looks wrong here, it's wrong there first.
- `resolve_conflict`'s conflict-set check, the MR-derived branch, and `onto_version = theirs.version` are the service-level guards Layer 3 deliberately does not have.

### DIFF
--- a/src/keboola_agent_cli/http_base.py
+++ b/src/keboola_agent_cli/http_base.py
@@ -454,6 +454,14 @@ class BaseHttpClient:
                 api_error_code = body.get("code")
                 if isinstance(api_error_code, str) and api_error_code:
                     details["api_error_code"] = api_error_code
+                # A Package HttpException additionally serializes its context
+                # as `params` (ExceptionConverter) -- e.g. the merge-conflict
+                # 409 carries the conflicting configurations in
+                # `params.errors`. Surface it so a caller does not have to
+                # re-fetch data the error already delivered.
+                api_error_params = body.get("params")
+                if isinstance(api_error_params, dict) and api_error_params:
+                    details["api_error_params"] = api_error_params
             # Real Keboola APIs answer with one of these keys in priority
             # order. Two caveats:
             #   1. Keboola Metastore puts the HTTP status code into `error`
--- a/src/keboola_agent_cli/http_base.py
+++ b/src/keboola_agent_cli/http_base.py
@@ -21,6 +21,8 @@ from .constants import (
     BACKOFF_BASE,
     ENV_CONVERSATION_ID,
     MAX_API_ERROR_LENGTH,
+    MAX_API_ERROR_PARAMS_LENGTH,
+    MAX_API_ERROR_PARAMS_LIST_ITEMS,
     MAX_EXCEPTION_ID_LENGTH,
     MAX_RETRIES,
     MAX_RETRY_AFTER_SECONDS,
@@ -38,6 +40,32 @@ logger = logging.getLogger(__name__)
 # extra log line (CWE-117), control characters -- is dropped before the id is
 # interpolated into an error message.
 _EXCEPTION_ID_DISALLOWED = re.compile(r"[^A-Za-z0-9._:-]+")
+
+
+def _bound_error_params(params: dict) -> tuple[dict | None, bool]:
+    """Cap the server-supplied ``params`` context like the message is capped.
+
+    Returns ``(bounded_params_or_None, truncated)``. Small payloads pass
+    through untouched. Over MAX_API_ERROR_PARAMS_LENGTH (serialized), every
+    top-level list is truncated to MAX_API_ERROR_PARAMS_LIST_ITEMS entries
+    (a merge 409's ``params.errors`` keeps its first N conflicting configs);
+    if the result still exceeds the cap, params is dropped entirely and only
+    the truncation marker survives.
+    """
+    try:
+        if len(json.dumps(params, default=str)) <= MAX_API_ERROR_PARAMS_LENGTH:
+            return params, False
+        slimmed = {
+            key: value[:MAX_API_ERROR_PARAMS_LIST_ITEMS] if isinstance(value, list) else value
+            for key, value in params.items()
+        }
+        if len(json.dumps(slimmed, default=str)) <= MAX_API_ERROR_PARAMS_LENGTH:
+            return slimmed, True
+        return None, True
+    except (TypeError, ValueError):
+        # Unserializable server payload -- drop it rather than crash the
+        # error path itself.
+        return None, True
 
 
 def build_user_agent() -> str:
@@ -458,10 +486,16 @@ class BaseHttpClient:
                 # as `params` (ExceptionConverter) -- e.g. the merge-conflict
                 # 409 carries the conflicting configurations in
                 # `params.errors`. Surface it so a caller does not have to
-                # re-fetch data the error already delivered.
+                # re-fetch data the error already delivered -- BOUNDED, like
+                # the message below: this rides every 4xx/5xx from every
+                # BaseHttpClient subclass into agent-consumed --json output.
                 api_error_params = body.get("params")
                 if isinstance(api_error_params, dict) and api_error_params:
-                    details["api_error_params"] = api_error_params
+                    bounded, truncated = _bound_error_params(api_error_params)
+                    if bounded is not None:
+                        details["api_error_params"] = bounded
+                    if truncated:
+                        details["api_error_params_truncated"] = True
             # Real Keboola APIs answer with one of these keys in priority
             # order. Two caveats:
             #   1. Keboola Metastore puts the HTTP status code into `error`
--- a/src/keboola_agent_cli/http_base.py
+++ b/src/keboola_agent_cli/http_base.py
@@ -437,6 +437,7 @@ class BaseHttpClient:
         url_label = base_url or self._base_url
 
         exception_id = ""
+        details: dict = {}
         try:
             body = response.json()
             # Keboola answers a 5xx with a generic `error` ("Application
@@ -445,6 +446,14 @@ class BaseHttpClient:
             # left the operator with nothing to escalate (issue #599).
             if isinstance(body, dict):
                 exception_id = self._safe_exception_id(body.get("exceptionId"))
+                # Keboola user errors also carry a machine-readable string
+                # `code` (e.g. `storage.mergeRequests.notReadyToMerge`).
+                # Surface it in details so a service can branch on it -- the
+                # message alone holds only the human `error` text (DMD-1899;
+                # the merge 409's two shapes differ exactly by this field).
+                api_error_code = body.get("code")
+                if isinstance(api_error_code, str) and api_error_code:
+                    details["api_error_code"] = api_error_code
             # Real Keboola APIs answer with one of these keys in priority
             # order. Two caveats:
             #   1. Keboola Metastore puts the HTTP status code into `error`
@@ -512,6 +521,7 @@ class BaseHttpClient:
                 status_code=status,
                 error_code=ErrorCode.AUTH_REJECTED,
                 retryable=False,
+                details=details,
             )
 
         if status == 403:
@@ -520,6 +530,7 @@ class BaseHttpClient:
                 status_code=status,
                 error_code=ErrorCode.ACCESS_DENIED,
                 retryable=False,
+                details=details,
             )
 
         if status == 404:
@@ -528,6 +539,7 @@ class BaseHttpClient:
                 status_code=status,
                 error_code=ErrorCode.NOT_FOUND,
                 retryable=False,
+                details=details,
             )
 
         # Appended AFTER the truncation above so they always survive into the
@@ -547,4 +559,5 @@ class BaseHttpClient:
             status_code=status,
             error_code=ErrorCode.API_ERROR,
             retryable=status in RETRYABLE_STATUS_CODES if retryable is None else retryable,
+            details=details,
         )
--- a/src/keboola_agent_cli/http_base.py
+++ b/src/keboola_agent_cli/http_base.py
@@ -550,6 +550,7 @@ class BaseHttpClient:
                     status_code=status,
                     error_code=ErrorCode.INVALID_TOKEN,
                     retryable=False,
+                    details=details,
                 )
             raise KeboolaApiError(
                 message=(
--- a/src/keboola_agent_cli/http_base.py
+++ b/src/keboola_agent_cli/http_base.py
@@ -489,6 +489,9 @@ class BaseHttpClient:
                 # re-fetch data the error already delivered -- BOUNDED, like
                 # the message below: this rides every 4xx/5xx from every
                 # BaseHttpClient subclass into agent-consumed --json output.
+                # Passed through UNMASKED by decision: it is server-authored
+                # error context (ids, names, version identifiers), not
+                # credentials -- the same trust the error message text gets.
                 api_error_params = body.get("params")
                 if isinstance(api_error_params, dict) and api_error_params:
                     bounded, truncated = _bound_error_params(api_error_params)
--- a/src/keboola_agent_cli/services/base.py
+++ b/src/keboola_agent_cli/services/base.py
@@ -122,14 +122,20 @@ def find_default_branch_id(branches: list[dict[str, Any]]) -> int | None:
     """The id of the ``isDefault`` branch in a ``list_dev_branches()`` result.
 
     The one home for the ``isDefault`` scan shared by the config, sync,
-    workspace and merge-request services. Returns ``None`` when no branch is
-    flagged (or the flagged entry carries no id) -- what that means (error
-    vs. fallback) stays the caller's decision. ``lib.py`` keeps its own loop
-    deliberately: the SDK facade does not import the services layer.
+    workspace and merge-request services. Returns ``None`` when no flagged
+    branch carries a usable (int-coercible) id -- an entry that cannot be
+    coerced is skipped, never raised on, so a degenerate payload cannot
+    crash a caller mid-operation. What ``None`` means (error vs. fallback)
+    stays the caller's decision. ``lib.py`` keeps its own loop deliberately:
+    the SDK facade does not import the services layer.
     """
     for branch in branches:
-        if branch.get("isDefault") and branch.get("id") is not None:
+        if not branch.get("isDefault"):
+            continue
+        try:
             return int(branch["id"])
+        except (KeyError, TypeError, ValueError):
+            continue
     return None
 
 
--- a/src/keboola_agent_cli/services/base.py
+++ b/src/keboola_agent_cli/services/base.py
@@ -118,6 +118,21 @@ def project_error_entry(
     }
 
 
+def find_default_branch_id(branches: list[dict[str, Any]]) -> int | None:
+    """The id of the ``isDefault`` branch in a ``list_dev_branches()`` result.
+
+    One home for the ``isDefault`` scan previously copy-pasted across
+    services (config, sync, workspace, merge-request). Returns ``None`` when
+    no branch is flagged -- what that means (error vs. fallback) stays the
+    caller's decision. ``lib.py`` keeps its own loop deliberately: the SDK
+    facade does not import the services layer.
+    """
+    for branch in branches:
+        if branch.get("isDefault"):
+            return int(branch["id"])
+    return None
+
+
 def default_client_factory(stack_url: str, token: str) -> KeboolaClient:
     """Create a KeboolaClient with the given stack URL and token.
 
--- a/src/keboola_agent_cli/services/base.py
+++ b/src/keboola_agent_cli/services/base.py
@@ -121,14 +121,14 @@ def project_error_entry(
 def find_default_branch_id(branches: list[dict[str, Any]]) -> int | None:
     """The id of the ``isDefault`` branch in a ``list_dev_branches()`` result.
 
-    One home for the ``isDefault`` scan previously copy-pasted across
-    services (config, sync, workspace, merge-request). Returns ``None`` when
-    no branch is flagged -- what that means (error vs. fallback) stays the
-    caller's decision. ``lib.py`` keeps its own loop deliberately: the SDK
-    facade does not import the services layer.
+    The one home for the ``isDefault`` scan shared by the config, sync,
+    workspace and merge-request services. Returns ``None`` when no branch is
+    flagged (or the flagged entry carries no id) -- what that means (error
+    vs. fallback) stays the caller's decision. ``lib.py`` keeps its own loop
+    deliberately: the SDK facade does not import the services layer.
     """
     for branch in branches:
-        if branch.get("isDefault"):
+        if branch.get("isDefault") and branch.get("id") is not None:
             return int(branch["id"])
     return None
 
--- a/src/keboola_agent_cli/services/merge_request_service.py
+++ b/src/keboola_agent_cli/services/merge_request_service.py
@@ -75,9 +75,14 @@ _ALLOWED_ACTIONS_BY_STATE: dict[str, tuple[str, ...]] = {
 
 # Everything `list_merge_requests`' --state filter accepts: the derived
 # vocabulary plus the raw lifecycle states (the two overlap on purpose).
-_STATE_FILTER_VOCABULARY: frozenset[str] = (
+# Public: Layer 1 enumerates it in --state help text and pre-validates to
+# exit 2 (house precedent: notification_service's KNOWN_EVENTS).
+STATE_FILTER_VOCABULARY: frozenset[str] = (
     frozenset(_DERIVED_STATE_BY_RAW) | frozenset(_DERIVED_STATE_BY_RAW.values()) | {"rejected"}
 )
+
+# The resolve_conflict take modes. Public for the same reason.
+TAKE_MODES: tuple[str, ...] = ("ours", "theirs", "delete")
 
 
 def _same_id(a: Any, b: Any) -> bool:
@@ -211,10 +216,13 @@ def derive_viewer(mr: dict[str, Any], admin_id: int | None) -> dict[str, bool | 
     """Derive the caller-relative flags: am I the creator, did I approve.
 
     ``admin_id`` is the caller's user id from ``verify_token`` (the response's
-    ``admin`` block); a scoped token has none, in which case both flags are
-    ``None`` -- honest "unknown", not ``False``. Server-first (``viewer``
-    with ``isCreator``/``hasApproved``, DMD-1988). These flags are what turns
-    a blocker into a next step: ``approvals`` + ``has_approved=True`` means
+    ``admin`` block); when it is ``None`` both flags are ``None`` -- honest
+    "unknown", not ``False``. In practice that branch is defense in depth
+    rather than a reachable path: a scoped token (no admin identity) is
+    denied by ``MergeRequestVoter`` on the detail/conflicts endpoints before
+    this function ever runs. Server-first (``viewer`` with
+    ``isCreator``/``hasApproved``, DMD-1988). These flags are what turns a
+    blocker into a next step: ``approvals`` + ``has_approved=True`` means
     "wait for the other reviewers", not "approve it".
     """
     server = _server_viewer(mr)
@@ -234,9 +242,16 @@ def derive_viewer(mr: dict[str, Any], admin_id: int | None) -> dict[str, bool | 
 
 
 def _enrich_row(mr: dict[str, Any]) -> dict[str, Any]:
-    """List-row enrichment: raw MR + derived_state (conflicts are not fetched
-    per row, so no blockers here -- detail-level enrichment does that)."""
-    return {**mr, "derived_state": derive_state(mr)}
+    """Row-level enrichment applied to every MR the service returns: raw MR +
+    derived_state + allowed_actions (both are free -- state-only). Conflicts
+    are not fetched here, so no blockers -- detail-level enrichment does that.
+    A --json consumer of create/update/transitions can answer "what can I do
+    next" without a second call (findings doc, DMD-1900)."""
+    return {
+        **mr,
+        "derived_state": derive_state(mr),
+        "allowed_actions": derive_allowed_actions(mr),
+    }
 
 
 class MergeRequestService(BaseService):
@@ -303,15 +318,23 @@ class MergeRequestService(BaseService):
         closed and known here, and a typo returning a silent ``count: 0``
         would read as "no MRs".
         """
-        if state is not None and state.lower() not in _STATE_FILTER_VOCABULARY:
+        if state is not None and state.lower() not in STATE_FILTER_VOCABULARY:
             raise ConfigError(
                 f"Unknown --state value {state!r}. Accepted values: "
-                f"{', '.join(sorted(_STATE_FILTER_VOCABULARY))}."
+                f"{', '.join(sorted(STATE_FILTER_VOCABULARY))}."
             )
         project = self._project(alias)
         client = self._client_factory(project.stack_url, project.token)
         try:
             rows = [_enrich_row(mr) for mr in client.merge_requests.list()]
+            # The list endpoint is ungated, so a project without the feature
+            # answers 200 + [] -- indistinguishable from a genuinely empty
+            # project, while every subsequent write would fail. Disambiguate
+            # only for the empty result (the has_feature call costs one
+            # verify_token GET, so it is not spent on non-empty lists).
+            feature_enabled = (
+                client.has_feature(BRANCHES_MERGE_REQUESTS_FEATURE) if not rows else None
+            )
         finally:
             client.close()
 
@@ -328,6 +351,8 @@ class MergeRequestService(BaseService):
             "count": len(rows),
             "merge_requests": rows,
         }
+        if feature_enabled is not None:
+            result["feature_enabled"] = feature_enabled
         if state is not None:
             result["state_filter"] = state
         return result
@@ -435,6 +460,13 @@ class MergeRequestService(BaseService):
         readable error) and the output must state which branch the MR was
         created from. A source branch can have at most one MR, ever; a
         second create answers 404 server-side.
+
+        ``auto_merge_strategy="immediately"`` is NOT metadata: a background
+        backend tick merges any *approved* MR carrying it (same
+        MergeProcessor as the merge endpoint, under a system token) -- and on
+        non-SOX defaults ``request_review`` lands straight in approved, so
+        create+submit alone can end in a production merge and the source
+        branch's deletion. See the notes doc, *Auto-merge*.
         """
         project = self._project(alias)
         client = self._client_factory(project.stack_url, project.token)
@@ -480,7 +512,12 @@ class MergeRequestService(BaseService):
     ) -> dict[str, Any]:
         """Update an MR's metadata. ``None`` = leave unchanged (the API cannot
         clear a field to null; an empty string clears description/externalId
-        server-side)."""
+        server-side).
+
+        ``auto_merge_strategy="immediately"`` on an already-approved MR is
+        enough for the backend's auto-merge tick to merge it -- no ``merge()``
+        call involved (see the notes doc, *Auto-merge*). Not just metadata.
+        """
         project = self._project(alias)
         client = self._client_factory(project.stack_url, project.token)
         try:
@@ -657,6 +694,7 @@ class MergeRequestService(BaseService):
         if mr_after.get("state"):
             result["state"] = mr_after["state"]
             result["derived_state"] = derive_state(mr_after)
+            result["allowed_actions"] = derive_allowed_actions(mr_after)
         if mapping_cleanup:
             result["mapping_cleanup"] = mapping_cleanup
         if cleanup_warnings:
@@ -740,9 +778,16 @@ class MergeRequestService(BaseService):
         }
 
     def get_config_diff(
-        self, alias: str, component_id: str, config_id: str, branch_id: int
+        self, alias: str, merge_request_id: int, component_id: str, config_id: str
     ) -> dict[str, Any]:
         """Three-way diff of one config, flattened to a per-path classification.
+
+        The branch is derived from the merge request (``branches.branchFromId``)
+        for the same reason ``resolve_conflict`` derives it: a caller-supplied
+        branch id has no relation to the MR, so the diff could silently show an
+        unrelated branch (findings doc, DMD-1899/DMD-1900). One GET buys the
+        symmetry; the resolved ``branch_id`` is echoed in the result so Layer 1
+        never re-derives the branch->MR relation itself.
 
         No three panes: each touched path is tagged ``changed_by`` --
         ``ours`` (only the dev branch changed it), ``theirs`` (only
@@ -761,6 +806,7 @@ class MergeRequestService(BaseService):
         project = self._project(alias)
         client = self._client_factory(project.stack_url, project.token)
         try:
+            branch_id = self._branch_from_id_of(client, merge_request_id)
             diff = client.get_config_diff(component_id, config_id, branch_id)
         finally:
             client.close()
@@ -768,6 +814,7 @@ class MergeRequestService(BaseService):
         ours = diff.get("ours")
         return {
             "alias": alias,
+            "merge_request_id": merge_request_id,
             "component_id": component_id,
             "config_id": config_id,
             "branch_id": branch_id,
@@ -874,8 +921,8 @@ class MergeRequestService(BaseService):
         """
         if (take is None) == (resolved is None):
             raise ConfigError("Pass exactly one of take=ours|theirs|delete or a resolved body.")
-        if take is not None and take not in ("ours", "theirs", "delete"):
-            raise ConfigError(f"Unknown take mode {take!r}: use ours, theirs or delete.")
+        if take is not None and take not in TAKE_MODES:
+            raise ConfigError(f"Unknown take mode {take!r}: use {', '.join(TAKE_MODES)}.")
 
         project = self._project(alias)
         client = self._client_factory(project.stack_url, project.token)
--- a/src/keboola_agent_cli/services/merge_request_service.py
+++ b/src/keboola_agent_cli/services/merge_request_service.py
@@ -43,19 +43,31 @@ _DERIVED_STATE_BY_RAW: dict[str, str] = {
     "canceled": "closed",
 }
 
-# Mechanical action availability per raw state -- what the UI buttons gate on
-# (approve/request-changes only in in_review|approved, send-for-review only in
-# development; editing/rebasing is allowed in development, in_review and
-# approved -- the dev branch is locked only while in_merge). `merge` appears
-# for `development` because a non-SOX merge from there succeeds when approvals
-# suffice (the backend auto-applies skip_review). Deliberately state-only:
-# roles and features are the pre-flight's job client-side, and the backend can
-# honor them once it serializes `allowedActions` (DMD-1988).
+# Mechanical action availability per raw state, verified against the Symfony
+# workflow (MergeRequestLifecycleStateMachine + guards; Opus wire review
+# 2026-08-27):
+# - `approve` exists ONLY in in_review (the transition's sole `from` place);
+#   from `approved` the backend answers 422. Even in in_review it is further
+#   gated by AddApprovalGuard (not the creator, not already approved,
+#   required count not yet reached) -- which no state-only table can express.
+#   With the non-SOX default of 0 required approvals, approve is 422 in every
+#   state and in_review itself is unreachable (request-review jumps straight
+#   to approved via skip_review).
+# - `merge` appears for `development` because a non-SOX merge from there
+#   succeeds when approvals suffice (the backend auto-applies skip_review).
+# - `update` is blocked server-side only in the terminal states
+#   (published/canceled) -- an in_merge MR is still updatable.
+# - `resolve_conflicts` (diff+rebase) is allowed while the MR is open; the
+#   in_merge branch lock is enforced on the SOX path only, but rebasing
+#   mid-merge is pointless, so the table deliberately omits it there.
+# Deliberately state-only: roles and features are the pre-flight's job
+# client-side, and the backend can honor them once it serializes
+# `allowedActions` (DMD-1988).
 _ALLOWED_ACTIONS_BY_STATE: dict[str, tuple[str, ...]] = {
     "development": ("request_review", "merge", "update", "resolve_conflicts"),
     "in_review": ("approve", "request_changes", "update", "resolve_conflicts"),
-    "approved": ("approve", "request_changes", "merge", "update", "resolve_conflicts"),
-    "in_merge": (),
+    "approved": ("request_changes", "merge", "update", "resolve_conflicts"),
+    "in_merge": ("update",),
     "published": (),
     "canceled": (),
 }
@@ -92,6 +104,19 @@ def derive_state(mr: dict[str, Any]) -> str:
       the hood, so a self-closed MR keeps ``state=development``.
     - otherwise the raw state mapped through ``_DERIVED_STATE_BY_RAW``
       (an unknown raw state passes through unchanged, defensively).
+
+    Reliability caveat (verified against Connection, Opus wire review
+    2026-08-27): ``reviewers[].status`` is populated only within a review
+    round anchored by an actual ``request_review`` event, and a non-reviewer's
+    decision (the creator's included -- the creator can never BE a reviewer)
+    is dropped whenever explicit reviewers exist. ``skip_review`` writes no
+    activity event, so in a project with the non-SOX default of 0 required
+    approvals every status is ``null`` and the ``rejected`` / self-``closed``
+    overrides never fire -- the same blind spot the UI badge has, since this
+    table is its port. The truth lives in the MR's activity log
+    (``changes_requested`` events, un-shadowed and un-anchored); serializing a
+    reliable ``derivedState`` from it is exactly what DMD-1988 asks Connection
+    to do. The fallback here stays best-effort by design.
     """
     server = mr.get("derivedState")
     if isinstance(server, str) and server:
@@ -334,7 +359,11 @@ class MergeRequestService(BaseService):
             if (mr.get("state") or "") in self._OPEN_STATES:
                 conflicts = client.merge_requests.conflicts(merge_request_id)
             # The verify_token call exists only to anchor the viewer
-            # polyfill (its admin block is the caller's identity); once the
+            # polyfill (its admin block is the caller's identity). Note the
+            # detail and conflicts endpoints themselves require an admin
+            # token (MergeRequestVoter denies a token with no admin) -- a
+            # scoped token fails above before viewer is ever derived; the
+            # None-flags path stays as defense in depth. Once the
             # server serializes `viewer` (DMD-1988), derive_viewer never
             # reads admin_id -- so skip the call and its cost with it. A
             # scoped token has no admin identity -> flags are None, not
@@ -507,6 +536,9 @@ class MergeRequestService(BaseService):
     # -- Merge ------------------------------------------------------------------
 
     _NOT_READY_CODE = "storage.mergeRequests.notReadyToMerge"
+    # MergeValidationException's string code (ExceptionConverter serializes
+    # it top-level as `code`, with the conflicting configs in `params.errors`).
+    _CONFLICT_CODE = "storage.mergeRequests.validation"
 
     def merge(self, alias: str, merge_request_id: int) -> dict[str, Any]:
         """Merge the MR into the default branch and clean local references.
@@ -517,13 +549,18 @@ class MergeRequestService(BaseService):
         when approvals are satisfied -- the backend auto-applies skip_review.
 
         The merge 409 is remapped onto its two wire shapes (the RFC's
-        decision; docs/error-codes.md):
+        decision; docs/error-codes.md). Both carry a machine string code
+        (ExceptionConverter serializes it top-level as ``code``):
 
         - ``storage.mergeRequests.notReadyToMerge`` (project merge lock /
           wrong state / another MR processing) -> ``MR_NOT_READY_TO_MERGE``,
           retryable -- all three causes are transient.
-        - a 409 without the string code is the conflict validation ->
-          ``MR_MERGE_CONFLICT``, not retryable, next step in the message.
+        - ``storage.mergeRequests.validation`` is the conflict validation ->
+          ``MR_MERGE_CONFLICT``, not retryable; the conflicting
+          configurations arrive in the 409's own ``params.errors`` and are
+          passed through in details. A code-less 409 is treated as a
+          conflict too (older stacks), but a 409 with any OTHER code passes
+          through unmapped -- never mislabeled as a conflict.
 
         A successful merge always also deletes the source branch -- as a
         second async job with no handle, so the result says the branch "is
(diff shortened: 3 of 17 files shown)


src/keboola_agent_cli/services/merge_request_service.py
```diff
@@ -0,0 +1,993 @@
...
+            # token (MergeRequestVoter denies a token with no admin) -- a
+            # scoped token fails above before viewer is ever derived; the
+            # None-flags path stays as defense in depth. Once the
+            # server serializes `viewer` (DMD-1988), derive_viewer never
+            # reads admin_id -- so skip the call and its cost with it. A
+            # scoped token has no admin identity -> flags are None, not
+            # False.
+            admin_id: int | None = None
+            if not isinstance(mr.get("viewer"), dict):
+                admin_id = client.verify_token().admin_id
```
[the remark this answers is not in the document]
martinsifra: Valid catch — the skip predicate was looser than `derive_viewer`'s server-field predicate, so a bare `viewer: {}` (or one with foreign keys) would skip `verify_token` and yield `{None, None}` instead of the local derivation. Fixed in cbdf1d1: both sites now share a single `_server_viewer()` predicate (dict AND has `isCreator`/`hasApproved`), so they cannot disagree; regression test added (`test_empty_server_viewer_falls_back_to_local_derivation`).

src/keboola_agent_cli/services/merge_request_service.py
```diff
@@ -0,0 +1,1064 @@
...
+            if body is None:
+                configuration = client.rebase_config_delete(
+                    component_id, config_id, branch_id, version=onto_version
+                )
+            else:
+                # `name` must also be non-empty: the diff envelope declares
+                # it nullable, but the rebase validator requires a non-empty
+                # trimmed string -- a null would sail through a bare presence
+                # check straight into a server 400.
+                missing = [key for key in ("name", "rows", "configuration") if key not in body]
```
zajca: Reviewed by Opus.

The guard exists precisely because rebase REPLACES rather than patches ("a missing key would silently wipe data, so it is refused instead of defaulted"), and the code already recognises that a present-but-null value defeats a bare presence check -- that is exactly why the `name` special case on the next line was added.

`configuration` has the same hole and a worse outcome. `missing` only tests `key not in body`, so a body carrying `"configuration": None` passes, and `rebase_config(..., configuration=body["configuration"])` posts `{"configuration": null}`. Per this PR's own wire-truth table (docs/merge-requests-notes.md: `RebaseRequest::mapValidatedData` uses `?? new stdClass()`), PHP's null-coalescing operator fires on null as well as on an absent key, so the backend substitutes `{}` -- the configuration body is replaced with an empty object and the rebase returns 200. No error, no warning.

**Failure scenario:** a caller invokes `resolve_conflict(alias, mr_id, component_id, config_id, resolved={"name": "My config", "rows": [], "configuration": None})`. `missing` is empty (all three keys present), the non-empty check applies only to `name`, so `rebase_config` POSTs `{"version": N, "diff": {"name": "My config", "rows": [], "configuration": null}}`. The backend's `?? new stdClass()` fires on the null and REPLACES the dev-branch configuration body with `{}`. The call returns 200 with `resolution: "custom"` -- the entire configuration payload is destroyed with no error surfaced.

Reachable two ways:
- The caller-authored path is fully under caller control. `resolve_conflict(resolved={...})` is the public service entry point that Layer 1 (DMD-1900) will feed from a user-supplied JSON/YAML resolution file; a file with a null `configuration` is refused nowhere.
- The `take=ours|theirs` path composes `body` from the server's `diff` envelope. The comment above acknowledges that envelope declares `name` nullable despite being required; if `configuration` is nullable on the same schema, a null there is indistinguishable from a legitimate one here.

Suggested fix: extend the non-null check beyond `name` -- treat a `None` value for `configuration` (and `rows`, whose `[]`-deletes-all-rows semantics make a null equally ambiguous) as missing, so both take the existing refusal path rather than reaching `rebase_config`.

The existing regression test `test_null_name_on_a_take_side_is_a_contract_violation` covers only `name`; no test pins the null-`configuration` case.

src/keboola_agent_cli/services/merge_request_service.py
```diff
@@ -0,0 +1,1064 @@
...
+        """
+        project = self._project(alias)
+        client = self._client_factory(project.stack_url, project.token)
+        try:
+            for mr in client.merge_requests.list():
+                if _same_id((mr.get("branches") or {}).get("branchFromId"), branch_id):
+                    return {"alias": alias, **_enrich_row(mr)}
+        finally:
+            client.close()
+        raise KeboolaApiError(
```
zajca, line 409: Reviewed by Opus.

This PR added `feature_enabled` to `list_merge_requests` for exactly this ambiguity, with the rationale spelled out in the comment at line 330: "The list endpoint is ungated, so a project without the feature answers 200 + [] -- indistinguishable from a genuinely empty project, while every subsequent write would fail."

`find_merge_request_for_branch` reads the same ungated `client.merge_requests.list()` and did not get the same treatment. On a project without `branches-merge-requests` the loop sees an empty list and raises NOT_FOUND whose message asserts a fact it cannot know ("Branch N has no merge request") and prescribes a next step that is guaranteed to fail -- `create_merge_request` runs `_require_merge_requests_feature` first and refuses with FEATURE_NOT_ENABLED.

**Failure scenario:** on a project where `branches-merge-requests` is not enabled, `find_merge_request_for_branch(alias, 123)` gets `200 + []` from the ungated list endpoint, matches nothing, and raises NOT_FOUND with "Branch 123 has no merge request in project 'prod'. Create one with `kbagent merge-request create`." The user runs that command and gets FEATURE_NOT_ENABLED ("Merge requests are not enabled on this project") -- the first error's diagnosis and its prescribed next step were both wrong, and the actual cause was knowable from the already-loaded features cache.

This matters more than for `list`: the docstring identifies this method as "the resolver behind L1's optional --mr-id", so it is on the path of every merge-request command invoked without an explicit id. The user is sent in a loop -- told to create an MR, then told merge requests are not enabled.

Suggested fix: mirror the `list_merge_requests` treatment -- on the no-match path only (so the cost is not paid on the happy path), call `client.has_feature(BRANCHES_MERGE_REQUESTS_FEATURE)` before raising, and either raise FeatureNotEnabledError or qualify the message when the feature is absent. The `client.close()` in the `finally` currently runs before the raise, so the check needs to move inside the `try`.

src/keboola_agent_cli/http_base.py
```diff
@@ -417,6 +418,22 @@ def _raise_api_error(
...
+                if isinstance(api_error_code, str) and api_error_code:
+                    details["api_error_code"] = api_error_code
+                # A Package HttpException additionally serializes its context
+                # as `params` (ExceptionConverter) -- e.g. the merge-conflict
+                # 409 carries the conflicting configurations in
+                # `params.errors`. Surface it so a caller does not have to
+                # re-fetch data the error already delivered.
+                api_error_params = body.get("params")
+                if isinstance(api_error_params, dict) and api_error_params:
+                    details["api_error_params"] = api_error_params
```
zajca: Reviewed by Opus.

`api_error_params` copies the response body's `params` object verbatim into `KeboolaApiError.details`, with no size bound, at all four raise sites (401/403/404 and the generic path) -- so it now rides along on every 4xx/5xx from every BaseHttpClient subclass (KeboolaClient, AiServiceClient, DataScienceClient, AuthClient via its `_raise_api_error` override), not just the merge 409 this was added for. Numerous command-layer sites re-propagate it with `details=exc.details` (commands/data_app.py, commands/_data_app_runtime.py, commands/_data_app_git.py, commands/_semantic_layer_helpers.py), and output.py emits `details` into the --json error envelope whenever non-empty.

The asymmetry is the concern: `api_message` twenty lines below is deliberately capped at MAX_API_ERROR_LENGTH, and the comment at line 495 establishes bounding server-supplied text as the rule for this function. `params` gets no equivalent. The merge-conflict case is the motivating example and also the worst one -- `params.errors` carries one entry per conflicting configuration, so a branch with hundreds of conflicts produces a proportionally large blob. In a CLI whose --json output is consumed by agents, that lands unbounded in a context window.

**Failure scenario:** a merge over a branch with several hundred conflicting configurations answers 409 with `params.errors` listing every one. `_raise_api_error` copies the whole object into `details["api_error_params"]`, `MergeRequestService._remap_merge_conflict` forwards `details=exc.details` into the MR_MERGE_CONFLICT error, and the --json error envelope emits the entire list verbatim -- while the human error message beside it was truncated to MAX_API_ERROR_LENGTH.

Suggested fix: bound what is copied (cap the serialized size, or cap list lengths inside `params`) so the details payload has a ceiling like the message does. Alternatively, narrow the passthrough to the specific keys a caller is documented to branch on rather than copying the whole object.

docs/merge-requests-layer1.md
```diff
@@ -0,0 +1,342 @@
...
+  nothing**. A Typer repeatable option that yields `()` instead of `None` would therefore
+  silently clear all reviewers on any `update` that did not mention them. Normalise with an
+  explicit `or None` at the call site.
+- **`update` semantics:** `None` = leave unchanged; an **empty string clears**
+  `description` / `externalId` (server-side `?? null` mapping). There is no clear-to-null for
+  anything else. Passing `--reviewer-id` at all **replaces** the set — it never appends.
+- `--auto-merge-strategy` accepts exactly `immediately` | `scheduled` | `none`.
+- `--external-id` is capped at 255 characters server-side.
+- **`--state` stays a plain `str`, not a Typer enum.** The vocabulary lives in the service
+  (`_STATE_FILTER_VOCABULARY`) and an unknown value already fails with the accepted list. A
```
zajca: Reviewed by Sonnet.

The Layer 1 RFC says the state-filter vocabulary "lives in the service (`_STATE_FILTER_VOCABULARY`)", implying an underscore-prefixed private name. The constant this same PR actually defines in `src/keboola_agent_cli/services/merge_request_service.py:80` is the public `STATE_FILTER_VOCABULARY` (no leading underscore) — its own comment says "Public: Layer 1 enumerates it in --state help text". A future implementer of DMD-1900 following this doc verbatim would try to import a name that doesn't exist. Low impact since it's a design doc for not-yet-built Layer 1, but worth a one-word fix to avoid confusion later.

zajca (changes requested):
Changes requested. See the published review findings.

zajca (commented):

src/keboola_agent_cli/services/merge_request_service.py
martinsifra: Confirmed and fixed in 1e239b4 — the guard now refuses present-but-null for **every** required key (`key not in body or body[key] is None`), so a `configuration: null` (which PHP's `?? new stdClass()` would turn into a silent full-body wipe) and a null `rows` both take the existing refusal path. A genuine `rows: []` (delete-all) stays legal — pinned by `test_empty_rows_list_stays_a_legal_resolution`; new regressions cover null `configuration` on the resolved path and null `rows` on a take side.

src/keboola_agent_cli/services/merge_request_service.py
martinsifra: Confirmed and fixed in 1e239b4 — `find_merge_request_for_branch` now runs `_require_merge_requests_feature` on the no-match path (inside the `try`, so the client is still open), mirroring the `list_merge_requests` treatment: a featureless project gets FEATURE_NOT_ENABLED (with the SOX/enable message split) instead of a NOT_FOUND prescribing a create that cannot succeed. The check is spent only when nothing matched — `test_find_happy_path_does_not_spend_the_feature_call` pins that the happy path never pays the extra GET.

src/keboola_agent_cli/http_base.py
martinsifra: Confirmed and fixed in 1e239b4 — `api_error_params` is now bounded like the message beside it: over `MAX_API_ERROR_PARAMS_LENGTH` (8192 serialized) every top-level list is truncated to `MAX_API_ERROR_PARAMS_LIST_ITEMS` (20) entries; if still over, params is dropped entirely. Either way `details.api_error_params_truncated: true` marks it, and an unserializable payload is dropped rather than crashing the error path. Three regression tests (small passes verbatim / oversized list truncated with the kept prefix verbatim / untruncatable dropped with marker).

docs/merge-requests-layer1.md
martinsifra: Already resolved — the Layer 1 RFC rewrite (980b28d, now pushed as part of this branch) references the public `STATE_FILTER_VOCABULARY` and `TAKE_MODES` names (docs/merge-requests-layer1.md:86: "importing the now-public STATE_FILTER_VOCABULARY and TAKE_MODES from the service"); the underscore-prefixed mention this comment anchored to no longer exists in the current revision.

zajca (changes requested):
Reviewed the full diff on the branch, ran the complete test suite and every quality gate locally.

**Gate status (on the PR branch):** `pytest tests/` 6527 passed / 183 skipped, `ruff check` + `ruff format --check` clean, `ty check` clean on the new modules, `make loc-check` / `check-sentinel-guards` / `changelog-check` all OK. I also reproduced the `FORCE_COLOR` caveat — `FORCE_COLOR=3 pytest tests/test_changelog_render.py` does fail 2 tests, pre-existing on main, unrelated to this PR.

The design work here is genuinely good — see "What works well" at the bottom. One finding blocks.

---

## BLOCKING

### 1. `resolve_conflict` guards 3 of the 5 replace keys — `isDisabled` / `description` silently lose data

`src/keboola_agent_cli/services/merge_request_service.py:1023-1024`

```python
is_disabled=bool(body.get("isDisabled", False)),
description=body.get("description"),
```

Layer 3 warns about exactly this in `rebase_config`'s own docstring (`client/configs.py`):

> `is_disabled`: **Required**: on a missing key the backend substitutes `false`, re-enabling a config that was disabled.
> `description`: … It has **no default precisely because that default would silently drop an existing description**.

Layer 2 reintroduces both defaults. The guard at `:987-991` covers only `name` / `rows` / `configuration`, and its stated reasoning — "rebase REPLACES; a missing key would silently wipe data" — applies verbatim to the other two:

- a custom `resolved` body without `isDisabled` **silently re-enables a disabled configuration** in the dev branch, and the merge then pushes that into production;
- a custom body without `description` wipes the description;
- the `take=` path has the same hole whenever the envelope omits a key — and the guard on the other three exists precisely because the envelope is not trusted.

`tests/test_merge_request_service.py:892` (`test_custom_body_rebases_verbatim`) pins the permissive behaviour as expected: a body of `{name, rows, configuration}` passes and sends `is_disabled=False`.

Suggested fix — check for **presence**, not non-None (`description: None` is a legitimate resolved value, and the `_side` fixture sends exactly that):

```python
missing = [k for k in ("name", "rows", "configuration") if k not in body or body[k] is None]
missing += [k for k in ("isDisabled", "description") if k not in body]
```

plus `is_disabled=bool(body["isDisabled"])`.

---

## NON-BLOCKING

### 2. `allowed_actions` points at a write that cannot succeed

`_enrich_row` (`merge_request_service.py:244`) is state-only. On a project without the feature, `list_merge_requests` returns `feature_enabled: false` **and** rows carrying `allowed_actions: ["request_review", "merge", "update", "resolve_conflicts"]`. A `--json`/agent consumer — the stated reason the field exists — is steered straight into the guaranteed `FEATURE_NOT_ENABLED` loop that `find_merge_request_for_branch:371` deliberately closes. The features cache is already warm at that point on the empty-list path, so correcting it is free.

### 3. `api_error_params` is a cross-cutting output-contract change with no changelog line

`http_base.py:485-497` adds up to 8 KB of server-supplied `params` to `details` on **every** 4xx/5xx from **all seven** clients, and existing commands already forward `details=exc.details` into `--json` (`commands/data_app.py`, `commands/job.py:442`, `_semantic_layer_helpers.py`, …). Two consequences:

- consumers treating `details` as a presence flag (it was empty on 401/403/404 until now) start seeing content;
- `params` reaches the output with no masking at all, whereas the message at least goes through truncation. Low risk, but it deserves to be a stated decision rather than a side effect.

Deferring the changelog to the bump PR follows the #616 precedent, but that entry needs to mention this too — not just "merge-request Layer 2".

### 4. `_classify_three_way`: a null side with a non-empty base emits contradictory signals

`merge_request_service.py:846-852` — when one side is `null` while `base` exists, every base key comes out as "that side removed it" (`changed_by: "theirs"`, `theirs: None`), while `theirs_deleted` reports `None`, meaning "the side does not exist at all". A Layer 1 renderer will make nonsense of that pair. Either skip `changes` for a null side or flag it separately.

### 5. `change_description` is silently dropped for `take="delete"`

`merge_request_service.py:973-975` — `rebase_config_delete` has no such parameter. Layer 1 will expose `--change-description`, the user will pass it, and nothing reports that it was ignored. A warning in the result (or plumbing it through L3) would close it.

---

## NIT

- **Comments narrating history/process**, against the repo's own "Code Comments — never narrate the change" rule: `constants.py:539` ("Renamed from `FEATURE_BRANCHES_MERGE_REQUESTS`"), `services/base.py:124` ("previously copy-pasted"), `merge_request_service.py:47,113` ("Opus wire review 2026-08-27"), `:371,978` ("Zajca's PR review"). "Verified against `MergeRequestLifecycleStateMachine`" is a fact about the code and belongs there; dated review references and reviewer names belong in the PR description.
- `services/search_service.py:205` still uses the raw string `"FEATURE_NOT_ENABLED"`. Now that the enum exists, this is a one-line cleanup in the same spirit as the `find_default_branch_id` hoist.
- `merge()` — the `except Exception` wraps `set_project_branch` too, so a failed config write also skips the mapping cleanup. Two separate `try` blocks would keep the two cleanups independent.
- `merge()` `:660` — `int(raw_branch_from)` raises `ValueError` on a non-numeric wire string, i.e. an unhandled exception instead of a `KeboolaApiError`. `_same_id` exists precisely because the payload mixes id types.
- `find_default_branch_id` — `int(branch["id"])` raises `KeyError` for an `isDefault: true` entry with no `id`; the rest of the function is defensive (`branch.get("isDefault")`).
- `cleanup_branch_id_from_mapping(branch_from_id)` matches on the numeric id alone with no project scoping — a sync workspace of a *different* project in the CWD with the same branch id gets unlinked. Inherited from `BranchService.delete_branch`, so not a regression, but `merge` extends it to another path.

---

## What works well

- **The tests are above average**: a regression test per review finding, `MagicMock(spec=KeboolaClient)` + `spec=MergeRequests` at the L3 seam (a renamed client method fails the tests instead of leaving them falsely green), and the `_DEFAULT_SIDE` sentinel distinguishing "keep the fixture default" from "the side does not exist".
- `_server_viewer` as the **single** predicate for "did DMD-1988 land" — both `derive_viewer` and the `verify_token` skip read it, so they cannot drift apart. Right call.
- `resolve_conflict` / `get_config_diff` deriving the branch from the MR instead of taking it as a parameter is honest make-the-wrong-call-unrepresentable design.
- The 409 remap letting an unknown code pass through **unmapped** rather than guessing "conflict" is exactly right.
- The `find_default_branch_id` hoist and the `compute_diff` → `compute_diff_entries` split are clean refactors with tests; I verified the formatted output stays byte-identical.
- The layer1/2/3 + notes doc set carrying the wire-truth table is what made the nested-`diff`-envelope bug visible before production.

**Verdict: NEEDS-WORK** — finding #1 (silent `isDisabled` / `description` loss on rebase). Everything else is non-blocking.

martinsifra:
Thanks — applied in 01fb0fa. Per finding:

**BLOCKING #1 (isDisabled/description)** — confirmed and fixed. All five replaced-body keys are now required; one refinement over the suggested fix: `isDisabled` refuses absent-**or-null** (a null would still reach `bool()` as `False` — the exact silent re-enable), while `description` refuses only **absence** (an explicit null is a decision on a genuinely nullable field, an omission is not). `is_disabled` forwards verbatim. Regression tests: missing `isDisabled`, missing vs. explicit-null `description`, take-side `isDisabled: true` forwarded.

**#2 (allowed_actions on a featureless project)** — respectfully declined, with the reasoning now documented on `_enrich_row`. The stated scenario cannot occur: `feature_enabled: false` is emitted only on the *empty*-list path, where there are no rows to carry `allowed_actions`. The real residual is a project that *lost* the feature after MRs existed — closing that costs a `verify_token` GET on **every non-empty list** (the cache is warm only on the empty path), for a rare configuration where the pre-flight already answers any attempted write with the precise `FEATURE_NOT_ENABLED`. The proper fix is DMD-1988's server-side `allowedActions`, which can honor features.

**#3 (api_error_params output contract)** — agreed. The unmasked pass-through is now a stated decision in the code (server-authored error context, same trust as the message text), and the PR body's bump-PR note now says the release changelog entry must cover `details.api_error_params` (+`_truncated`) as a cross-cutting output-contract addition, not just "merge-request Layer 2".

**#4 (null side vs. base)** — confirmed and fixed: `_classify_three_way` emits no per-path rows when either side is null; the side-level fact stays on `ours_deleted`/`theirs_deleted`. Docstring + test updated.

**#5 (change_description on take=delete)** — confirmed; fixed with a `warnings[]` entry in the result. Plumbing through L3 is not possible: `changeDescription` lives *inside* the diff envelope and the delete tombstone is exactly `{"version": N, "diff": {}}` — the wire has nowhere to carry it (the server records its default message). Layer 1 will additionally reject the combination at exit 2 per its RFC.

**Nits** — all applied except the last: history-narrating comments removed (facts like "verified against MergeRequestLifecycleStateMachine" stayed, names/dates went); `search_service` now uses `ErrorCode.FEATURE_NOT_ENABLED`; `merge()`'s two cleanups run in independent try blocks (pinned by a test that a failed config write no longer skips the mapping unlink); wire branch ids coerce via a `_coerce_branch_id` helper (non-numeric degrades instead of raising); `find_default_branch_id` tolerates an `isDefault` entry without an id. The **unscoped `cleanup_branch_id_from_mapping`** is real but inherited from `BranchService.delete_branch` — fixing it means fixing both call sites together, so it's noted as a follow-up rather than diverging the two here.

Full triage record: `tasks/pr-703-zajca-review-2.md` (local). Gates on the branch: 107 service tests, ruff/format/ty clean.

src/keboola_agent_cli/services/merge_request_service.py
```diff
@@ -0,0 +1,1083 @@
...
+                    config_id,
+                    branch_id,
+                    version=onto_version,
+                    name=body["name"],
+                    rows=body["rows"],
+                    configuration=body["configuration"],
+                    is_disabled=bool(body.get("isDisabled", False)),
+                    description=body.get("description"),
+                    change_description=change_description,
+                )
```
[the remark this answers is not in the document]
martinsifra: Valid — and independently raised as the blocking finding of the human review a day later; fixed in 01fb0fa. The guard now requires all five replaced-body keys: `name`/`rows`/`configuration`/`isDisabled` refuse absent-or-null (a null `isDisabled` would still reach `bool()` as `False` — the same silent re-enable), `description` refuses absence only — an explicit `description: null` is a legitimate clear-the-description decision, exactly as this comment suggests. `is_disabled` forwards verbatim via `body["isDisabled"]`. Pinned by `test_missing_is_disabled_is_refused_not_defaulted` and `test_missing_description_is_refused_but_explicit_null_is_legal`.

src/keboola_agent_cli/services/merge_request_service.py
```diff
@@ -0,0 +1,1152 @@
...
+            self._require_merge_requests_feature(client)
+            # branchFromId is nullable once the MR is published -- capture it
+            # from the pre-merge payload, not the post-merge one. Coerced to
+            # int so the was_active comparison and the mapping cleanup below
+            # cannot be defeated by a string-serialized wire id (the payload
+            # mixes int and str ids -- the reason _same_id exists).
+            raw_branch_from = (
+                client.merge_requests.get(merge_request_id).get("branches") or {}
+            ).get("branchFromId")
+            branch_from_id = _coerce_branch_id(raw_branch_from)
```
zajca, line 682: **`_coerce_branch_id`'s degradation is unrecorded, and leaves the local active branch pointing at a deleted branch.**

A `branchFromId` that does not coerce to int becomes `None` -- which is *also* the legitimate documented state ("branchFromId is nullable once the MR is published"). Nothing distinguishes them: no log line, no `cleanup_warnings` entry.

Reproduced (MR 7 `approved`, `branches.branchFromId = "0123x"`, `active_branch_id = 123`):

```
branch_from_id    : None
was_active        : False
cleanup_warnings  : None
message           : "Merge request 7 merged into production."
mapping cleanup   : []
active branch now : 123     <- the branch the merge just deleted
```

The merge succeeds, `set_project_branch(alias, None)` is skipped (`was_active` is False), the sync-mapping entry is left stale, and the result reports an unqualified success. Every following `kbagent` call in that project then runs against a deleted branch.

Before this change the bare `int()` raised `ValueError` *before* `merge_requests.merge()` was issued -- loud and unhelpful, but nothing was left inconsistent. The direction of the fix is right; the gap is that the degradation is unrecorded. `test_non_numeric_wire_branch_id_degrades_instead_of_raising` currently pins this state as expected (`branch_from_id is None`, `called == []`).

Suggested fix: when `raw_branch_from is not None and branch_from_id is None`, log a warning and append a `cleanup_warnings` entry naming the raw value and the fact that local cleanup was skipped.

src/keboola_agent_cli/services/merge_request_service.py
```diff
@@ -0,0 +1,1152 @@
...
+                    )
+                configuration = client.rebase_config(
+                    component_id,
+                    config_id,
+                    branch_id,
+                    version=onto_version,
+                    name=body["name"],
+                    rows=body["rows"],
+                    configuration=body["configuration"],
+                    is_disabled=bool(body["isDisabled"]),
```
zajca: **`bool()` on a load-bearing boolean accepts any truthy value, so a string `"false"` silently disables the config.**

The presence guard above was tightened precisely because this value is load-bearing on a REPLACE, but the value itself is still coerced with no type check. Reproduced against `resolve_conflict(..., resolved=body)`:

```
isDisabled='false' -> is_disabled=True      isDisabled='0'  -> is_disabled=True
isDisabled='no'    -> is_disabled=True      isDisabled=''   -> is_disabled=False
```

`rebase_config` REPLACES, so a caller-authored three-way body carrying `"isDisabled": "false"` resolves the conflict by **disabling the configuration**; the MR then merges that into production and the command reports `resolution: "custom"`, exit 0. This is the mirror of the blocking finding this commit fixed -- and the more likely direction now that callers *must* supply the key: Layer 1 will feed `resolved=` from a user-authored JSON/YAML file, and kbagent's own `--set PATH=VALUE` convention produces strings. `''` -> `False` is the same hole in the other direction.

`test_take_side_forwards_is_disabled_verbatim` would not catch this (it passes a real `True`).

Worth noting that the same commit added `_coerce_branch_id` for exactly the int flavour of this hazard; the bool flavour has no equivalent.

Suggested fix: reject a non-`bool` with the same `missing`-style error and forward `body["isDisabled"]` unchanged.

src/keboola_agent_cli/services/merge_request_service.py
```diff
@@ -0,0 +1,1152 @@
...
+                    key
+                    for key in ("name", "rows", "configuration", "isDisabled")
+                    if key not in body or body[key] is None
+                ]
+                # description is the one genuinely nullable replaced field --
+                # only ABSENCE is refused (an explicit null is a legitimate
+                # resolved value; the wire treats null and absent the same,
+                # but a caller who spelled it out has decided, one who omitted
+                # it has not).
+                if "description" not in body:
```
zajca: **On the `take=` path this presence requirement refuses a wire-equivalent input, and the workaround it prescribes is self-contradicting.**

Requiring presence is right on the `resolved=` path -- it forces the caller to decide. On the `take=` path there is no caller intent to elicit: `body` is the server-produced `diff` envelope.

`client/configs.py:759-762` -- `rebase_config` itself omits the key when `description is None`:

```python
if description is not None:
    diff["description"] = description
```

documented there as "`None` omits the key rather than sending an explicit null, which costs no expressiveness because server-side the two are indistinguishable (`isset` mapping)".

So the two states this guard separates on the take path produce a **byte-identical** request, and the remedy the error names -- hand-author a body with `description: null` -- is exactly what the code just refused to do automatically.

Reproduced (`del theirs["diff"]["description"]`, then `take="theirs"`):

```
KeboolaApiError: The diff's theirs side carries no description -- cannot compose a
replace body from it. Author the resolution manually and pass it as a resolved body instead.
rebase_config called: False
```

Suggested fix: on the `take=` path treat an absent `description` as `None`; keep the hard refusal for `name` / `rows` / `configuration` / `isDisabled`, and keep all five required on the `resolved=` path.

**Do not extend the same defaulting to `isDisabled`** -- the two are not symmetric. `rebase_config` sends `isDisabled` unconditionally, and per its own docstring an absent key makes the backend substitute `false`, i.e. re-enable a disabled config. Defaulting it here would reintroduce the blocking finding from the previous round.

Two things worth closing alongside it: the comment above claims the envelope's schema marks all content keys required -- if that does not hold for `description`, every `take=ours` / `take=theirs` on a config without one fails 100% of the time, so it is worth confirming against a live stack. And no test distinguishes "only `description` is missing" from the all-keys-missing case `test_take_side_without_content_keys_blames_the_server_not_the_caller` covers.

src/keboola_agent_cli/services/merge_request_service.py
```diff
@@ -0,0 +1,1152 @@
...
+        mr_after: dict[str, Any] = results if isinstance(results, dict) else {}
+
+        message_parts = [f"Merge request {merge_request_id} merged into production."]
+        if branch_from_id is not None:
+            message_parts.append(
+                f"Source branch {branch_from_id} is being deleted (a separate async "
+                "job -- it may still briefly exist)."
+            )
+        if was_active:
+            message_parts.append("Active branch reset to main.")
```
zajca, line 738: **The message reports a reset that may not have happened.**

Splitting the cleanup into two independent `try` blocks is the right call, but `message_parts` is still built from `was_active` -- the pre-condition, not the outcome of the block above. Reproduced (`set_project_branch` raises `OSError`):

```
was_active              : True
message                 : "... Active branch reset to main."
cleanup_warnings        : ["Post-merge active-branch reset failed: config write failed"]
stored active_branch_id : 123      <- unchanged
```

The message asserts the exact opposite of the warning beside it, the config still points at the branch that is being deleted, and no recovery step is named. `test_failed_branch_reset_does_not_skip_mapping_cleanup` walks this path and asserts only the warning, so the contradiction is pinned as acceptable.

Suggested fix: gate this line on the outcome of the reset rather than on `was_active`, and name `kbagent branch reset` in the warning.

src/keboola_agent_cli/services/merge_request_service.py
```diff
@@ -0,0 +1,1152 @@
...
+        """Intersect the two pairwise diffs (base->ours, base->theirs) per path.
+
+        Only meaningful when BOTH sides exist: a null side (the config never
+        existed there) is a side-level fact carried by the ``ours_deleted`` /
+        ``theirs_deleted`` flags, and fabricating per-path rows against an
+        empty stand-in would emit contradictions -- every base key rendered
+        as "that side removed it" next to a flag saying the side does not
+        exist. So a null ours/theirs yields no ``changes`` at all.
+        """
+        if diff.get("ours") is None or diff.get("theirs") is None:
```
zajca: **The new short-circuit covers a null side but not a tombstoned one, so the contradiction it removes is still reachable.**

Reproduced with `ours={"version": 4, "isDeleted": True}` (present, no `diff` envelope):

```
ours_deleted: True
{'path': 'configuration',       'changed_by': 'ours',   'ours': None, 'theirs': {'limit': 100}}
{'path': 'configuration.limit', 'changed_by': 'theirs', 'ours': 100,  'theirs': 250}
{'path': 'name' / 'rows' / 'isDisabled' / 'description', 'changed_by': 'ours', 'ours': None}
```

The first two rows contradict each other, and `get_config_diff`'s docstring asserts "Deletions do not show up as paths: a soft-deleted side is flagged by the top-level `ours_deleted` / `theirs_deleted` booleans instead" -- not true for this input.

`resolve_conflict` already treats `side.get("isDeleted")` as "this side has no content to take" (it collapses to the delete resolution), so the classifier should skip it on the same criterion rather than on `is None`.

Suggested fix: `if not content(diff.get("ours")) or not content(diff.get("theirs")): return []` -- which also covers a `{}` side, where `ours_deleted` would read `False` while every base key rendered as removed.

src/keboola_agent_cli/services/merge_request_service.py
```diff
@@ -0,0 +1,1152 @@
...
+
+        The null check is best-effort, not airtight: `branchFromId` is nulled
+        by the FK's ON DELETE SET NULL when the source branch row is deleted,
+        and that deletion is a separate async job -- a freshly published MR
+        can still carry the id for a while. Harmless: the rebase then fails
+        server-side (the MR is no longer open), it just fails later.
+        """
+        mr = client.merge_requests.get(merge_request_id)
+        branch_from_id = _coerce_branch_id((mr.get("branches") or {}).get("branchFromId"))
+        if branch_from_id is None:
```
zajca, line 1172: **The error now asserts a cause the code did not establish.**

`_coerce_branch_id` collapses "absent" and "present but not numeric" into one `None`, but this message names only the absent cause. Reproduced with `branchFromId = "b-123"` on an `in_review` MR:

```
Merge request 7 has no source branch (state: in_review) -- a published or canceled MR cannot be resolved.
```

The message contradicts the state it prints in the same sentence, and sends the operator to check the MR lifecycle instead of the payload.

The trigger's likelihood equals the fix's own premise: if a non-numeric `branchFromId` is unreachable then `_coerce_branch_id` is dead code, and if it is reachable then this is what the operator sees.

Suggested fix: branch on the raw value -- `None` keeps this message, a present-but-uncoercible value gets its own ("`branchFromId` is `{raw!r}`, not a numeric branch id").

src/keboola_agent_cli/services/base.py
```diff
@@ -118,6 +118,21 @@ def project_error_entry(
...
+    """The id of the ``isDefault`` branch in a ``list_dev_branches()`` result.
+
+    The one home for the ``isDefault`` scan shared by the config, sync,
+    workspace and merge-request services. Returns ``None`` when no branch is
+    flagged (or the flagged entry carries no id) -- what that means (error
+    vs. fallback) stays the caller's decision. ``lib.py`` keeps its own loop
+    deliberately: the SDK facade does not import the services layer.
+    """
+    for branch in branches:
+        if branch.get("isDefault") and branch.get("id") is not None:
```
zajca: **Half-hardened: the `ValueError` this guards against is still one input away, and the helper has no test.**

Probed directly:

```
[{'isDefault': True, 'id': None}]                                -> None              (this change)
[{'isDefault': True, 'id': 'main'}]                              -> raises ValueError
[{'isDefault': True, 'id': None}, {'isDefault': True, 'id': 9}]  -> 9  (a different branch)
```

The same commit added `_coerce_branch_id` in `merge_request_service` for exactly this hazard, but this shared helper -- used by the config, sync, workspace and merge-request services -- kept the weaker policy, so the crash the change set out to remove is still reachable.

The caller that degrades silently is `sync_service.py:332-339`: `branches=[...] if default_branch_id else []`. A `None` there is not an error path -- `sync init` writes `.keboola/manifest.json` with an empty `branches` list and returns `status: "initialized"`, exit 0. (It is also a truthiness test rather than `is not None`.) `config_service.py:1487` and `workspace_service.py:235` do raise on `None`, but their message then says "no default branch found" for a project that reported one.

`grep -rn find_default_branch_id tests/` returns nothing -- neither this change nor the hoist itself is pinned by a unit test.

Suggested fix: give the helper `_coerce_branch_id`'s policy (skip what cannot be coerced, never raise), decide explicitly what `sync init` should do on `None`, and add the unit test.

zajca (changes requested):
Reviewed the changes since my last review (`565e239..4dbfc00`) -- the two commits applying that round plus the doc-reference repoint. Ran the full suite and every quality gate locally, and had two independent adversarial passes go over the same diff; every finding below was reproduced with a throwaway probe test.

**Gate status on `4dbfc00`:** `pytest tests/` 6531 passed / 61 skipped, `ruff check` + `ruff format --check` clean, `ty check` clean on the touched modules, `make loc-check` / `check-sentinel-guards` OK, `gh pr checks` 3/3 green, branch contains current `origin/main` (0.92.0). No `docs/merge-requests*` reference remains in `src/` / `tests/` / `docs/`, so the dangling-reference cleanup in `4dbfc00` is verified.

**The blocking finding from the previous round is fixed correctly** -- all five replaced keys required, `is_disabled` never defaulted, `description` refusing absence but accepting an explicit null, and a regression test for each. The other six fixes hold too, but four of them left a residual gap, and two of those matter.

## What blocks

- **`merge()`'s `_coerce_branch_id` degradation is silent** and leaves the project's active branch pointing at the branch the merge just deleted, under an unqualified success message. The previous `int()` crashed *before* the merge was issued, so nothing was left inconsistent; the fix traded a loud crash for quiet local corruption.
- **`bool(body["isDisabled"])`** maps `"false"` / `"0"` / `"no"` to `True`, so a caller-authored resolution body disables the configuration and the MR merges that into production -- exit 0, `resolution: "custom"`. This is the mirror of the finding just fixed, and now the *more* likely direction since callers are required to supply the key.

## Non-blocking, same round

- The `description` presence requirement is applied to the server-produced envelope on the `take=` path, where it separates two states that serialize to a byte-identical request -- and the workaround it prescribes is the thing it refused to do. Note the suggested fix must **not** be extended to `isDisabled`; the two are not symmetric.
- `merge()`'s message still claims "Active branch reset to main." when that write failed.
- `_classify_three_way`'s new short-circuit misses a tombstoned side, so the contradiction it was added to remove is still reachable.
- `_branch_from_id_of` diagnoses an unparseable id as "published or canceled MR", contradicting the state it prints alongside.
- `find_default_branch_id` is hardened against `id: None` but not against a non-numeric id, and has no unit test at all.

## Two notes that are not inline-anchorable

- **The `try` split achieves less than it reads.** `sync/branch_mapping.py:157-163` already swallows `FileNotFoundError` / `ConfigError` / `ValueError` from `load_branch_mapping` and returns `None`, which `merge()` cannot tell apart from "no entry referenced the branch". So the most likely mapping-cleanup failure -- an unreadable `branch-mapping.json` -- still produces no warning; only a `save_branch_mapping` `OSError` reaches the new `except`. Pre-existing and outside this diff, flagged because it bears on whether the split delivers its stated goal.
- `merge_request_service.py:130` is 110 chars against `line-length = 100`; the comment edit left that docstring line unwrapped. `E501` is in the ignore list and `ruff format` does not rewrap docstrings, so no gate catches it. (`:624` at 101 chars is pre-existing.)

## Confirmed as correct, not re-raised

- **`_enrich_row` staying feature-blind:** my "the features cache is already warm" applies only to the *empty*-list path, where there are no rows to enrich -- `list_merge_requests` spends `has_feature` only when `not rows`. The cost argument is right and the documented decision stands.
- **The strict five-key guard on the `resolved=` path** is the safe direction, and it fails loud rather than silently; `client/configs.py:702-708` backs the reasoning.
- **`api_error_params`:** bounded via constants (no inline literals), an over-large nested payload is dropped entirely, and the unmasked passthrough is now a stated decision. The release-PR changelog note is in the PR body.
- **`str(ErrorCode.FEATURE_NOT_ENABLED)`** equals the previous literal and matches the `services/base.py:111` convention -- no wire change for consumers.

**Verdict: NEEDS-WORK.** Nothing here destroys remote data: three findings are "the output contradicts itself", two are wrong-or-silent derivations with real local consequences, two are unfinished consistency. The two under "What blocks" are the ones I would want closed before merge; the rest are dangling ends of this same round.

src/keboola_agent_cli/services/merge_request_service.py
martinsifra: Confirmed and fixed in 7b2bba9 — the degradation is now recorded: `merge()` appends a `warnings[]` entry naming the raw value, the skipped cleanups, and the manual recovery (`kbagent branch reset` + `sync branch-unlink`), plus a log line. The regression test now asserts the warning instead of pinning the silence.

src/keboola_agent_cli/services/merge_request_service.py
martinsifra: Confirmed — and half of it had already landed in the parallel f32d013 (the `isinstance(..., bool)` refusal predates your suggested repro on current HEAD). Completed in 7b2bba9: the non-bool refusal on a *take* side now raises VALIDATION_ERROR blaming the server envelope (was ConfigError blaming the caller), and `is_disabled` forwards verbatim — the `bool()` coercion is gone. New tests: string `"false"` refused on the resolved path, non-bool on a take side blames the server, `False` forwards as `False`.

src/keboola_agent_cli/services/merge_request_service.py
martinsifra: Confirmed and fixed in 7b2bba9 — on the take path an absent `description` composes as `None` (wire-identical per `rebase_config`'s own omit-for-None behavior); the `resolved=` path still requires presence (caller intent), and `isDisabled` stays strict everywhere exactly per your warning — defaulting it would reintroduce the previous round's blocking finding. The schema-claim comment is softened, and both asymmetry tests exist now (`test_take_side_with_absent_description_composes_none` + `test_resolved_body_still_requires_description_presence`).

src/keboola_agent_cli/services/merge_request_service.py
martinsifra: Confirmed and fixed in 7b2bba9 — the message is gated on `branch_reset_done` (the block's outcome), not `was_active` (its precondition), and the failure warning names `kbagent branch reset --project <alias>` as the recovery. The test now asserts both: warning carries the remedy AND the message makes no reset claim.

src/keboola_agent_cli/services/merge_request_service.py
martinsifra: Confirmed and fixed in 7b2bba9 — the classifier now skips a side by the same criterion `resolve_conflict` uses for 'no content to take': null, `isDeleted`, or an empty envelope all yield `changes: []`, leaving the story to the `*_deleted` flags. Pinned by `test_tombstoned_side_yields_no_classification_rows`.

src/keboola_agent_cli/services/merge_request_service.py
martinsifra: Confirmed and fixed in 7b2bba9 — 'absent' and 'present but not numeric' get separate messages; a payload fault now reads `branchFromId is 'b-123' -- not a numeric branch id` instead of a lifecycle claim contradicting the state it prints. Pinned by `test_non_numeric_branch_from_id_names_the_payload_not_the_lifecycle`.

src/keboola_agent_cli/services/base.py
martinsifra: Confirmed and fixed in 7b2bba9 — `find_default_branch_id` now has the coerce-or-skip policy (an uncoercible id is skipped, never raised on), first unit tests (`TestFindDefaultBranchId`: numeric str coerces, 'main'/None/absent skip, multi-default falls through to the next usable entry, no-default → None), and `sync_service`'s truthiness test tightened to `is not None`. On `sync init` writing an empty `branches` list for a `None` default: that flow predates the hoist (the old inline `next(..., None)` had the identical None path), so it is deliberately unchanged here — flagged for the DMD-1900 round rather than smuggled into this one.

martinsifra:
Branch hygiene note: the Layer 1 work (DMD-1900 commits) that had ended up on this branch — my `git push HEAD:…` yesterday published a parallel session's unpushed L1 commits into this PR — has been moved out to its own branch `ms/dmd-1900/cli-layer-1`. PR #703 is back to **pure Layer 2** (18 commits, none tagged DMD-1900).

The third-review fixes were re-applied onto the pure-L2 tree as **7cd1855** (the earlier `7b2bba9` cited in the thread replies was authored on top of the L1 tree). Same fixes, three adaptations: `merge()`'s soft-failure key stays `cleanup_warnings` here (the `warnings` rename is Layer 1's), the full `isDisabled` boolean check lands as part of finding #2 (half of it had only existed on the L1 tree), and the L1-only test class does not come along. Re-verified on this tree: reviewer-style reproduction of all 7 findings PASS, 117 service tests, `make check` green (6544 passed).

zajca (approved):
Reviewed the changes since my last review (`4dbfc00..7cd1855`) -- the single commit applying that round. Ran the full suite and every quality gate locally, and reproduced each finding below with a throwaway probe test.

**Gate status on `7cd1855`:** `pytest tests/` 6544 passed / 183 skipped, `ruff check` + `ruff format --check` clean, `ty check` clean on all three touched modules, `make loc-check` / `check-sentinel-guards` OK, `gh pr checks` 3/3 green, branch contains current `origin/main`.

## Both blocking findings are fixed correctly

- **The `_coerce_branch_id` degradation is now recorded**: `logger.warning` plus a `cleanup_warnings` entry naming the raw value, the skipped cleanups and the manual recovery (`branch reset` + `sync branch-unlink`). Probed with `branchFromId: "not-a-number"` -- the warning fires, the merge still succeeds, and the legitimate published-MR null path stays silent as it should.
- **`bool(body["isDisabled"])` is gone** -- the value forwards verbatim and a non-bool is refused. The guard order is safe: a missing or null `isDisabled` lands in `missing` first, so `body["isDisabled"]` cannot `KeyError`. Blame is correctly split: `VALIDATION_ERROR` naming the server envelope on a take side, `ConfigError` on a caller-authored body.
- **The `description` relaxation on the `take=` path is genuinely wire-equivalent** -- `client/configs.py:760-761` omits the key for `None`, and the `resolved=` path still requires presence. The relaxation was correctly not extended to `isDisabled`.

## Non-blocking, this round

### 1. `_classify_three_way` does not use the criterion it claims to share

`merge_request_service.py:887-905`. The docstring and the commit message both state the new check is "the same 'no content to take' criterion `resolve_conflict` uses". That holds for `isDeleted` only. The new `bool(side.get("diff"))` clause is not shared -- probed with a live side carrying an empty envelope, `{"version": 4, "diff": {}}` and no `isDeleted`:

- `get_config_diff` returns `ours_deleted: False`, `changes: []`
- `resolve_conflict(take="ours")` on the identical shape raises `VALIDATION_ERROR: The diff's ours side carries no name, rows, configuration, isDisabled -- ... Author the resolution manually and pass it as a resolved body instead.`

I would not change the behaviour: collapsing an empty envelope to the delete resolution would destroy a configuration, and refusing it in `resolve_conflict` is the safe direction. The fix belongs in the docstring. Also worth noting the empty-envelope clause has no test -- the new test covers the tombstone half only.

### 2. `find_default_branch_id` does the opposite of `_branch_from_id_of`

`base.py:128-136`. The same commit split "absent" from "present but not numeric" in `_branch_from_id_of` precisely because one message contradicted the state printed beside it. `find_default_branch_id` collapses the two into one `None`, and its callers then print `Project '{alias}' reports no default branch` (`merge_request_service.py:624`) and `No default branch found for project` (`workspace_service.py:239`) -- the same misleading-diagnosis shape, one function over. Strictly better than the previous `ValueError` mid-operation, just asymmetric with its sibling; a `logger.warning` on the skip would close it.

### 3. The `merge()` degradation is recorded in prose only

`merge_request_service.py:693-704`. The machine-readable result is byte-identical to the legitimate published-MR null -- `branch_from_id: null`, `was_active: false`, no flag -- so a JSON consumer can only notice by string-matching `cleanup_warnings`. And in that path `message` reads just `Merge request 7 merged into production.`: the "Source branch N is being deleted" sentence is gated on `branch_from_id`, so the message says nothing about the branch while the warning next to it talks about "the merged branch". A structured key (`cleanup_skipped: true`, or echoing the raw value) would make the degradation machine-detectable.

### 4. Smaller ends

- `config_service.py:172` still tests `if folder_branch_id:` by truthiness while `sync_service.py:338` was tightened to `is not None` -- one site left out of the sweep.
- The `isDisabled` type check runs *before* the `missing` check, so a body missing `name` and carrying `"false"` reports only the `isDisabled` fault (probe-confirmed). Errors are not accumulated -- noting, not asking for a change.
- The outcome-gated message has only the negative assertion (`"Active branch reset to main" not in result["message"]`); nothing asserts the string appears when the reset succeeds. The reset itself stays covered behaviourally via `active_branch_id is None`.
- `merge_request_service.py:130` is still 110 chars against `line-length = 100` -- carried over from the last round (`E501` is ignored and `ruff format` does not rewrap docstrings, so no gate catches it).

## Cross-PR

The commit message notes Layer 1 renames `merge()`'s soft-failure key. Here it stays `cleanup_warnings` while `resolve_conflict` uses `warnings`; whichever lands second needs to reconcile the two names.

**Verdict: APPROVE.** Nothing in this round loses data or fails silently. The two blocking findings are closed the right way, and the four remaining items are consistency and diagnostics -- none of them worth holding the merge for.

martinsifra:
Closing note on the review leftovers: PR #703 is frozen at the approved head (`7cd1855`). The non-blocking items from the approval (classifier docstring/test, structured `cleanup_skipped` flag, `find_default_branch_id` diagnostics, the small ends) plus the deferrals from the earlier rounds (feature-blind `allowed_actions`, project-unscoped mapping cleanup, the `FORCE_COLOR` test fragility on main) are **not** being added here — they land with Layer 1 (DMD-1900, `ms/dmd-1900/cli-layer-1`), several of them being visible only from the command layer anyway.

They are written down, with code sites and recommended fixes, in the RFC branch: [`docs/merge-requests-layer2-followups.md` on `ms/merge-requests-rfcs`](https://github.com/keboola/cli/blob/ms/merge-requests-rfcs/docs/merge-requests-layer2-followups.md) (commit 10f0ba7), linked from the Layer 1 RFC. The soft-failure key unification (`cleanup_warnings` → `warnings`) is already done on the L1 branch and is handled there at rebase time.

PR #705, opened 2026-08-26
docs(plugin): retire version gates below the 0.80.0 floor

padak:
Retires the stale version gates you asked about, at the **0.80.0 floor, safety gates kept**.

**557 → 334 gates. 223 tags removed across 18 files. Not one word of content deleted.**

## Why

A gate earns its place only while some live install predates it. I checked `auto_update.py` rather than assuming: `maybe_auto_update()` self-upgrades pip/uv installs on startup, so the population an old gate protects rounds to nothing — only a standalone binary (self-update is disabled for brew/choco/apt/dnf), an explicit `KBAGENT_AUTO_UPDATE=false`, a dev tree, or a pip install stranded below 0.62.0 by the #424 rename can sit on an old version.

Meanwhile the stale gate keeps doing damage, because `keboola-expert.md` Rule 6 turns every tag into a refusal: the agent declines a command the user actually has. CONTRIBUTING already calls that *strictly worse than no gate*.

The two failure modes are asymmetric, and that asymmetry is the whole argument:

| | Failure | Visibility |
|---|---|---|
| Gate kept too long | agent refuses a command the user **has** | **silent, permanent** — the user never learns it exists |
| Gate removed too early | agent suggests a command the user **lacks** | **loud, self-correcting** — `No such command 'x'`, and `kbagent context` / `--help` on the real install are authoritative |

## Tag goes, content stays

This is de-tagging, not deletion. Nearly every old entry is still true — only the qualifier is obsolete:

```diff
-## `config detail` has a bulk mode (since 0.23.0)
+## `config detail` has a bulk mode
```

The body ("omit `--config-id` for every config under a component…") is current, useful, and untouched.

## Safety gates kept, at any age

The rule applied: **keep the tag wherever not knowing the version causes silent data loss or a false assurance rather than an error message.** Six qualify, all verified by reading the section rather than the title:

| Gate | Why it stays |
|---|---|
| `sync pull --force` (0.53.0) | pre-0.53.0 it **silently corrupted the sync baseline** — local edits stranded with no signal |
| plaintext `#`-secret audit (0.55.0) | an agent on 0.54.0 reports "no plaintext secrets" having checked nothing — a **false all-clear on a leaked credential** |
| manage-token default-deny (0.29.0) | below 0.29.0 the env var **is** read; assuming otherwise is a token-exposure bug |
| `--deny-writes` / `--deny-destructive` (0.22.0) | agent would trust a firewall that isn't there |
| dev-portal writes require a human (0.49.0) | human-in-the-loop guarantee |
| sync trust cluster / `--theirs` (0.72.0) | destructive reconcile |

## Also out of scope, deliberately

- **`changelog.py`** (33 gates) — the historical record; the version *is* the content.
- **`src/**/*.py` except `commands/context.py`** (13) — `# DEPRECATED (since 0.43.4)`, `UNAUTHORIZED = "…"  # (0.40.0+)`. Developer provenance; no agent reads these files.
- **`X+` inside a sentence** (59) — frequently load-bearing prose. `created by < 0.66.1 stay dormant until re-run on 0.66.1+` is a migration instruction; stripping the version destroys it. `With 0.25.0+:` would become `With :`.
- **`kbagent-pr-reviewer.md`** (3) — prose *about* the gate mechanism, using `0.22.0+` as an example.

## How, and what review caught

The de-tagger is conservative by construction: anything whose rewrite looks suspicious is skipped and reported rather than guessed at. The first attempt had two real bugs, both caught by reading the diff:

1. The regex consumed emphasis markers, turning `**Partial-state envelope**:` into `**Partial-state envelope*:`. Fixed by not matching `*` around the parenthetical.
2. A tag at the head of a continuation line left `. On projects with **branched storage**…` — a line starting with a period. The rewritten script now detects and skips that shape; both instances were then fixed by hand (the bold sentence absorbs the full stop).

## New tooling

`make gate-floor-report` (`--list-below VERSION`) gives the next floor bump a worklist instead of a hand-rolled grep — the same class of hand-run command that lost the merge race in #702. Four tests.

The policy is documented as release checklist step **4b**, including the exclusions and the safety-gate rule, so this stays a decision rather than folklore.

## Checks

`make check`: **6360 passed, 12 skipped**. `check_command_sync.py` OK (267 commands) — worth noting since `commands-reference.md` took 86 changed lines. `keboola-expert.md` is well inside its 70 000 B budget.

---

<a href="https://app.devin.ai/review/keboola/cli/pull/705" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


padak:
Superseded — reopening from a clean branch. This branch name already carried the merged #702, and GitHub never fired CI for the new commit on it.

PR #718, opened 2026-08-29
feat(cli): add --conversation-id global flag (#716)

### LINKED ISSUE #716: kbagent: SKILL.md's one-time `export KBAGENT_CONVERSATION_ID` is unsatisfiable in Claude Code (no persistent shell), forcing compound commands that defeat permission allow-rules
## Summary

`skills/kbagent/SKILL.md` rule 2 tells the agent to set the conversation ID with a one-time `export`:

> 2. **Set conversation ID**: before first kbagent call, run `export KBAGENT_CONVERSATION_ID="<unique-id>"` (e.g. session UUID). All API requests include this as `X-Conversation-ID` header for platform observability.

`references/gotchas.md` repeats it:

> Set `KBAGENT_CONVERSATION_ID` env var before running kbagent commands.

**This instruction cannot be satisfied in Claude Code.** The Bash tool does not persist shell state between calls — the working directory persists, but environment variables and shell functions do not. An `export` in one tool call is gone by the next one.

Since there is no `--conversation-id` CLI flag, the env var is the only channel, so an agent trying to comply has exactly one option: re-prepend the `export` to **every single command**. That is what actually happens in practice:

```bash
export KBAGENT_CONVERSATION_ID="kbagent-dataapp-inspect-1787921092"
WS=14356542
kbagent --json workspace query --project keboola-ai --workspace-id $WS --sql "SELECT ..." 2>&1
```

## Why this is more than cosmetic

Claude Code permission allow-rules are **prefix matches on the command string**. A command that begins with `export ...` can never match a rule like:

```json
"permissions": { "allow": ["Bash(kbagent --json workspace query:*)"] }
```

So the documented usage pattern makes kbagent **impossible to allowlist**. Every invocation falls through to the auto-mode safety classifier, which has no model for a custom binary doing `workspace query --sql "..."` against a production stack and therefore prompts. In a long read-only analysis session this means a manual confirmation on essentially every query, plus tripping the "N consecutive actions were blocked" guard.

Net effect: a rule intended to *improve* observability actively degrades the agent experience and blocks the permission system from working as designed.

## Reproduce

1. In Claude Code, run `export FOO=bar` in one Bash call.
2. Run `echo $FOO` in the next Bash call → empty.
3. Confirm no flag alternative exists:

```console
$ kbagent --help | grep -ci conversation
0
```

## Current state in v0.91.0

Both instructions are present as quoted above:

```
skills/kbagent/SKILL.md:36
skills/kbagent/references/gotchas.md:2405
```

A grep of the whole `skills/` tree finds no acknowledgement anywhere that agent harnesses do not persist shell state — no mention of `settings.json` `env` blocks, no per-command env prefix guidance, and no note that a standalone `export` dies between tool calls. The only `settings.json` references are in `permissions-workflow.md`, and they concern deny-rules.

## Suggested fixes

Roughly in order of preference:

1. **Add a `--conversation-id` global flag.** Cleanest fix. Commands then start with `kbagent`, so they are allowlistable, and the ID stays per-session:
   ```bash
   kbagent --json --conversation-id <id> workspace query --project p --sql "..."
   ```
2. **Fix the docs to stop implying a persistent shell.** If the env var must remain the only channel, document the two shapes that actually work in an agent harness — a per-command prefix, or Claude Code's `settings.json` `env` block — and state explicitly that a standalone `export` does not persist across tool calls.
3. **Fall back to a generated ID when unset.** `gotchas.md` says "If unset, the header is omitted." Emitting a per-invocation or per-day ID instead would preserve some observability without forcing agents into compound commands.
4. **Ship recommended allow-rules with the plugin** for the read-only verbs (`workspace query`, `config detail`, `data-app detail`, `bucket/table/job list|detail`), so users get sane defaults. Note these only pay off once (1) or (2) lands and commands actually begin with `kbagent`.

## Environment

- kbagent `v0.91.0`
- Claude Code on macOS (darwin 25.5.0), auto mode enabled

padak:
## What

`skills/kbagent/SKILL.md` rule 2 instructs the agent to set the observability conversation ID with a
one-time `export KBAGENT_CONVERSATION_ID="<unique-id>"`. **That instruction cannot be satisfied in an
agent harness.** Claude Code's Bash tool persists the working directory but not environment
variables, so the export is gone by the next tool call, and there was no flag alternative
(`kbagent --help | grep -c conversation` → 0).

The only way to comply was therefore to re-prepend the export to every command. Because Claude Code
permission allow-rules are prefix matches on the command string, a command starting with
`export ...` can never match `Bash(kbagent --json workspace query:*)` — so kbagent became impossible
to allowlist, every invocation fell through to the safety classifier, and a rule intended to
*improve* observability degraded the agent experience instead.

This PR implements option 1 from the issue (the flag) and option 2 (the doc correction).

## How

A `--conversation-id` global option on the Typer callback, published into the environment rather
than threaded through the Typer context — that is where every consumer already reads it:

- `BaseHttpClient.__init__` stamps `X-Conversation-ID` from `os.environ` for all seven HTTP clients;
- `doctor`'s `conversation_id` check reports on it;
- scheduled-agent / `cli_command` subprocesses inherit it for free;
- `serve` already sets it exactly this way (`commands/serve.py:271`).

Precedence: an explicit flag beats an inherited `KBAGENT_CONVERSATION_ID` (it is the more specific
instruction). With neither set, the header stays omitted, exactly as before — this is purely
additive.

## Not done here

Options 3 and 4 from the issue are product decisions rather than bug fixes, so they are deliberately
left out:

- **Auto-generating an ID when unset** changes what the platform's observability data looks like
  (every unset invocation would mint a fresh unique ID), which is a call for whoever consumes that
  telemetry.
- **Shipping recommended allow-rules with the plugin** is a plugin-config decision about which verbs
  are safe to pre-approve on someone else's machine.

## How it was tested

New `TestConversationIdFlag` in `tests/test_cli.py` (5 cases): flag sets the ID, flag wins over the
env var, env var still honoured without the flag, neither set leaves it unset, and — the one that
matters — the flag actually reaches the outgoing `X-Conversation-ID` header on a constructed client,
not just the environment.

Also verified end-to-end out-of-process: `kbagent --json --conversation-id kbagent-test-4242 doctor`
leaves the env set and a subsequently constructed client carries
`x-conversation-id: kbagent-test-4242`.

`make check` exits 0.

## Doc surfaces

Every surface that carried the unsatisfiable instruction now states that a standalone `export` does
not survive between tool calls, and points at the flag (or the harness's own env block for a whole
session):

- `plugins/kbagent/skills/kbagent/SKILL.md` rule 2 — the line the issue quotes.
- `plugins/kbagent/skills/kbagent/references/gotchas.md` "Conversation ID" — rewritten, tagged
  `(since vNEXT, #716)`.
- `src/keboola_agent_cli/commands/context.py` `AGENT_CONTEXT` — both the "IMPORTANT: Set Conversation
  ID" block and the env-var table.
- `CLAUDE.md` global-options line.

One process note worth flagging: a `(since vNEXT)` gate **cannot** be written on the `# ` comment
lines inside CLAUDE.md's `## All CLI Commands` fenced block — `scripts/check_version_gates.py` parses
them as ATX markdown headings, and a placeholder in a heading is always fatal. The version gate for
this entry therefore lives in `gotchas.md`, with a comment in CLAUDE.md explaining why.

No CLI *command* was added / renamed / removed (this is a global option), so the command-sync
surfaces are untouched; `make command-sync-check` and `make skill-check` are green.

No version bump and no `changelog.py` entry — that belongs to the release PR.

Fixes #716

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -348,13 +348,15 @@ plugins/kbagent/
 
 ```
 # Global options: --json, --verbose, --no-color, --config-dir, --conversation-id, --deny-writes, --deny-destructive, --allow-env-manage-token
-# --conversation-id ID (since vNEXT, #716): sets the X-Conversation-ID observability header for the
+# --conversation-id ID (#716): sets the X-Conversation-ID observability header for the
 #   invocation; equivalent to KBAGENT_CONVERSATION_ID and takes precedence over it. Exists because an
 #   agent harness does not persist shell state between tool calls, so the documented one-time `export`
 #   was unsatisfiable -- and re-prepending `export ...` to every command stops it matching a
 #   `Bash(kbagent ...)` permission allow-rule, so every call fell through to the safety classifier.
 #   For a whole session, set the env var in the harness's own env block instead (Claude Code:
-#   settings.json -> env). Neither set = header omitted, as before.
+#   settings.json -> env). Neither set = header omitted, as before. Version gate for this entry
+#   lives in gotchas.md -- a `(since vNEXT)` tag cannot be written on these `# ` comment lines,
+#   because check_version_gates.py parses them as ATX markdown headings (where a `vNEXT` is fatal).
 # Headless / token-only (0.50.0+): export KBAGENT_PROJECT_FROM_ENV=1 + KBC_TOKEN + KBC_STORAGE_API_URL to synthesize an in-memory `__env__` project (no `project add`, no config.json on disk; token never persisted). Use `--project __env__`. Same env setup also powers `kbagent serve`.
 
 kbagent auth login [--stack URL|alias] [--device-code] [--register-projects]
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -347,7 +347,14 @@ plugins/kbagent/
 > "Plugin synchronization map" for the full list.
 
 ```
-# Global options: --json, --verbose, --no-color, --config-dir, --deny-writes, --deny-destructive, --allow-env-manage-token
+# Global options: --json, --verbose, --no-color, --config-dir, --conversation-id, --deny-writes, --deny-destructive, --allow-env-manage-token
+# --conversation-id ID (since vNEXT, #716): sets the X-Conversation-ID observability header for the
+#   invocation; equivalent to KBAGENT_CONVERSATION_ID and takes precedence over it. Exists because an
+#   agent harness does not persist shell state between tool calls, so the documented one-time `export`
+#   was unsatisfiable -- and re-prepending `export ...` to every command stops it matching a
+#   `Bash(kbagent ...)` permission allow-rule, so every call fell through to the safety classifier.
+#   For a whole session, set the env var in the harness's own env block instead (Claude Code:
+#   settings.json -> env). Neither set = header omitted, as before.
 # Headless / token-only (0.50.0+): export KBAGENT_PROJECT_FROM_ENV=1 + KBC_TOKEN + KBC_STORAGE_API_URL to synthesize an in-memory `__env__` project (no `project add`, no config.json on disk; token never persisted). Use `--project __env__`. Same env setup also powers `kbagent serve`.
 
 kbagent auth login [--stack URL|alias] [--device-code] [--register-projects]
--- a/plugins/kbagent/skills/kbagent/SKILL.md
+++ b/plugins/kbagent/skills/kbagent/SKILL.md
@@ -33,7 +33,7 @@ If kbagent is not installed or you need the full standalone reference, run `kbag
 ## Rules
 
 1. **Always use `--json`**: `kbagent --json <command>` for parseable output
-2. **Set conversation ID**: before first kbagent call, run `export KBAGENT_CONVERSATION_ID="<unique-id>"` (e.g. session UUID). All API requests include this as `X-Conversation-ID` header for platform observability.
+2. **Set conversation ID**: pass `--conversation-id "<unique-id>"` (e.g. session UUID) on every kbagent call -- `kbagent --json --conversation-id <id> <command>`. All API requests include it as the `X-Conversation-ID` header for platform observability. **Do not use a standalone `export`**: agent harnesses (Claude Code included) do not persist shell state between tool calls, so the variable is gone by the next command -- and prefixing each command with `export ...` stops it matching a `Bash(kbagent ...)` permission allow-rule. For a whole session, set `KBAGENT_CONVERSATION_ID` in the harness's own env block (Claude Code: `settings.json` -> `env`) instead.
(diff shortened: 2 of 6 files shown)


padak (commented):
## Independent verification review (issue #716)

Re-verified the issue and the fix from a clean worktree, independently of the session that authored this PR.

### Issue reproduction — confirmed valid

- `KBAGENT_CONVERSATION_ID` is read only from `os.environ` at client construction (`http_base.py:111-113`); no flag existed anywhere in the CLI.
- The one-time-`export` instruction was present on all three agent-facing surfaces (`SKILL.md:36`, `gotchas.md:2405`, `context.py` AGENT_CONTEXT), and none of them acknowledged that agent harnesses do not persist shell state.
- The permission-rule consequence is real: Claude Code allow-rules are prefix matches, so `export ...; kbagent ...` compound commands can never match `Bash(kbagent ...:*)`.

### Fix verification

- **Design**: publishing the flag into `os.environ` in the root callback is the right minimal seam — all seven HTTP clients read the env var lazily (after the callback), `doctor` picks it up, subprocesses inherit it, and `serve` already respects a pre-set value when auto-generating its own ID, so `kbagent --conversation-id X serve` pins the serve session for free.
- **Tests**: ran `tests/test_cli.py::TestConversationIdFlag` locally on this branch — 5/5 pass, including the wire-level assertion that the header actually lands on a constructed `httpx` client. Env hygiene via `patch.dict` is correct (no leakage between tests).
- **Version gating**: verified the CLAUDE.md claim — `check_version_gates.py`'s `HEADING_RE` (`^ {0,3}#{1,6} `) is line-based with no fence awareness, so a `vNEXT` on the `# --conversation-id` comment line inside the code block would indeed be fatal on every PR. Keeping the `(since vNEXT, #716)` gate in `gotchas.md` only is correct.
- **Doc surfaces** (convention #17): `context.py`, `CLAUDE.md`, `SKILL.md`, `gotchas.md` all updated. `keboola-expert.md` and `commands-reference.md` never mentioned conversation ID, so skipping them is right (no drift introduced).
- **Feature-PR rules**: no version bump, no `changelog.py` entry — correct per the release process.

### Nits (non-blocking)

1. The four env-propagation tests discard the CLI `result` without asserting exit code 0. If `project list` ever started failing in the callback path, these tests would keep passing on a half-executed command. One `assert result.exit_code == 0` per test would pin that down.
2. Undocumented edge: an explicit empty flag (`--conversation-id ""`) overwrites a set env var with `""`, which suppresses the header entirely. Harmless (and arguably a feature — an explicit opt-out), but worth a sentence in `gotchas.md` if anyone ever hits it.

### Verdict

Fix is correct, minimal, and complete for options 1+2 of the issue; deferring options 3 (auto-generated fallback ID) and 4 (shipped allow-rules) as product decisions is the right call. Ready to merge from my side.

PR #731, opened 2026-09-01
feat(telemetry): make kbagent usage visible in Keboola telemetry

soustruh:
kbagent now posts usage telemetry. Resolves [AI-3780](https://linear.app/keboola/issue/AI-3780/create-telemetry-for-kbagent).

## What it does

- Each CLI command posts one usage event to its own project (`POST /v2/storage/events`), the same mechanism the original Keboola Go CLI uses. Connection stores it as `ext.keboola.cli.`.
- The event records the command name, the outcome, and the duration. It never records argument values.
- A line typed in the REPL posts the same event as a direct command.
- `kbagent serve` posts `ext.keboola.cli.serve` per REST request. Each route maps to its CLI command, so serve and CLI usage share one `command` and differ only by `interface`.
- The Encryption API calls (kbagent makes them when a command encrypts secret values) now send kbagent's `User-Agent`. They built their own headers, so they used httpx's default `python-httpx/<version>` instead of `keboola-cli/<version>`. They now match every other Keboola call, including `X-Conversation-ID`.

## Safety

- Best-effort: it cannot change a command's exit code, and it ignores a blocked events endpoint after a short timeout.
- Opt-out: `KBAGENT_DISABLE_TELEMETRY=1` or `DO_NOT_TRACK=1`.
- This is usage telemetry, not an audit trail. Connection records mutations server-side regardless.

## Testing

- Six targeted tests: CLI success shape, failure plus best-effort, opt-out for both env vars, REPL parity, serve-map completeness, serve end-to-end.
- Full suite green (6419 passed). No version bump (feature PR, `vNEXT` convention).

## Follow-up (not this repo)

- A transformation reads the new `ext.keboola.cli.` event into the warehouse, next to the existing `ext.keboola.keboola-as-code.` model.


soustruh (commented):
Automated review (kbagent-pr-reviewer). Verdict: **request changes.** Two blocking issues break the PR's own "never stalls the command" claim.

## Blocking

**B-1 – Standalone binaries post no telemetry.** `build/package/entry.py` still calls `app()`. The release build freezes the PyInstaller binary from this entry point (`.github/workflows/release-kbagent.yml`), not the console script, so it never reaches the new `run()` wrapper. Every binary install (brew, choco, apt, dnf, winget, deb, rpm) posts zero events. Fix: call `run()` there.

**B-2 – A blocked events endpoint stalls every command by about 12 seconds.** `trigger_event()` posts through the shared `_do_request`. That path retries a connect failure three times with backoff, and each attempt runs up to the 3 second timeout. Against an unroutable host the call took 12.34 seconds (measured). This contradicts the docstring and the README, which both promise a short timeout and no stall. It is exactly the firewalled-network case the safety section describes. Fix: send the events POST as one bounded attempt with no retries.

## Non-blocking

- **NB-1 – The `-p` alias is not recognized.** `_option_value` matches only `--project`. A command run with `-p` (feature, stream, project, token) attributes the event to the wrong project or drops it.
- **NB-2 – `run()` has no test.** The new console entry point is untested. Add one for a success, a mapped error, and an unmapped exception.
- **NB-3 – Undisclosed change.** The encryption sub-client User-Agent fix is unrelated to telemetry and unmentioned. Disclose it in the description, or move it to its own fix.
- **NB-4 – No gotcha entry.** A background POST after every command is new behavior. Add a `(since vNEXT)` note to `gotchas.md` with the two opt-out env vars.

The rest reads well: clean CLI / REPL / serve separation, an exhaustive route→command map with a bidirectional completeness test, and `make check` green (6419 passed).

soustruh:
**Second review — branch at 6b3217d**

Tests pass. No new command was added. Four points, all fixed in 6b3217d:

1. **A project logged in with `auth login` could pause the command.** Before sending the event, kbagent could refresh that login's access token. The refresh calls the server and can wait for another kbagent process to finish its own refresh. The event's short timeout does not cover that wait, so a normal command could pause for a few seconds. Fixed: the event now uses the login's current token and never refreshes it. If the token would need a refresh first, kbagent skips that one event.

2. **`kbagent serve` recorded a crashed request as a success.** A request that failed with an error the server did not catch was logged with status 200 instead of 500. Fixed: it now logs 500 for that case.

3. **`auth status` exit 3 was recorded as a failure.** Exit 3 means you are signed out, which the command's own help calls a normal result, not an error. Fixed: exit 3 from `auth status` is now recorded as a normal outcome.

4. **One docs file was updated, the other was not.** The note about the new events went into `gotchas.md` but not into the agent's own reference file, `keboola-expert.md`. Fixed: both now describe the events and the two behaviors above.

New tests cover the `auth login` path, the failed serve request, and the exit-3 case.

src/keboola_agent_cli/server/app.py
```diff
@@ -604,6 +609,86 @@ def _default_permission_engine(
...
+        if route is None or isinstance(route, Mount):
+            return
+        if path in telemetry.SERVE_SKIP_PATHS or path.startswith("/health"):
+            return
+
+        # command == the equivalent CLI command so serve and CLI telemetry share one
+        # `command` (only `interface` differs). The map is exhaustive (enforced by a
+        # test); an empty value marks a route with no CLI counterpart -- still logged,
+        # under a route-derived label.
+        template = getattr(route, "path", "")
```
zajca: Reviewed by Opus.

`SERVE_COMMAND_MAP` was generated from `app.openapi()["paths"]` (the exhaustiveness test in `tests/test_serve_telemetry.py:25-33` compares against exactly that), which FastAPI builds from `APIRoute.path_format` -- converter suffixes stripped. The middleware instead looks the route up by `getattr(route, "path", "")`, which is the raw declared path *with* the converter (`fastapi/routing.py` sets `self.path = path` and derives `path_format` separately via `compile_path`).

**Failure scenario:** `GET /storage/buckets/prod/in.c-mybucket` matches the route declared as `/buckets/{project}/{bucket_id:path}`, so `route.path` is `/storage/buckets/{project}/{bucket_id:path}` while the map key is `("GET", "/storage/buckets/{project}/{bucket_id}")`. The lookup misses and the event is posted with `command="bucket detail"` (the endpoint function name) instead of the mapped `"storage bucket-detail"`, so serve and CLI telemetry for the same operation land under different command names and cannot be joined.

Every route declared with a `{...:path}` converter is affected -- 16 of them: all `storage` bucket/table/column/snapshot operations, `sharing unshare`/`unlink`, and `data-app secrets-get`. Their map entries (`_serve_command_map.py:38,77,130,135-138,227-244`) are dead code, and the test cannot catch it because it never consults `route.path`.

The same fallback also fires for routes excluded from the schema: `GET /index.html` (`app.py:1009`, `include_in_schema=False`) is not in the map and not in `SERVE_SKIP_PATHS`, so it posts `command=" serve ui index"` (note the leading space from `_serve_ui_index`).

This defeats the PR's stated goal that serve and CLI share one `command` vocabulary. Fix: key on `getattr(route, "path_format", None) or getattr(route, "path", "")`, and make the exhaustiveness test iterate `app.routes` using the same attribute the middleware reads rather than the OpenAPI document, so schema-excluded routes are covered too.

src/keboola_agent_cli/telemetry.py
```diff
@@ -0,0 +1,282 @@
...
+            token=project.token,
+            command=command,
+            project_id=project.project_id,
+            success=exit_code == 0 or expected_nonzero,
+            error=_resolve_error_text(error),
+            duration_s=duration_s,
+            configuration_id=None,
+            extra_params=None,
+        )
+    except Exception:
```
zajca: Reviewed by Opus.

The entire telemetry path is exception-opaque and logs nothing: `emit_cli_invocation` has `except Exception: return` here and at line 166, `send_serve_event` has the same at line 230, and the middleware wraps `_emit` in `contextlib.suppress(Exception)` (`server/app.py:648`). `telemetry.py` does not import `logging` at all.

**Failure scenario:** the `keboola.cli` component id is not registered on a stack, so `POST /v2/storage/events` answers 4xx. `_send_event` raises `KeboolaApiError`, `except Exception: return` discards it, and every command exits 0 with no output even under `kbagent --verbose`. Telemetry is 100% broken on that stack and nothing anywhere reports it.

Best-effort is the right behaviour, but silent is not the same as unobservable. There is currently no way -- not even with `--verbose`, which sets `logging.DEBUG` in `cli.py` -- to tell whether an event was posted, skipped, or rejected. The rest of the codebase does not work this way: `http_base._do_request` logs every retry decision at `logger.debug`.

This matters most for the failure modes that are invisible by construction: an unregistered component id, a `params` payload rejected by the component's event schema (`client/misc.py:31` documents that `params` must match it, and serve sends an extra `path` key the CLI never sends), a 403 from a restricted token, or a firewalled events endpoint. Any of these makes telemetry a permanent no-op that ships silently and is only discovered when the downstream warehouse transformation finds no rows.

Add a `logger.debug("usage event skipped/failed: %s", exc)` in each handler (and one on the deliberate skip paths at lines 166 and 168), keeping the swallow itself.

src/keboola_agent_cli/server/app.py
```diff
@@ -729,6 +814,10 @@ async def _lifespan(app_: FastAPI):
     )
     install_registry(app, registry)
 
+    # Outermost middleware: posts a best-effort per-request usage event
+    # (ext.keboola.cli.serve) after the response, off the event loop.
+    app.add_middleware(_UsageTelemetryMiddleware, config_store=config_store)
```
zajca: Reviewed by Opus.

The middleware posts an event for every matched, non-infra serve request, with no sampling, batching, or read/write distinction.

**Failure scenario:** an operator runs `kbagent serve --ui-dist ...` and leaves the Jobs page open for an 8-hour workday. `GET /jobs?project=prod` fires every 8 seconds (`web/frontend/src/pages/Jobs.tsx:36`, `refetchInterval: 8000`), producing roughly 3,600 `ext.keboola.cli.serve` events with `command="job list"` in the customer's project event log for that one idle tab, each opening and tearing down its own TLS connection to the Storage events endpoint.

Other timers compound it: `JobsAll.tsx:76` (15s), `Agents.tsx:188` (10s) and `Agents.tsx:1154` (5s). This writes hundreds of events/hour into the customer's own project event log -- the log surfaced in the Connection UI and consumed by `keboola-expert.md`'s own event-audit guidance.

Each event is also a fresh connection: `_send_event` calls `make_telemetry_client` (a new `KeboolaClient`, hence a new `httpx.Client`), posts, and closes it in the `finally`, so every poll pays a full TCP+TLS handshake. And each one occupies a slot in asyncio's default executor (`min(32, cpu+4)` workers) for up to the request timeout, contending with the real `asyncio.to_thread` work in `server/routers/jobs.py:156,171`.

CLI telemetry is one event per human/agent invocation, which is proportionate. Serve telemetry is one event per HTTP request, which is not -- the volume is driven by UI refresh intervals, not by user actions. Consider restricting serve events to mutating methods, deduplicating/sampling repeated `(command, project)` pairs within a short window, or reusing a single long-lived client for the events endpoint.

src/keboola_agent_cli/telemetry.py
```diff
@@ -0,0 +1,282 @@
...
+
+# Longest error text carried in ``results.error``. Keeps the event well under the
+# events API's 200 KB cap and avoids dumping large response/state buffers.
+_MAX_ERROR_LEN = 1000
+
+# Command wrappers that are not a unit of work themselves. ``serve`` posts a
+# per-request event from its middleware; ``repl`` posts a per-line event from its
+# own loop. The outer invocation of each is excluded so it does not add a second,
+# session-level event on top of the real work it dispatches.
+_CLI_EXCLUDED_COMMANDS = frozenset({"serve", "repl"})
```
zajca, line 53: Reviewed by Opus.

`_CLI_EXCLUDED_COMMANDS` only excludes `serve` and `repl`, so every other resolvable command posts an event synchronously in `run()`'s `finally` before the process exits. That turns commands with no network dependency at all into network-dependent ones.

**Failure scenario:** on a machine with no route to the stack (offline laptop, air-gapped runner) and `KBAGENT_DISABLE_TELEMETRY` unset, `kbagent context` -- which only echoes a hardcoded module-level string (`commands/context.py`) and does nothing else -- now blocks on a connect attempt to `POST /v2/storage/events` and takes seconds instead of milliseconds, on every single invocation. `changelog` (bundled text) and `version` are in the same category.

The cost is not bounded by 3s either. `TELEMETRY_TIMEOUT` is passed as httpx's request-level `timeout`, which applies per phase (connect, read, write, pool), so an unreachable or blackholed endpoint can add several seconds. kbagent's primary consumer is an AI agent issuing many sequential commands, where this compounds.

The user-facing docs added in this PR overstate the guarantee: `README.md` says telemetry "can never change a command's exit code or stall it" and `gotchas.md` says "never stalls it". Either exclude the local-only commands from telemetry, or reword the guarantee to state the actual bound.

src/keboola_agent_cli/http_base.py
```diff
@@ -196,12 +197,16 @@ def _do_request(
         # An explicit override wins: RETRY_SAFE_METHODS reasons about the METHOD,
         # but idempotency is a property of the endpoint. See the `retry_safe` arg.
         retry_safe = method.upper() in RETRY_SAFE_METHODS if retry_safe is None else retry_safe
+        # A caller can cap the attempt count. Best-effort telemetry passes 1 so a
+        # blocked or unreachable events endpoint fails in one bounded attempt --
+        # never the ~12s retry+backoff that would stall the command behind it.
+        attempts = MAX_RETRIES if max_attempts is None else max_attempts
```
zajca, line 207: Reviewed by Opus.

The loop bound is now `attempts` (from the new `max_attempts` argument), but the three retry log statements and the exhaustion error still interpolate the module constant: `logger.debug("Retry attempt %d/%d", attempt + 1, MAX_RETRIES, ...)` at lines 239, 270 and 294, and `f"Request failed after {MAX_RETRIES} retries"` at line 320.

**Failure scenario:** a future caller passes `max_attempts=2` for a semi-critical request. On the first retry the debug log reads `Retry attempt 1/3` and, after both attempts fail with no `last_response`, the raised error reads `Request failed after 3 retries` -- both describing a 3-attempt budget the request never had, misdirecting anyone debugging from the logs.

The only current caller passes `max_attempts=1`, which never reaches those lines, so this is latent rather than live -- but it is a trap for the next caller, and cheap to fix by threading `attempts` through.

Also, `max_attempts` is the only parameter of `_do_request` missing from the docstring's `Args:` block (lines 175-187 document `method`, `path`, `client`, `base_url`, `retry_safe` and `**kwargs`), and the summary line still reads "up to MAX_RETRIES times".

src/keboola_agent_cli/server/app.py
```diff
@@ -604,6 +609,86 @@ def _default_permission_engine(
...
+
+        try:
+            await self.app(scope, receive, _send)
+        except Exception:
+            # An unhandled error is turned into a 500 above this middleware, so
+            # `_send` never sees the response start -- record 500, not the default
+            # 200, or the one request telemetry most wants to catch reads as success.
+            status["code"] = 500
+            raise
+        finally:
```
zajca, line 659: Reviewed by Sonnet.

`_UsageTelemetryMiddleware.__call__` runs `await self._emit(...)` inside its own `finally` block (app.py:646-649), and `_emit` awaits `asyncio.to_thread(telemetry.send_serve_event, ...)` (app.py:680-688), which performs a synchronous HTTP POST bounded by `TELEMETRY_TIMEOUT=3.0s` with no retry. Because this happens in `finally`, the ASGI task representing the current request does not return control to the server until that POST attempt finishes (success, timeout, or connection failure) -- even though the response bytes were already flushed to the client via `_send` earlier in `self.app(scope, receive, _send)`.

The class docstring claims 'the event loop is never stalled', which is true for *other* connections (the loop keeps running), but the specific request's own ASGI task is stalled for up to 3s whenever the events endpoint is slow, unreachable, or firewalled -- e.g. behind a corporate egress policy that blocks or rate-limits calls to connection.keboola.com. That reintroduces, at the serve layer, exactly the stalling risk the prior review round explicitly fixed for the CLI path (commit 1fdc9344, 'stop a blocked events endpoint from stalling commands'): there the fix bounded a single invocation's added latency to ~4s once; here, every single REST request handled by `kbagent serve` pays a similar bounded delay, and on a persistent HTTP/1.1 keep-alive connection (the Node BFF or `kbagent http` proxy's typical usage pattern) this can serialize the next request on the same connection behind the current one's telemetry attempt, and extends graceful-shutdown drain time under SIGTERM while requests are in flight.

Failure scenario: `kbagent serve` runs behind a network policy that blocks outbound POST to `connection.keboola.com/v2/storage/events` (or the events endpoint is merely slow). Every REST request the BFF/UI sends -- job runs, config reads, storage browsing -- now takes up to an extra ~3s to complete from the caller's perspective on that connection, even though none of that time is spent doing the caller's actual work; this is the opposite of the 'never stalls the command' guarantee the telemetry feature is designed to provide.

Suggested direction: schedule `self._emit(...)` as a detached task (e.g. `asyncio.create_task`, tracked so it can be awaited/cancelled during the lifespan shutdown handler already defined in `create_app`) instead of awaiting it inline, so a slow or blocked events endpoint can never add latency to the request/response cycle it is instrumenting.

zajca (commented):

src/keboola_agent_cli/server/app.py
soustruh: Fixed in 2b4707c. The middleware now keys the lookup on `path_format` (the OpenAPI shape the map is generated from), so a `{...:path}` route resolves to its mapped command instead of the function-name fallback. Schema-excluded UI routes (the SPA shell) are skipped, so they no longer post an event either. Test: `test_serve_maps_a_path_converter_route`.

src/keboola_agent_cli/telemetry.py
soustruh: Fixed in 2b4707c. Each swallow and skip path now leaves a DEBUG line, findable with `--verbose`. The swallow itself is unchanged, so it stays best-effort. Test: `test_a_rejected_event_is_recorded_at_debug` (a 403 leaves a debug line).

src/keboola_agent_cli/server/app.py
soustruh: Fixed in 2b4707c, both parts. serve now posts an event only for mutating requests, so read polling no longer writes to the event log. And a static-token client is reused across events instead of a fresh connection each time, closed at serve shutdown. Tests: `test_serve_skips_read_requests`, `test_static_token_telemetry_client_is_shared`.

src/keboola_agent_cli/telemetry.py
soustruh: Fixed in 2b4707c. `context`, `changelog` and `version` are excluded, so a local command never does a network POST. The README and gotchas no longer say "never stalls". They state the actual bound: one attempt, one short timeout, no retry. Test: `test_local_only_commands_post_no_event`.

src/keboola_agent_cli/http_base.py
soustruh: Fixed in 2b4707c. The retry logs and the exhaustion message now use `attempts` (the real budget), not `MAX_RETRIES`, and `max_attempts` is documented in the Args block. Test: `test_retry_logs_use_the_max_attempts_budget` (a `max_attempts=2` run logs `attempt 1/2`).

src/keboola_agent_cli/server/app.py
soustruh: Fixed in 2b4707c. The event no longer runs in the request's `finally`. It is scheduled as a detached task, so a slow or blocked events endpoint can never add latency to the request it instruments. In-flight events are drained at serve shutdown so they are not lost. Test: `test_serve_does_not_wait_for_the_telemetry_post`.

zajca (approved):

PR #739, opened 2026-09-03
chore(lint): upgrade ruff to >=0.16 and use its default rule set

soustruh:
## What

- Upgrade ruff to `>=0.16` (minimum only). The lock moves from 0.15.17 to 0.16.5.
- Remove the explicit `select`. Use ruff's default rule set, plus `extend-select`, a small `ignore`, and one per-file ignore.
- Fix every new lint finding in the code. Do not suppress findings.

## Why

- ruff 0.16.0 changed the default rule set from 59 rules to 413.
- The old `select` (`E,F,I,W,UP,B,SIM,RUF`) was set in one early commit and never changed. Because it was explicit, ruff did not apply the new default.
- The new default does not include ~62 rules the repo used (for example `B904`, `E711`, `E712`, `E731`, `SIM105`). `extend-select` adds these back.
- Result: ruff runs more rules and reports more problems in the code.

## Rule config

`extend-select`: `B, E, F, G, I, RUF, SIM, UP, W`. Each rule has a reason in `pyproject.toml`.

`ignore` (each has a reason):

- `E501`: the formatter controls line length.
- `B008`: Typer uses `= typer.Option(...)` as an argument default on purpose.
- `BLE001`: this CLI catches broad `except Exception` on purpose, to collect errors across projects.
- `TRY004`: input checks raise `ValueError`. The CLI maps `ValueError` to exit code 2 (usage error). `TypeError` would change the exit code.
- `ISC004` (changelog.py only): that file holds many strings that are split across lines by hand.

## Findings fixed

`PYI034` (`__enter__ -> Self`), `PYI036`, `PLW1510` (explicit `check=`), `S110` and `S112` (debug log instead of silent `pass`), `C408`, `PIE810`, `FURB188`, `FURB192`, `PLC0206`, `PERF402`, `PLE0704`, `RUF036`, `FLY002`, `TRY401`, and `ISC004` (8 outside changelog.py).

## Verification

- `ruff check` passes. `ruff format . --check` passes.
- Test suite: 6406 passed, 7 skipped.
- The pre-commit hook (`scripts/pre-commit`) and CI (`check` job) both run ruff through `uv run`. They use ruff 0.16.5 from the lock.

## Open PRs

- `#728` and `#731` need no changes.
- Under the new config, both branches show only the old findings from `main`. This commit fixes those findings.
- Both branches go green after a normal rebase onto `main`.


soustruh:
Also verified manually, LGTM ദ്ദി(•ᴗ•)

PR #747, opened 2026-09-08
fix(cli): partial failures set the exit code and the headline (CLI-6)

### LINKED ISSUE #745: Commands report success and exit 0 even when per-item operations fail
## Summary

Some CLI commands print a green success line and exit 0 even when part of the work fails. `sync push` and `sync clone` are the confirmed cases. A per-item failure is non-fatal. The command adds it to `result["errors"]` and does not change the exit code. So a run where some or all items fail still reports success.

## Confirmed cases

- `sync push`: the command always prints `formatter.success("Pushed: N created, ...")`. It lists per-item errors only as warnings below it (`commands/sync.py`).
- `sync clone`: `_format_clone_result` always prints `formatter.success("Cloned into X: N created")`. It prints this for any `N`, and whether or not there are errors (`commands/sync.py`).
- The command layer raises a non-zero exit only for `FileNotFoundError`, `FileExistsError`, `ConfigError`, and `KeboolaApiError`. Collected per-item failures never change the exit code.

## Impact

- A clone where every config fails still reports `Success ... 0 created` and exit 0.
- A push where half the configs fail reports success and exit 0.
- A script cannot tell success from partial or total failure by the exit code. A person must read the `created` count and the warning lines.

## Expected

The success line and the exit code should match what happened:

- Zero items done and some failed: not a success. Use a non-zero exit.
- Some items done and some failed: state the partial result plainly. Put the failed count in the main line. Set the exit code by the team's convention.
- No failures: success and exit 0. No change.

## Scope

Audit every command that prints a fixed success line while it collects per-item errors. This is not a sync-only problem. Any such command has the same defect. Check the sync group first: `push`, `clone`, `pull`. Then check the other multi-item commands, for example the bulk storage and describe commands, `org setup`, and `project invite`.

## Notes

I found this while fixing CLI-5 (#743). The branch-remap change let a fresh clone succeed. Its three cross-config-reference failures then showed only as warnings under a green "Success".

Related: CLI-5 / #743.

padak:
## What

Commands that survive a per-item failure collected it into the result and then printed a fixed green `Success:` line and exited **0** anyway. A `sync clone` where *every* config failed reported `Success ... 0 created` and exit 0; a `sync push` where half the configs failed reported success too. The failures showed only as warnings underneath, so nothing in the exit code distinguished a clean run from a total failure -- a person had to read the `created` count and count the warning lines, and a script could not branch on it at all.

## Why

Per-item resilience is deliberate: one bad config must not abort a whole push. The defect is that the resilience never reached the *reporting*. This makes the outcome legible in both places a caller looks:

- **exit code 1** whenever at least one item failed -- the convention `storage delete-table`, `storage file-tag` and `storage describe-migrate` already used, now applied where the audit found it missing;
- **the human headline states the failure** instead of claiming success: `Failed: Pushed: 3 created, 0 updated, 0 deleted, 1 failed`.

`--json` output is **unchanged** and is emitted *before* the non-zero exit, so a JSON caller still receives the complete payload including the per-item error detail. Exit 1 here never means "no output".

## Commands changed

| Command | Failure signal |
|---|---|
| `sync push` (single + `--all-projects`) | `errors[]` / `summary.failed` |
| `sync clone` | `errors[]` |
| `sync pull` / `sync diff` with `--all-projects` | `summary.failed` |
| `org setup` | `projects_failed` |
| `project invite --from-csv` | `failed` |

A clean run is untouched: no failures still means the green success line and exit 0. `no_changes` and `--dry-run` results are not failures.

## Scope decision: read-only fan-outs excluded

`billing credits`, `job list`, `schedule list`, `notification list` and the other multi-project **reads** document per-project degradation as intended behavior ("per-project failures degrade individually, the run never aborts"). Failing the whole read because one project is unreachable would break existing callers, so they still exit 0 and report the per-project error in `errors`. This is stated explicitly in the new `gotchas.md` entry so an agent knows which array to check.

The audit also confirmed the bulk storage commands (`delete-table`, `truncate-table`, `delete-column`, `delete-bucket`, `file-delete`, `file-tag`, `snapshot-delete`, `describe-batch`, `describe-migrate`) were **already** correct -- no change needed there.

## Implementation note

`sync push`'s inline human rendering moved into `_render_push_result` because its early `return`s for `no_changes` / `dry_run` would otherwise bypass the exit call. Behavior for those two statuses is byte-identical.

`commands/sync.py` grows by 23 code lines and stays over the 800-line soft ceiling it was already over before this PR (1156 -> 1179). Splitting it is a much larger refactor than this bug fix warrants; flagging it rather than bundling it.

## How it was tested

`tests/test_partial_failure_exit_code.py` -- 16 tests covering every changed command, both directions (failure -> exit 1, clean -> exit 0), the `no_changes` / `--dry-run` non-failure cases, and that `--json` still emits the full payload alongside the non-zero exit.

Verified the tests are not vacuous: with `src/` reverted to `main`, **9 of the 16 fail**; the other 7 are the clean-run control cases that must keep passing.

`make check` passes (6585 passed, 12 skipped).

## Docs

Per convention #17: `gotchas.md` (new entry, `(since vNEXT)`), `commands-reference.md` (four entries, incl. correcting the now-wrong "Exit 0 even with `failed > 0` -- inspect the JSON" line), `CLAUDE.md` command list, and `context.py` `AGENT_CONTEXT` exit-code section. No command was added, renamed or removed.

Fixes #745


soustruh:
Dear Claude, without reviewing this PR any further, I'd just like to highlight this one thing – `exit_on_item_failures` raises `typer.Exit` itself, while the existing `map_error_to_exit_code` is a pure mapper that returns the code and lets the caller raise. Two exit-code helpers side by side with opposite shapes is inconsistent. Consider returning the code (`int | None`) and keeping the `raise` at the call site, matching the existing convention.

PR #744, opened 2026-09-08
fix(sync): clone into a fresh target defaults to its production branch (CLI-5)

soustruh:
## Why

A fresh `sync clone` reused the source project's branch id. `repoint_manifest_project` re-pointed only `project.id` and `api_host`. It left `manifest.branches` unchanged. So the copied manifest kept the source project's production branch. `_resolve_branch_id` then took `manifest.branches[0].id`, the source branch. A clone with no `--branch` resolved that branch. The target does not have it, so the clone failed with `Branch id "..." does not exists`. The target is empty and has one branch: its own production branch. So `--branch` should not be required. `sync init` already resolves a project's default branch from the API. `sync clone` did not.

## What changed

- `repoint_manifest_project` takes an optional `default_branch_id` and re-points the production branch onto it.
- The clone path resolves the target's default branch from the API (`list_dev_branches` + `find_default_branch_id`), the same as `sync init`.
- A fresh clone now defaults to the target's production branch. `--branch` is optional. It still works as an override for a dev branch.
- The `--branch` help text is correct after this change, so it stays as is.

## Tests

- `test_repoint_manifest_project_remaps_default_branch`: repoint re-points the branch when a default branch id is given, and leaves it unchanged when the id is absent.
- `test_clone_repoints_branch_to_target_default`: a fresh clone re-points the manifest onto the target's default branch, with a mocked `list_dev_branches`.
- The full sync/clone suite stays green.

## Verified live

Against an empty target project, the same command without `--branch`:

- Global build (no fix): the clone fails on the source branch id and creates nothing.
- This branch: the clone succeeds and creates the configs on the target's own production branch.

Closes #743. Linear: CLI-5.

Docs: `(since vNEXT)` note added to `sync-workflow.md`.

Issue #649, opened 2026-08-22
sync: production diff/push after 'sync pull --branch' flags the whole orphaned main/ tree as added (mass-duplicate risk)


padak:
## Summary

After `sync pull --branch <dev>` re-targets `manifest.configurations` to the dev branch, the default-branch tree (`main/`) becomes orphaned on disk. A subsequent **production** `sync diff` / `sync push` then classifies **every config in `main/`** as `added` with an empty `config_id` — one `sync push` away from duplicating the entire production project.

This is the same failure class as #482, but on the axis #482's fix does not cover: #482 scoped the untracked-file scan to the source branch subtree, which protects a *dev-branch* push from *orphaned* trees. The production push reads `main/` as its legitimate source — and `main/` is exactly the tree that `pull --branch` just orphaned.

## Reproduced live (v0.88.0, project 5946)

```
sync init + sync pull                  # 14 configs tracked on main
branch create "issue644-b"             # -> id 51406
sync pull --branch 51406               # materializes issue644-b/ subtree (17 configs)
                                       # AND re-targets ALL manifest entries to branch 51406
branch reset                           # back to production
sync diff                              # target: production
```

Result:

```
summary: {'added': 19, 'modified': 0, ..., 'unchanged': 15, 'remote_only': 0}
  added | 01m0nnjyyv1szn29yjmsnh06y1 | ZZ ... Exp C      <- dev-only config, tracked on dev, diffed against production
  added | 01m0nnnx4efmnqq3ktdk064901 | ZZ ... Exp B      <- dev-only config, tracked on dev, diffed against production
  added | (empty) | e2e-pr1-e2e-1784722173               <- ordinary production config, file in main/ WITH correct id
  added | (empty) | ... (14 more, the whole main/ tree)
```

`sync push --dry-run` confirms these would all be POSTed as new configurations. 34-duplicate incidents of the #644 shape are one plausible outcome; this one is worse — it needs **zero** scaffolds, just the documented multi-branch pull.

## Root cause (two independent legs)

[...]

PR #753, opened 2026-09-10
fix(sync): keep the config folder on pull and clone (CLI-9)

soustruh:
## What changed

`sync pull` captures each config's UI folder (`KBC.configuration.folderName`) into the manifest, and `sync clone` recreates it in the target — for a plain clone, a `--branch` clone, and a git-branching production clone.

## The bug

A config's folder is config metadata under `KBC.configuration.folderName`. It comes from a branch-only `search/component-configurations` call, not from the `list_components_with_configs` body that pull reads for config data. `sync pull` never fetched it, so `sync pull` and `sync clone` dropped the folder and every config landed in the tree root. The old Go CLI (`keboola-as-code`) kept the folder, so this was a regression.

## The fix

- `sync pull` fetches the folder for each config and stores `KBC.configuration.folderName` in the manifest entry's metadata.
- `sync clone` re-points its production configs onto the branch push resolves for the target (from `_resolve_branch_id`, production normalized to `0`), so the create-path writeback matches the placeholder and forwards the `KBC.*` metadata. This holds for a plain, `--branch`, and git-branching production clone. Dev-branch configs keep their own tree. `_config_dir` falls back to `branches[0].path`, not a literal `"main"`, so overrides resolve on a non-`"main"` branch dir.
- No push-side code changed — the create path already forwards `KBC.*` manifest metadata.

## Degraded lookup

If the folder lookup fails during a pull (API error, no resolvable branch, non-dict body), pull keeps each config's folder from the previous pull instead of stripping it from every entry, and reports `folder_lookup_failed`. Human `sync pull` prints a warning; `--json` carries the flag.

## Scope

The create path carries the folder. Changing a folder on an existing config through a plain `sync push` still does not propagate — `propagate_kbc_metadata` runs only on create. Use `config set-folder`.

## Testing

- End-to-end clone → push tests (real push, not mocked) for a plain, git-branching, and `--branch` clone, each asserting `set_config_metadata` receives the folder.
- Pull tests: a failed lookup keeps the prior folder; a successful empty lookup drops a removed folder.
- Unit tests for the branch re-point and the `_config_dir` fallback.
- Full non-e2e suite green; `ruff`, `ty`, `make loc-check`, version gates clean.

Linear: CLI-9

### DIFF
--- a/src/keboola_agent_cli/services/sync_service.py
+++ b/src/keboola_agent_cli/services/sync_service.py
@@ -518,6 +518,7 @@ class SyncService(BaseService):
         with client:
             components = client.list_components_with_configs(branch_id=branch_id)
             self._ensure_branch_registered(manifest, branch_id, client)
+            folder_map = self._fetch_config_folders(client, branch_id)
 
             if not no_storage:
                 try:
@@ -979,6 +980,14 @@ class SyncService(BaseService):
                         # shape version is stamped alongside it (issue #686).
                         CONFIG_HASH_VERSION_KEY: CONFIG_HASH_VERSION,
                     }
+                # Keep the UI folder (KBC.configuration.folderName) with the
+                # config. Pull dropped it before, so every pulled or cloned
+                # config landed in the root (CLI-9). The push create path
+                # already forwards KBC.* manifest metadata through
+                # propagate_kbc_metadata, so storing it here closes the loop.
+                folder_name = folder_map.get(lookup_key)
+                if folder_name:
+                    cfg_metadata["KBC.configuration.folderName"] = folder_name
                 new_configurations.append(
                     ManifestConfiguration(
                         branchId=branch_id or 0,
@@ -2041,6 +2050,27 @@ class SyncService(BaseService):
     # ------------------------------------------------------------------
     # Private helpers
     # ------------------------------------------------------------------
+
+    @staticmethod
+    def _fetch_config_folders(client: Any, branch_id: int | None) -> dict[str, str]:
+        """Map ``{component_id}/{config_id}`` to its UI folder name.
+
+        The folder is config metadata (``KBC.configuration.folderName``), not a
+        field in the ``list_components_with_configs`` body, so pull must fetch it
+        separately. The search endpoint that serves it is branch-only, so a
+        production pull (``branch_id`` is ``None``) resolves the default branch
+        id first -- the same fallback ``ConfigService`` uses. A lookup failure
+        degrades to an empty map: a missing folder must never abort the pull.
+        """
+        try:
+            folder_branch_id = branch_id or find_default_branch_id(client.list_dev_branches())
+            if not folder_branch_id:
+                return {}
+            result = client.list_config_folder_metadata(branch_id=folder_branch_id)
+            return result if isinstance(result, dict) else {}
+        except Exception:
+            logger.debug("config-folder metadata lookup failed", exc_info=True)
+            return {}
 
     @staticmethod
     def _resolve_branch_id(
--- a/plugins/kbagent/skills/kbagent/references/gotchas.md
+++ b/plugins/kbagent/skills/kbagent/references/gotchas.md
@@ -750,6 +750,27 @@ Useful for targeting a freshly-created dev branch without running `branch use` o
 to the manifest or to the config store, so subsequent commands without `--branch`
 fall back to the normal priority chain.
 
+## `sync pull` / `sync clone` now keep the config folder (`KBC.configuration.folderName`)
+
+*(since vNEXT)*
+
+A config's UI folder is config metadata under `KBC.configuration.folderName`. It
+comes from a branch-only `search/component-configurations` call, not from the
+`list_components_with_configs` body that pull reads for config data. Before vNEXT
+`sync pull` never fetched it, so `sync pull` and `sync clone` silently dropped the
+folder and every config landed in the tree root. Nothing warned about it. On a
+pre-vNEXT kbagent a golden-reference `sync clone` still produces a folder-less
+clone -- check the version, or verify the folders came across after a clone.
+
+Since vNEXT pull fetches the folder for each config and stores
+`KBC.configuration.folderName` in the manifest entry's `metadata`. The push create
+path already forwards `KBC.*` manifest metadata (see "`sync push` fresh-CREATE
+writeback now updates placeholders in place" above), so a clone recreates the
+folder in the target with no push-side change. Only the CREATE path carries it:
+changing a folder on an existing config through a plain `sync push` still does not
+propagate (`propagate_kbc_metadata` runs only on create) -- use
+`config set-folder`.
+
 ## `storage create-table --if-not-exists` returns `action: skipped` instead of raising on duplicate display name
 
 Opt-in flag (default `False`, so existing callers are unaffected). When set,
--- a/tests/test_sync_storage_jobs.py
+++ b/tests/test_sync_storage_jobs.py
@@ -30,7 +30,7 @@ from keboola_agent_cli.services._sync_storage import (
     write_storage_metadata,
 )
 from keboola_agent_cli.services.sync_service import SyncService
-from keboola_agent_cli.sync.manifest import ManifestConfiguration
+from keboola_agent_cli.sync.manifest import ManifestConfiguration, load_manifest
 
 # ---------------------------------------------------------------------------
 # Shared test data
@@ -1452,3 +1452,127 @@ class TestPullJobsFallback:
         # call_args_list, not call_count: see the thread-safety note in
         # test_falls_back_to_per_config_when_limit_insufficient.
         assert len(pull_client.list_jobs.call_args_list) == 101
+
+
+# ===================================================================
+# 6. Config folder (KBC.configuration.folderName) capture on pull (CLI-9)
+# ===================================================================
+
+
+class TestPullConfigFolder:
+    """pull() must keep each config's UI folder in the manifest (CLI-9).
+
+    The folder is config metadata (``KBC.configuration.folderName``), served by
+    a separate branch-only search endpoint, not by ``list_components_with_configs``.
+    Before the fix pull never fetched it, so every pulled or cloned config lost
+    its folder. The push create path already forwards ``KBC.*`` manifest
+    metadata (see ``TestPropagateKbcMetadata`` in test_sync_service.py), so
+    capturing it on pull closes the round-trip.
+    """
+
+    def test_pull_captures_config_folder_into_manifest(
+        self, tmp_config_dir: Path, tmp_path: Path
+    ) -> None:
+        """A config with a folder lands with KBC.configuration.folderName set."""
+        project_root = tmp_path / "project"
+        project_root.mkdir()
+
+        store = _init_project(tmp_config_dir, project_root)
+
+        pull_client = _make_sync_mock_client(
+            components_response=SAMPLE_COMPONENTS_SIMPLE,
+        )
+        # Only cfg-001 has a folder; cfg-002 has none.
+        pull_client.list_config_folder_metadata.return_value = {
+            "keboola.ex-http/cfg-001": "Extractors",
+        }
+        svc = SyncService(
+            config_store=store,
+            client_factory=lambda url, token: pull_client,
+        )
+
+        svc.pull(alias="prod", project_root=project_root)
+
+        manifest = load_manifest(project_root)
+        by_id = {c.id: c for c in manifest.configurations}
+        assert by_id["cfg-001"].metadata.get("KBC.configuration.folderName") == "Extractors"
+        # A config with no folder must NOT get the key (round-trip would
+        # otherwise create a bogus folder on push).
+        assert "KBC.configuration.folderName" not in by_id["cfg-002"].metadata
+
+    def test_pull_folder_lookup_uses_resolved_branch(
+        self, tmp_config_dir: Path, tmp_path: Path
+    ) -> None:
+        """The folder search runs against the branch the pull resolved."""
+        project_root = tmp_path / "project"
+        project_root.mkdir()
+
+        store = _init_project(tmp_config_dir, project_root)
+
+        pull_client = _make_sync_mock_client(
+            components_response=SAMPLE_COMPONENTS_SIMPLE,
+        )
+        pull_client.list_config_folder_metadata.return_value = {}
+        svc = SyncService(
+            config_store=store,
+            client_factory=lambda url, token: pull_client,
+        )
+
+        svc.pull(alias="prod", project_root=project_root)
+
+        # SAMPLE_BRANCHES default id is 12345 -> manifest fallback branch id.
+        pull_client.list_config_folder_metadata.assert_called_once_with(branch_id=12345)
+
+
+class TestFetchConfigFolders:
+    """Unit tests for SyncService._fetch_config_folders."""
+
+    def test_uses_given_branch_without_dev_branch_lookup(self) -> None:
+        """A resolved branch id is used directly; no default-branch lookup."""
+        client = MagicMock()
+        client.list_config_folder_metadata.return_value = {"comp/1": "Folder A"}
+
+        result = SyncService._fetch_config_folders(client, 42)
+
+        assert result == {"comp/1": "Folder A"}
+        client.list_config_folder_metadata.assert_called_once_with(branch_id=42)
+        client.list_dev_branches.assert_not_called()
+
+    def test_resolves_default_branch_for_production(self) -> None:
+        """branch_id=None (production) resolves the default branch first."""
+        client = MagicMock()
+        client.list_dev_branches.return_value = [{"id": 777, "isDefault": True}]
+        client.list_config_folder_metadata.return_value = {"comp/1": "Prod Folder"}
+
+        result = SyncService._fetch_config_folders(client, None)
+
+        assert result == {"comp/1": "Prod Folder"}
+        client.list_config_folder_metadata.assert_called_once_with(branch_id=777)
+
+    def test_empty_when_no_default_branch(self) -> None:
+        """No resolvable branch -> empty map, endpoint never called."""
+        client = MagicMock()
+        client.list_dev_branches.return_value = []
+
+        result = SyncService._fetch_config_folders(client, None)
+
+        assert result == {}
+        client.list_config_folder_metadata.assert_not_called()
+
+    def test_degrades_to_empty_on_api_error(self) -> None:
+        """A folder-lookup failure must not abort the pull."""
+        client = MagicMock()
+        client.list_config_folder_metadata.side_effect = RuntimeError("boom")
+
+        result = SyncService._fetch_config_folders(client, 42)
+
+        assert result == {}
+
+    def test_non_dict_response_becomes_empty(self) -> None:
+        """A non-dict response degrades to an empty map."""
+        client = MagicMock()
+        client.list_config_folder_metadata.return_value = ["unexpected"]
+
+        result = SyncService._fetch_config_folders(client, 42)
+
+        assert result == {}
--- a/tests/test_sync_storage_jobs.py
+++ b/tests/test_sync_storage_jobs.py
@@ -1462,12 +1462,12 @@ class TestPullJobsFallback:
 class TestPullConfigFolder:
     """pull() must keep each config's UI folder in the manifest (CLI-9).
 
-    The folder is config metadata (``KBC.configuration.folderName``), served by
-    a separate branch-only search endpoint, not by ``list_components_with_configs``.
+    The folder is config metadata (``KBC.configuration.folderName``) from a
+    separate branch-only search endpoint, not from ``list_components_with_configs``.
     Before the fix pull never fetched it, so every pulled or cloned config lost
     its folder. The push create path already forwards ``KBC.*`` manifest
-    metadata (see ``TestPropagateKbcMetadata`` in test_sync_service.py), so
-    capturing it on pull closes the round-trip.
+    metadata (see the ``test_propagate_kbc_metadata_*`` tests in
+    test_sync_service.py), so capturing it on pull closes the round-trip.
     """
 
     def test_pull_captures_config_folder_into_manifest(


soustruh (commented):
# Review of #753 — fix(sync): keep the config folder on pull and clone (CLI-9)

## Summary

This closes a real regression against the reference Go CLI: `sync pull` never fetched a config's UI folder (`KBC.configuration.folderName`), so every pulled or cloned config landed in the tree root. The fix adds one static helper (`_fetch_config_folders`) that reuses the exact branch-resolution idiom already established in `config_service.py`'s `list_config`, injects the folder into `cfg_metadata` for both the locally-modified and normal code paths, and degrades to an empty map on any lookup failure so a pull never aborts over it. Source-level tracing confirms every claim in the PR description holds up: the branch fallback matches `config_service.py`, the push-side propagation via `propagate_kbc_metadata` / `writeback_create_config_in_manifest` already preserves the key (and is already tested), and a folder removed remotely is correctly dropped on re-pull because `cfg_metadata` is rebuilt fresh every time rather than merged with old metadata. Seven new focused unit tests cover the helper and the pull integration; the full suite and `make check` are clean. Verdict is REQUEST CHANGES for one reason only: this repo already documents the mirror-image push-side fix for this exact metadata key in exhaustive, version-tagged detail in `gotchas.md` / `sync-workflow.md` — the pull-side half of the same story has no equivalent entry.

## Verdict

- **Blocking findings**: 1
- **Nits**: 1

## Blocking findings

### `[B-1]` `plugins/kbagent/skills/kbagent/references/gotchas.md` — no entry for the pull-side folder-loss fix

`sync pull`/`sync clone` silently dropped every config's UI folder before this fix (every config landed in the tree root). This repo already documents the mirror-image push-side fix for the same metadata key in exhaustive, version-tagged detail (`gotchas.md:697-704`, `sync-workflow.md:194-198`, tagged `v0.47.0`) — the diff adds no matching bullet for the pull-side gap it closes. On a pre-fix kbagent install, `sync clone` (keboola-expert.md matrix row: "Provision a new project from a golden reference") still silently produces a folder-less clone, and nothing warns an agent to check the version or verify folders came across. Per CONTRIBUTING.md's Plugin synchronization map, a missing `gotchas.md` tag for a behavior change is BLOCKING. Add a `(since vNEXT)`-tagged bullet next to the existing "KBC.configuration.\* metadata propagates on CREATE" note.

## Non-blocking findings

### `[NB-1]` `src/keboola_agent_cli/services/sync_service.py:2054` / `services/config_service.py:164` — duplicated folder-fetch-with-branch-fallback logic

The new `_fetch_config_folders` static method reimplements the exact try/except block already in `config_service.py`'s `list_config` — same fallback chain (`branch_id or find_default_branch_id(...)`), same `isinstance(result, dict)` guard, and the identical `"config-folder metadata lookup failed"` debug message. `base.py`'s own docstring already states it is "the one home for the `isDefault` scan shared by the config, sync, workspace and merge-request services" — this is a good candidate to extract there next to `find_default_branch_id`, so a future change to the fallback logic doesn't need to land twice.

### `[NB-2]` `src/keboola_agent_cli/client/configs.py:309` — unverified pagination on the folder-search endpoint at pull scale

`list_config_folder_metadata` calls `GET .../search/component-configurations` with no `limit`/offset handling. The method pre-dates this PR (used by `config list`), but `sync pull` is a new, larger-scale caller — a whole-project fan-out rather than a single interactive listing. If the endpoint caps results server-side, a project with more folder-tagged configs than one page would have `folder_map.get(lookup_key)` return `None` for the overflow configs — indistinguishable from "no folder assigned" (`sync_service.py:988`), so pull would silently drop folders for *some* configs rather than none. Could not verify against a live project (no `E2E_API_TOKEN`/`E2E_URL` or `/tmp/kbagent-e2e` in this environment). Recommend confirming against a project with 100+ foldered configs, or checking the Storage API docs for this route's default page size.

## Nits

- `[NIT-1]` `tests/test_sync_storage_jobs.py:1469` — the `TestPullConfigFolder` docstring points to "`TestPropagateKbcMetadata` in test_sync_service.py" for the push-side tests; that class doesn't exist. The actual coverage is `TestFreshCreateWriteback` (`test_sync_service.py:3013`, `test_propagate_kbc_metadata_*` methods at lines 3139-3211).

## Verification log

- Working tree verified: `/home/soustruh/cc-dev/cli-9`, branch `fix/cli-9-config-folder`, `HEAD=64293c87e234099aac408bf9c9f94f447262067c`, `git status` clean — matches the PR head.
- `gh auth status` → authenticated as `soustruh` ✓
- Read `CONTRIBUTING.md` (Checklist for new commands, Plugin synchronization map, Releasing a new version), `CLAUDE.md` (convention #17, `## All CLI Commands`), `plugins/kbagent/agents/keboola-expert.md` §1/§2/§3 before touching the diff.
- `gh pr view 753 -R keboola/cli --json ...` → OPEN, `fix(sync):` prefix matches a genuine bug fix, +155/-1, 2 files (`services/sync_service.py`, `tests/test_sync_storage_jobs.py`) ✓
- `gh pr diff 753 -R keboola/cli` → 196 lines; `git diff origin/main...HEAD --stat` matches the PR's reported file list exactly (no drift from a stale local `main`) ✓
- `git diff origin/main...HEAD -- src/keboola_agent_cli/cli.py 'src/keboola_agent_cli/commands/**/*.py'` → empty: no command/flag added or changed, so `commands/context.py`, `CLAUDE.md` "All CLI Commands", `permissions.py` `OPERATION_REGISTRY`, `commands-reference.md`, and the keboola-expert.md Tool Selection Matrix are correctly untouched (their triggers require a command/flag change) ✓
- Layer-violation greps (typer/click/formatter in services, httpx/requests in commands) → empty ✓
- Convention greps (magic numbers, raw `error_code=` strings, bare `except:`, `print()`, token literals) → empty; the only "token" hits are the pre-existing `lambda url, token: pull_client` factory pattern, not a leak ✓
- `make check` → **exit 0**, `6576 passed, 12 skipped, 5 warnings in 83.12s` (re-run with full log captured, `EXIT_CODE=0` confirmed; stage markers for `ruff check`, `ruff format --check`, `ty check`, `check_version_gates.py`, `check_command_sync.py`, `generate_changelog.py --check`, `check_error_codes.py`, `check_sentinel_guards.py`, `check_file_size.py`, and `pytest` all present in the log) ✓
- `make loc-check` → `sync_service.py` not listed among WARN/FAIL files; `scripts/check_file_size.py` reports it at 1603 code lines vs. the grandfathered baseline of 1655 — the file shrank overall despite this PR's +30 lines, so the hard-ceiling ratchet is respected ✓
- `uv run pytest tests/test_sync_storage_jobs.py::TestPullConfigFolder tests/test_sync_storage_jobs.py::TestFetchConfigFolders -v` → all 7 new tests pass in isolation ✓
- `uv run pytest tests/test_sync_storage_jobs.py tests/test_sync_service.py -q` → 172 passed, confirming no regression among the broader pull/push test modules from the new `list_config_folder_metadata`/`list_dev_branches` calls ✓
- Read `sync_service.py` `pull()` end-to-end (lines 462-1010, 2054-2140): confirmed `branch_id` passed to `_fetch_config_folders` is the fully-resolved value from `_resolve_branch_id` (production commonly resolves to the manifest's registered branch id, not `None`, once `sync init` has run); confirmed the folder injection at lines 988-990 sits *after* the `locally_modified`/`else` branch, so it applies uniformly to both; confirmed `cfg_metadata` is a fresh dict per pull (not merged from `existing_metadata`), so a folder removed on the remote is correctly absent after re-pull ✓
- Compared `sync_service.py:2054-2072` (`_fetch_config_folders`) against `config_service.py:164-176` (`list_config`'s inline folder-fetch) line by line — same `find_default_branch_id` fallback, same endpoint, same degrade-on-error design ✓
- Read `_sync_writeback.py` (`propagate_kbc_metadata`, `writeback_create_config_in_manifest`) and `_sync_clone.py` (`clone_project`) to verify the PR description's push/clone-side claims independently rather than trust them — confirmed accurate; `writeback_create_config_in_manifest`'s own docstring already names `KBC.configuration.folderName` as an example of preserved metadata ✓
- Checked `sync.manifest.ManifestConfiguration.metadata` → `dict[str, Any]` with `model_config = ConfigDict(extra="allow")` — adding a new dict key has no backward/forward-compat schema risk for older/newer kbagent reading the same manifest ✓
- Checked `permissions.py` `OPERATION_REGISTRY` → `sync.pull` already registered as `"read"`, pre-existing and untouched (no new command) ✓
- Checked `plugins/kbagent/skills/kbagent/references/gotchas.md`, `sync-workflow.md`, `commands-reference.md`, `commands/context.py` for any existing mention of pull-side folder handling → none found, confirming `[B-1]` is a genuine gap and not already covered elsewhere ✓
- `grep -rn "KBC.configuration.folderName"` across `src/` → consistent string everywhere (client, service, command, changelog, context.py) — no typo risk ✓
- Could **not** reproduce live against a real Keboola project: no `E2E_API_TOKEN`/`E2E_URL` set, and `/tmp/kbagent-e2e` does not exist in this environment. Relied on source-level tracing plus the green full test suite instead.

(none)

soustruh:
All four findings handled.

Addressed in `3c5801b`:

- **B-1 (blocking)**: added a `(since vNEXT)` gotcha for the pull side in `gotchas.md`, next to the mirror-image push-side note. The tag sits on the first body line, not the heading, so `check_version_gates.py` stays green.
- **NIT-1**: fixed the `TestPullConfigFolder` docstring. It now points at the `test_propagate_kbc_metadata_*` tests in `test_sync_service.py`.

Deferred to keep this PR surgical, tracked as follow-ups:

- **NB-1** (duplicated folder-fetch logic, extract one helper into `services/base.py`): CLI-10.
- **NB-2** (folder-search paging in `list_config_folder_metadata`): CLI-11. I could not confirm the endpoint's page size from the docs (pages moved, apiary returned HTTP 502), so the ticket verifies it against a real project before any paging change.

`make check` stays green.

zajca (changes requested):
Reviewed at 3c5801b against base `main` @ 99a000b (merge-base confirmed identical, so this is the true base→head diff).

The pull-side capture is correct and well tested. Two findings below; the first one means the headline behaviour in the PR title/description does not actually hold.

---

## 1. (blocking) `sync clone` does not recreate the folder — the create path never sees the `KBC.*` metadata

`services/_sync_writeback.py:146-156` matches the placeholder entry on `(branch_id, component_id, path)`. `sync/clone.py:83-86` (`repoint_manifest_project`) re-points **only** `manifest.branches[0].id` onto the target's default branch (CLI-5 / #744); it leaves `branch_id` on every copied `ManifestConfiguration` pointing at the **source** project's production branch id.

On push into the target, `_resolve_branch_id` returns the target's default branch id (`sync_service.py:2137-2140`), so the match fails, a *second* entry is appended carrying only `pull_hash`/`pull_config_hash`, and `propagate_kbc_metadata` finds no `KBC.*` keys to POST.

Reproduced by driving the repo's own functions (no mocks of the logic under test):

```
source entry branch_id: 12345
after repoint -> branches[0].id: 555 | entry.branch_id: 12345
matched existing entry: False
metadata handed to propagate_kbc_metadata: {'pull_hash': 'fh', 'pull_config_hash': 'ch'}
set_config_metadata called: False
manifest entries: 2        <- plus a stale duplicate entry
```

This is production-realistic: `sync init` writes the real default branch id (`sync_service.py:332-338`) and pull stores `branchId=branch_id or 0` with that real id — as this PR's own test asserts (`assert_called_once_with(branch_id=12345)`). Two different projects never share a production branch id, and passing `--branch` mismatches as well.

Why the suite stays green: the clone fixtures in `tests/test_sync_clone.py:37-43` build manifests with `branch id = 0`, which hides the mismatch. There is no test covering pull → clone → propagated folder end to end.

Extra concern: `plugins/kbagent/skills/kbagent/references/gotchas.md` now states "a clone recreates the folder in the target with no push-side change". A documented guarantee the code does not honour is worse than silence — it tells users (and the agent reading gotchas.md) not to bother verifying.

Suggested direction: either re-point `cfg.branch_id` alongside `manifest.branches[0].id` in `repoint_manifest_project` (which also removes the duplicate-entry symptom), or resolve the writeback entry without the branch-id predicate on a fresh clone. Whichever way, please add a test that goes pull → clone → assert `set_config_metadata` receives `KBC.configuration.folderName`, and keep the gotchas.md wording in step with whatever ships.

## 2. A degraded folder lookup silently deletes folder names already in the manifest

`services/sync_service.py:2054-2072` (helper), `:988-990` (write), `:1059` (`manifest.configurations = new_configurations`)

`_fetch_config_folders` degrades to `{}` on any failure — API error, unresolvable branch, non-dict body — and only logs at `debug`. Pull rebuilds `cfg_metadata` from scratch and then replaces `manifest.configurations` wholesale, and nothing carries the previous `KBC.configuration.folderName` over. `existing_metadata` (`:579-581`) is right there and is already consulted for `config_hash_version` (`:958`), just not for the folder.

Net effect: one transient failure of `search/component-configurations` (or a stack/token where that endpoint is unavailable) during `sync pull` strips the folder from **every** manifest entry, silently. The pull output shows nothing, so a user cannot tell "no folders configured" from "the lookup failed" — which quietly reintroduces exactly the regression this PR fixes.

`pull()` has no `warnings[]` channel like push, but `pull_details` is available; even a single detail entry noting the degraded lookup would make it diagnosable. Preserving the prior value when the map came back empty *because of a failure* (as opposed to a successful empty result) would be better still.

---

### Verified as fine

- `lookup_key = f"{component_id}/{config_id}"` (`sync_service.py:656`, `config_id` always `str`) matches the key format built in `client/configs.py:309-341`.
- The helper faithfully mirrors the established idiom in `config_service.py:164-177`, default-branch fallback included.
- `if folder_name:` correctly omits the key for a folder-less config, and the test asserts it.
- No version bump and no `changelog.py` entry — correct for a feature PR; `(since vNEXT)` is the right placeholder and is not in an ATX heading.
- Locally green: 7 new tests, `test_sync_storage_jobs.py` + `test_sync_service.py` (172 passed), `ruff check`, `ruff format --check`, `make loc-check`.

src/keboola_agent_cli/services/_sync_clone.py
```diff
@@ -121,14 +121,21 @@ def clone_project(
...
         repoint_manifest_project(
             manifest,
             project_id=target_project.project_id or 0,
             api_host=urlparse(target_project.stack_url).netloc,
-            default_branch_id=(
-                _target_default_branch_id(service, target_project)
-                if branch_override is None
-                else None
+            default_branch_id=target_default_branch_id,
+            config_branch_id=(
```
zajca: Reviewed by Opus.

When `branch_override` is given, `default_branch_id` is `None` (so `manifest.branches[0].id` keeps the **source** branch id) while `config_branch_id=branch_override` stamps every `ManifestConfiguration.branch_id` with a branch that is not in `manifest.branches`. The three override helpers run immediately afterwards (`_sync_clone.py:141-143`) and resolve the on-disk directory through `sync/clone.py::_config_dir`, which uses `branch_path_map(manifest).get(cfg.branch_id, _DEFAULT_BRANCH_DIR)` — a hardcoded `"main"` fallback, unlike `branch_scope.branch_tree_path` which falls back to `branches[0].path`. Every override therefore reads `<target>/main/<path>`.

Reproduced against a reference tree whose branch dir is `master` (a `sync init --git-branching` workspace on a `master` default branch), cloning with `--branch 777`:

```
base  (99a000b): cfg branch_id=[12345]  bucket rewrites: 1  -> in.c-new.users
head  (51b51b0): cfg branch_id=[777]    bucket rewrites: 0  -> in.c-old.users (unchanged)
```

`apply_instance_rename` is worse than a no-op: its `if src.exists()` guard skips the on-disk move but `cfg.path` is rewritten anyway, so the manifest ends up pointing at a directory that does not exist:

```
renamed: 1
manifest path: extractor/keboola.ex-http/Globex
dir Globex exists: False
dir Acme still exists: True
```

The subsequent `diff`/`push` then classifies that config as `deleted` (its `pull_hash` is set) and the config is never created in the target, while the stale `Acme/` tree is left untracked. All of this is silent — the clone reports `status: cloned`.

Suggested fix: register `branch_override` in `manifest.branches` as part of the re-point (or pass the resolved branch path through), and align `_config_dir`'s fallback with `branch_tree_path` so it degrades to `branches[0].path` rather than the literal `"main"`. A clone test combining `--branch` with `bucket_map`/`instance_rename` on a non-`main` branch dir would have caught this; the added unit test `test_repoint_manifest_project_config_branch_without_default` asserts the new `branch_id` but never exercises the helpers that consume it.

src/keboola_agent_cli/services/_sync_clone.py
```diff
@@ -121,14 +121,21 @@ def clone_project(
...
+                branch_override if branch_override is not None else target_default_branch_id
```
zajca, line 143: Reviewed by Opus.

The clone stamps `cfg.branch_id` with the target's **numeric** default branch id, but `writeback_create_config_in_manifest` matches on `entry.branch_id == (branch_id or 0)` (`_sync_writeback.py:148-155`), and the canonical manifest spelling for production is `0` — `pull()` writes `branchId=branch_id or 0` (`sync_service.py:1000`). Which of the two spellings matches depends entirely on what `_resolve_branch_id` returns for the target:

- non-git-branching, no `--branch`: falls through to `manifest.branches[0].id` → the numeric id. Post-PR this matches; pre-PR (`branchId: 0`) it did not. This is the bug the PR fixes.
- git-branching enabled on the default git branch: returns `None` (`sync_service.py:2116-2121`, "the default git branch always resolves to `None` (production)") → `target_branch = 0`. Post-PR `cfg.branch_id` is the numeric id and **no longer matches**; pre-PR it was `0` and did match.

`copy_reference_tree` is a plain `shutil.copytree`, so a git-branching reference tree arrives at the target with `gitBranching.enabled: true`, `.git/` and `branch-mapping.json` intact — this path is reachable, not hypothetical.

Verified directly against `writeback_create_config_in_manifest` at head:

```
git-branching production clone (push resolves None):
  cfg.branch_id=555 push branch_id=None -> entries=2 matched=False KBC={}                              <- head
  cfg.branch_id=0   push branch_id=None -> entries=1 matched=True  KBC={'KBC.configuration.folderName': 'Extractors'}  <- base
non-git-branching production clone (push resolves 555):
  cfg.branch_id=555 push branch_id=555  -> entries=1 matched=True  KBC={'KBC.configuration.folderName': 'Extractors'}  <- head
  cfg.branch_id=0   push branch_id=555  -> entries=2 matched=False KBC={}                              <- base
```

So for git-branching workspaces the head reintroduces both symptoms the PR description attributes to the old behaviour: the folder is dropped and a duplicate manifest entry is appended. Rather than hardcoding one spelling, derive `config_branch_id` from what push will actually resolve (`SyncService._resolve_branch_id(target_project, manifest, target_path, branch_override) or 0`), or normalise production equivalence inside `writeback_create_config_in_manifest` so `0`, `None` and the default branch id all match. The new `test_clone_propagates_config_folder_to_target` covers only the non-git-branching shape.

src/keboola_agent_cli/services/sync_service.py
```diff
@@ -1061,6 +1078,10 @@ def pull(
             "jobs_written": jobs_written,
             "storage": storage_stats,
             "details": pull_details,
+            # True when the config-folder lookup failed: folders were kept from
+            # the previous pull, not refreshed. Lets a caller tell "no folders"
+            # from "the lookup failed" (CLI-9).
+            "folder_lookup_failed": folder_map is None,
```
zajca, line 1084: Reviewed by Sonnet.

`pull()` now returns `folder_lookup_failed` so callers can tell a stale-folder pull from a real "no folders" pull, but `_format_pull_result()` in `commands/sync.py` (unchanged by this PR) never reads this key. In non-JSON mode (the default), a user whose folder lookup fails (API error, unresolved branch, or bad body) sees a normal success message with no indication that config folders were not refreshed this run -- the only trace is a `logger.warning` call that most interactive users never see. Since this flag exists specifically to let a caller distinguish the two cases, leaving the primary human-facing output path silent defeats part of the fix's purpose; consider printing a warning line (e.g. `formatter.console.print("[yellow]Warning: config-folder lookup failed; existing folders kept[/yellow]")`) when the flag is true.

zajca (changes requested):

soustruh (commented):
# Review of #753 — fix(sync): keep the config folder on pull and clone (CLI-9)

## Summary

This PR captures a config's Keboola UI folder (`KBC.configuration.folderName`) during `sync pull` and makes `repoint_manifest_project` re-point every cloned config's `branch_id`, so `sync clone`'s create-path writeback can match its placeholder and forward that metadata to the target. Both findings from the prior human review (clone never recreating the folder; a degraded folder lookup wiping every folder) are genuinely fixed for the plain, non-git-branching, no-override clone – confirmed by running the PR's own new tests and by independently driving the real `sync_service.py` / `_sync_clone.py` / `sync/clone.py` functions myself, not by trusting the commit message. The fix for the first finding does not hold for a git-branching-enabled reference tree, which is a real, already-documented `sync clone` scenario in this repo (`sync init --git-branching`): the new single `config_branch_id` parameter is stamped onto every manifest entry regardless of which branch it actually lives in or what `_resolve_branch_id` will return at push time, and this reproduces the exact duplicate-entry / dropped-metadata symptom the PR claims to have closed, plus a separate directory-resolution break for `--branch` clones combined with bucket-map/variable-values/instance-rename overrides. Verdict is REQUEST CHANGES because of those two BLOCKING findings; the second original finding (degraded lookup) and its tests hold up on independent verification.

## Verdict

- **Blocking findings**: 2
- **Nits**: 1

## Blocking findings

### `[B-1]` `src/keboola_agent_cli/sync/clone.py:96-98` (`repoint_manifest_project`) / `:101-102` (`_config_dir`) — one `config_branch_id` stamped onto every config breaks directory resolution for git-branching / multi-branch source trees

`repoint_manifest_project`'s new loop (`for cfg in manifest.configurations: cfg.branch_id = config_branch_id`) sets the SAME branch id on every copied config, but `_config_dir` resolves a config's on-disk directory via `branch_map.get(branch_id, _DEFAULT_BRANCH_DIR)`, where `_DEFAULT_BRANCH_DIR` is the literal string `"main"`, not `manifest.branches[0].path` the way `sync/branch_scope.py::branch_tree_path` (the module built specifically because "a single manifest can hold entries from multiple branches") already does it correctly.

Reproduced by driving the real functions: a `--branch 777` clone of a git-branching source whose branch-0 directory is `master` (`init_sync`'s git-branching path names it via `get_default_branch(project_root)`, not literally `"main"`) resolves every override lookup to `<target>/main/...` instead of the real `<target>/master/...`. Separately, a multi-branch reference tree (production id 100 path `"main"`, plus a dev-branch config id 200 path `"dev-feature"` – the layout `#649`'s own module docstring documents as the normal result of `sync pull --branch <dev>`) cloned with **no** `--branch` at all stamps the dev-branch config with the SAME numeric id as production, so it too resolves to `"main"` instead of its real `"dev-feature"` tree:

```
AFTER clone repoint:  id -> (branch_id, resolved dir) {'prod-cfg': (555, 'main'), 'dev-cfg': (555, 'main')}
```

Consequence: `apply_bucket_map` / `apply_variable_values` silently no-op (the file at the wrong path does not exist, so the rewrite counters stay 0, no error reported); `apply_instance_rename`'s `if src.exists(): ...` guard skips the physical move but rewrites `cfg.path` unconditionally, so the manifest and on-disk tree diverge for that config, and the top-level clone result still reports `status: cloned`.

Fix direction: resolve each config's directory the way `branch_tree_path` does (id match, else `branches[0].path`, never a hardcoded literal), and/or restrict the new `config_branch_id` stamping to configs that were already on the source's own default branch instead of every entry.

### `[B-2]` `src/keboola_agent_cli/services/_sync_writeback.py:145` (`target_branch = branch_id or 0`) — the folder is still lost for a git-branching production clone; the exact bug this PR claims to fix reproduces on HEAD

For a git-branching-enabled reference tree cloned with no `--branch` override, `SyncService._resolve_branch_id` (`sync_service.py:2098`) returns `None` at push time (git-branching path: current git branch equals the default branch, whose `branch-mapping.json` entry, written verbatim by `init_sync` and copied verbatim by `copy_reference_tree`, maps to `keboola_id: null`). `writeback_create_config_in_manifest` then computes `target_branch = None or 0 = 0`, but the placeholder entry's `branch_id` was stamped with the numeric `target_default_branch_id` by `_sync_clone.py:136-138`, so the match fails. Reproduced end-to-end against the real functions (no mocks of the logic under test):

```
post-repoint: branches=[(555, 'master')] cfg.branch_id=[555]
SyncService._resolve_branch_id (push-time) -> None
writeback previous_id (empty = appended, not matched): ''
manifest.configurations count: 2
  entry: ext-golden branch_id=555 KBC.*: {'KBC.configuration.folderName': 'Extractors'}
  entry: NEW-ULID-1 branch_id=0 KBC.*: {}
```

A duplicate entry is appended with no `KBC.*` metadata, and the original placeholder (which does carry the folder) is never touched again – identical to the pre-fix symptom. Neither `test_clone_propagates_config_folder_to_target` nor either `repoint_manifest_project` unit test uses a `git_branching`-enabled manifest, so nothing in the new suite catches this. The reworded `gotchas.md:766-770` ("the folder is recreated in the target") is consequently still an over-promise for this reachable case, the same category of problem the prior wording had, just narrower.

Fix direction: derive `config_branch_id` in `clone_project` from what push will actually resolve for the target (call `SyncService._resolve_branch_id` there, the way the new docstring already claims happens), or normalize `writeback_create_config_in_manifest`'s comparison the way `branch_tree_path` does instead of a raw `branch_id or 0`.

## Non-blocking findings

### `[NB-1]` `src/keboola_agent_cli/commands/sync.py:162` (`_format_pull_result`) — `folder_lookup_failed` is never read in human-mode `sync pull` output

The new `folder_lookup_failed` field exists specifically so a caller can tell "no folders configured" from "the lookup failed," but `_format_pull_result` destructures `details` / `storage` / `jobs_written` / `configs_pulled` and never touches this key: a plain (non-`--json`) `sync pull` whose lookup degrades shows an ordinary success message, with only a `logger.warning` call most interactive users never see. `--json` mode is unaffected: `commands/sync.py` emits `formatter.output(data)` with the raw dict, `folder_lookup_failed` included. Fix: print a one-line warning in `_format_pull_result` when the flag is `True`.

### `[NB-2]` `tests/test_sync_storage_jobs.py` (`TestPullConfigFolder`) — no test pins that a genuinely successful, empty lookup drops a previously-captured folder

`sync_service.py:991-998` is written correctly (`folder_map is not None` and empty means `.get(lookup_key)` returns `None`, so the metadata key is omitted – verified by reading the code), and `TestFetchConfigFolders::test_empty_dict_on_successful_empty_lookup` proves the helper itself returns `{}` on a genuine empty result. But no `TestPullConfigFolder` test drives a full pull that captures a folder, then re-pulls with a real (non-failing) empty map and asserts the key is gone from the manifest: the mirror image of `test_pull_preserves_folder_when_lookup_fails`, which only exercises the failure branch. Add one.

## Nits

- `[NIT-1]` PR description (GitHub PR body): still describes the pre-review design ("`sync clone` recreates the folder in the target with no push-side change") and does not mention the branch-repoint fix, `folder_lookup_failed`, or the still-open git-branching gap (`[B-2]`). Worth updating before merge so a reviewer reading only the description has the current picture.

## Verification log

- Confirmed working tree: `/home/soustruh/cc-dev/cli-9`, branch `fix/cli-9-config-folder`, `HEAD=51b51b0b4390e0270a26552ce96d4a8c35347765`, `git status` clean; matches the PR head and the branch given in the task. `gh auth status` -> authenticated as `soustruh` ✓
- Read `CONTRIBUTING.md` ("Checklist: Adding a New CLI Command", "Plugin synchronization map", "Releasing a new version"), `CLAUDE.md` (convention #17, `## All CLI Commands`), `plugins/kbagent/agents/keboola-expert.md` §1 and §3 in full before touching the diff, per the required-reading step.
- `gh pr view 753 -R keboola/cli --json ...` -> OPEN, `fix(sync):` prefix matches a genuine bug fix, +399/-5 across 6 files ✓
- Discovered local `main` was 3 commits stale vs `origin/main` (missing already-merged PR #749); redid every base comparison against `origin/main` (`git fetch origin main`, `git merge-base origin/main HEAD` = `99a000b`, matching `gh pr view`'s file list and `+399/-5` exactly), avoiding misattribution of PR #749's `CLAUDE.md` / `commands-reference.md` changes to this PR.
- `git diff origin/main...HEAD -- src/keboola_agent_cli/cli.py 'src/keboola_agent_cli/commands/**/*.py'` -> empty: no command/flag added, so `commands/context.py`, `CLAUDE.md` "All CLI Commands", `permissions.py` `OPERATION_REGISTRY`, `commands-reference.md`, and the keboola-expert.md Tool Selection Matrix (which already carries a `sync clone` row) are correctly untouched ✓
- Fetched full PR review history via `gh api repos/keboola/cli/pulls/753/reviews` and `/comments`: prior human review by `zajca` at commit `3c5801b` raised the two findings in this task; a second `zajca`-attributed review at commit `51b51b0` (today, `2026-09-14T09:07:41Z`, about 15 minutes after the fix commit) already posted three inline findings (`_sync_clone.py:137`, `:138`, `sync_service.py:1084`) that correspond to `[B-1]` / `[B-2]` / `[NB-1]` below. My findings corroborate that open thread but are independently reproduced against the real functions, not relayed from it.
- Layer-violation greps (typer/click/formatter in services, httpx/requests in commands) -> empty ✓ (no commands/services boundary touched)
- Convention greps (magic numbers, raw `error_code=` strings, bare `except:`, `print()`, token literals) -> empty; the only "token" hits are the pre-existing `client_factory=lambda url, token: ...` test-fixture parameter name, not a leak ✓
- Backward-compat check: no `SyncPullResult`/similar typed model wraps `pull()`'s dict (grep of `result_models.py`), so the new `folder_lookup_failed` key is a safe additive field; `grep -rn folder_lookup_failed src/ tests/ plugins/ docs/` shows only the service, its own tests, and the gotchas.md entry, no stale references ✓
- `make check` -> **exit 0**, `6582 passed, 12 skipped, 5 warnings`; stage markers for ruff check/format, `ty check` (1 pre-existing, PR-unrelated diagnostic in `scripts/hatch_build.py`, a file this PR does not touch), `check_command_sync.py` ("OK: all 268 CLI commands registered and documented"), `check_error_codes.py`, `check_sentinel_guards.py`, `check_file_size.py` (no touched file over budget), `check_version_gates.py` all present and green ✓
- `gh pr checks 753 -R keboola/cli` -> all 3 checks (`check`, `test (3.12)`, Windows wheel build) pass ✓
- `uv run pytest tests/test_sync_clone.py::TestCopyAndRepoint tests/test_sync_clone.py::TestCloneProjectOrchestration::test_clone_propagates_config_folder_to_target tests/test_sync_storage_jobs.py::TestPullConfigFolder tests/test_sync_storage_jobs.py::TestFetchConfigFolders -v` -> all 17 tests pass ✓
- Wrote and ran three standalone repro scripts importing the actual repo modules (`keboola_agent_cli.sync.clone`, `.services.sync_service`, `.services._sync_writeback`) with real `Manifest` / `ManifestConfiguration` objects and a real git checkout (for `_resolve_branch_id`'s `get_current_branch` / `load_branch_mapping` calls); outputs quoted verbatim in `[B-1]` and `[B-2]` above. Also confirmed `scope_manifest`'s classification (via `branch_tree_path`) still correctly treats the copied configs as in-tree / "added" in both the plain and git-branching scenarios: the failure is isolated to `_config_dir` and `writeback_create_config_in_manifest`'s cruder branch-identity checks, not to the diff/push classification itself (task item (c) holds; no `orphaned[]` regression).
- Confirmed the pre-existing CLI-5 stale-branch fix (`default_branch_id`, #744) is unchanged by this PR and not affected by the new findings; it re-points `manifest.branches[0].id` exactly as before, and the new `config_branch_id` parameter is additive.
- Confirmed finding #2 (degraded lookup): read `sync_service.py:991-998` and `:1084`, ran `test_pull_preserves_folder_when_lookup_fails` and `test_pull_reports_folder_lookup_ok` (both pass); preservation on failure and correct `folder_lookup_failed` reporting both verified. The "genuine empty drops a removed folder" direction is correct by inspection but not test-pinned (`[NB-2]`).
- Confirmed `(since vNEXT)` sits on its own body line (`gotchas.md:755`), not in the ATX heading (`gotchas.md:753`); compliant with CONTRIBUTING.md's "Version tags must stay out of markdown headings" rule.
- Checked `plugins/kbagent/agents/keboola-expert.md` for any mention of config folders / `folderName`: none, in §1, §2, or §3; the existing "Provision a new project from a golden reference" matrix row (line 119) already covers `sync clone` as a command and makes no folder-specific claim, so no matrix update is required here, consistent with the precedent set when the push-side fix (v0.47.0) also did not touch this file.
- Could not reproduce against a live Keboola project (no `E2E_API_TOKEN` / `E2E_URL`, no `/tmp/kbagent-e2e` in this environment); the findings above are branch-id / path bookkeeping bugs that do not require a live API response to reproduce, so the direct-function repros above are a legitimate substitute for this specific bug class.

(none)

src/keboola_agent_cli/services/_sync_clone.py
soustruh: Fixed in 8ea7203. The clone no longer stamps a branch id that is absent from `manifest.branches`. It sets the resolved push branch as `branches[0].id` (for `--branch` too). `_config_dir` now falls back to `branches[0].path`, not a literal `"main"` (new `_default_branch_dir`, mirroring `branch_tree_path`). A new test clones with `--branch` and a `bucket_map` on a `"master"` dir through the REAL push. It asserts the bucket is rewritten and the folder reaches the target.

src/keboola_agent_cli/services/_sync_clone.py
soustruh: Fixed in 8ea7203. `config_branch_id` is now derived from `_resolve_branch_id(...)` and normalized to `0` for production, so `0`, `None`, and the numeric default all match the writeback. The clone re-points only the source-default-branch configs, so dev-branch entries keep their tree. A new test drives a git-branching production clone through the REAL push and asserts `set_config_metadata` receives the folder. The created entry sits on branch `0` with no duplicate.

src/keboola_agent_cli/services/sync_service.py
soustruh: Fixed in 8ea7203. `_format_pull_result` prints a warning when `folder_lookup_failed` is true. A new test also pins the mirror case: a successful empty lookup drops a folder removed on the remote.

soustruh (commented):
# Review of #753 — fix(sync): keep the config folder on pull and clone (CLI-9)

## Summary

This is the third review pass on CLI-9, re-reviewing commit `8ea7203` against the four issues raised by a human review (zajca) and a prior kbagent re-review (id 5196362682) on commit `51b51b0`: B-1 (a `--branch` clone with overrides orphaned the config because `_config_dir` fell back to a hardcoded `"main"`), B-2 (a git-branching production clone still dropped the folder because the writeback's `branch_id or 0` comparison never matched the stamped numeric id), NB-1 (`folder_lookup_failed` was invisible in human `sync pull` output), and NB-2 (no test pinned that a successful-but-empty lookup drops a previously-captured folder). All four are independently verified fixed by reading the code, tracing the exact value flow by hand, and driving the real functions (not the PR's own assertions) via the existing test suite plus a standalone reproduction script. One residual, narrower gap remains in the NB-1 fix (silent in the default `--all-projects` path), reported below as non-blocking. Verdict is COMMENT: no blocking findings, one non-blocking gap worth a follow-up.

## Verdict

- **Verdict**: COMMENT
- **Blocking findings**: 0
- **Nits**: 1

## Status of the four original findings

- **B-1 (branch-dir resolution for `--branch` + overrides) — FIXED.** `sync/clone.py:109-116` (`_default_branch_dir`) now returns `manifest.branches[0].path` instead of a literal `"main"`, and `_sync_clone.py:138-145` stamps `manifest.branches[0].id` with the override branch id unconditionally (not only when no `--branch` is given), so `branch_path_map` resolves the override id to the real branch directory (e.g. `"master"`) directly, with the fallback as a second line of defense. Verified by running `tests/test_sync_clone.py::TestCopyAndRepoint::test_config_dir_falls_back_to_default_branch_path` and `::TestCloneProjectOrchestration::test_clone_branch_override_resolves_non_main_dir` — the latter drives a REAL push against a `"master"`-rooted reference tree with `branch_override=777` and a `bucket_map`, and asserts both the rewritten file content and `bucket_rewrites == 1`.
- **B-2 (git-branching production clone drops the folder) — FIXED.** `_sync_clone.py:152-159` now derives `push_branch_id` from `service._resolve_branch_id(...)` (the same resolver `push()` itself uses) and calls the new `repoint_default_branch_configs` (`sync/clone.py:89-107`) with `new_branch_id=push_branch_id or 0`, so a git-branching production clone (where `_resolve_branch_id` returns `None`, confirmed by reading `sync_service.py:2116-2121`) stamps the config's `branch_id` to `0` — matching `_sync_writeback.py:145`'s `target_branch = branch_id or 0` exactly. Verified by running `test_clone_propagates_folder_git_branching`, which monkeypatches `git_utils.get_current_branch` to simulate the git-branching-default-branch case and drives a REAL push; it asserts `set_config_metadata` fires with the folder AND that the created manifest entry sits on `branch_id == 0`.
- **NB-1 (human `sync pull` silent on a degraded lookup) — FIXED for the reported location, gap found in an adjacent path.** `commands/sync.py:164-168` now prints a warning in `_format_pull_result` when `folder_lookup_failed` is true — confirmed both by reading the code and by an independent CliRunner-driven script (see Verification log) for single-project `sync pull`. See `[NB-1-residual]` below for a path this fix does not cover.
- **NB-2 (no test for successful-empty-lookup dropping a folder) — FIXED.** `tests/test_sync_storage_jobs.py:1581-1614` (`test_pull_drops_folder_on_successful_empty_lookup`) is new in this commit, pulls once to capture a folder, pulls again with a real (non-failing) empty `list_config_folder_metadata` map, and asserts the manifest key is gone. Passes; the underlying merge logic (`sync_service.py:991-998`) was already correct, this only closes the test gap.

## Regression watch (multi-branch trees, CLI-5 #744, subsequent push)

- **Multi-branch source trees: no regression, actually improved.** The prior commit (`51b51b0`) stamped `cfg.branch_id` on **every** manifest config unconditionally (including dev-branch entries) via `config_branch_id`. `8ea7203` replaces this with `repoint_default_branch_configs`, which only re-points configs whose `branch_id` matched the *source's own* default branch, captured before the re-point (`_sync_clone.py:126`). A dev-branch config keeps its original branch id and its own on-disk tree. Confirmed via `test_repoint_default_branch_configs_stamps_only_source_default` (a 2-config manifest, prod + dev, only prod moves).
- **CLI-5 / #744 (skip the extra `list_dev_branches` call when `--branch` is given): unaffected.** `_resolve_branch_id` makes no API calls (pure function over the manifest/git state), so adding the new call in `_sync_clone.py:152` does not reintroduce the extra API round-trip. `test_clone_with_explicit_branch_skips_default_branch_fetch` still passes and still asserts `list_dev_branches.assert_not_called()`.
- **Subsequent `sync push` targeting after a `--branch` clone: consistent, not a regression.** Because `manifest.branches[0].id` is persisted as the override id, a later plain `sync push` (no `--branch`) falls through `_resolve_branch_id`'s priority chain to the same `manifest.branches[0].id` and keeps targeting the branch the clone was aimed at — intentional, not accidental drift.

## Non-blocking findings

### `[NB-1-residual]` `src/keboola_agent_cli/commands/sync.py:481` (`_format_all_results`) / `:392-416` (`_pull_one_liner`) — the new folder-lookup warning is silent in the default `--all-projects` human-mode path

`_format_pull_result` (`:162-168`) is only invoked by `_format_all_results` when `verbose` is true (`elif verbose and per_project_formatter:`, line 481); the default (non-verbose) `--all-projects` path renders only `_pull_one_liner` (`:392-416`), which never reads `folder_lookup_failed`. I verified this by driving the real CLI with a mocked `SyncService.pull_all` returning `folder_lookup_failed: True`:

```
single-project pull (human):                    contains warning: True
--all-projects pull, non-verbose (human):       contains warning: False   <- gap
--all-projects pull, --verbose (human):         contains warning: True
```

A fleet-wide `sync pull --all-projects` (no `--verbose`) — plausibly the more common invocation for multi-project setups — gives no human-visible signal that folder metadata was not refreshed for one or more projects; `--json` is unaffected (the flag is still in the raw per-project dict). This is a narrower version of the same original NB-1 gap, in a path the fix didn't touch. Fix: have `_pull_one_liner` flag `folder_lookup_failed`, or always call `_format_pull_result` (not gated on `verbose`) when any project's lookup failed.

(none)

## Nits

- `[NIT-1]` `plugins/kbagent/skills/kbagent/references/commands-reference.md:336,338` — the `sync pull` / `sync clone` entries do not mention the new config-folder round-trip behavior, even though `gotchas.md` was updated. This file's own precedent (e.g. the "Ignored components (since 0.91.0)" and "Since 0.91.0 (#686)" notes on the same two lines) shows it already carries non-flag behavior additions like this one; CONTRIBUTING.md's trigger column for this file literally says "flag changes" (not met here), so this is optional, not a merge blocker, and `gotchas.md` already carries the authoritative version-tagged description for AI agents.

## Verification log

- Confirmed working tree: `/home/soustruh/cc-dev/cli-9`, branch `fix/cli-9-config-folder`, `HEAD=8ea7203` (`git log --oneline -1`) ✓ — matches `<branch>`/PR head.
- Read `CONTRIBUTING.md` ("Checklist: Adding a New CLI Command", "Plugin synchronization map", "Releasing a new version"), `CLAUDE.md` (convention #17, `## All CLI Commands`), `plugins/kbagent/agents/keboola-expert.md` §1–§3, per the required-reading step.
- `gh pr view 753 -R keboola/cli --json title,body,files,...` → OPEN, `fix(sync):` prefix matches a genuine bug fix, +637/-12 across 7 files ✓.
- `git log --oneline main..HEAD` → 4 commits (`64293c8` oldest fix, `3c5801b` docs, `51b51b0` first re-point fix, `8ea7203` second fix, this review's subject) ✓.
- `git diff 51b51b0..8ea7203` → 6 files, 300(+)/69(-): `gotchas.md`, `commands/sync.py`, `services/_sync_clone.py`, `sync/clone.py`, `tests/test_sync_clone.py`, `tests/test_sync_storage_jobs.py`. `services/sync_service.py` untouched by this commit (the pull-side folder logic predates it and was not in question).
- Layer-violation greps (typer/click/formatter in services; httpx/requests in commands) on the full PR diff → empty ✓.
- Convention greps (magic numbers, raw `error_code=` literals, bare `except:`, `print()`, token literals) on the full PR diff → empty; the only "token" hits are pre-existing `client_factory=lambda url, token: ...` test-fixture parameter names ✓.
- No new/removed/renamed CLI command in this PR (`commands/sync.py`'s only change is the 5-line warning print) → `OPERATION_REGISTRY`, `commands/context.py`, `CLAUDE.md` "All CLI Commands", and `keboola-expert.md`'s Tool Selection Matrix correctly require no update; confirmed by `make check`'s command-sync gate.
- `make check` → **exit 0**. `ruff check`: all checks passed. `ruff format --check`: 563 files already formatted. `ty check`: 1 diagnostic, a pre-existing `unresolved-import` warning in `scripts/hatch_build.py` (a file this PR does not touch; downgraded rule, non-blocking per CLAUDE.md convention #`ty`). `check_command_sync.py`: OK, 268 commands registered + documented. `check_error_codes.py`, `check_sentinel_guards.py`, `check_version_gates.py`, `check_file_size.py`: all OK (file-size WARNs are pre-existing, unrelated files). Full suite: **6586 passed, 12 skipped, 5 warnings**.
- `gh pr checks 753 -R keboola/cli` → `check` pass, `test (3.12)` pass, Windows wheel build pass ✓.
- `uv run pytest tests/test_sync_clone.py::TestCopyAndRepoint tests/test_sync_clone.py::TestCloneProjectOrchestration tests/test_sync_storage_jobs.py::TestPullConfigFolder -v` → **22 passed**, including the three real-push clone tests (`test_clone_propagates_config_folder_to_target`, `test_clone_propagates_folder_git_branching`, `test_clone_branch_override_resolves_non_main_dir`) and all five `TestPullConfigFolder` tests.
- `uv run pytest tests/test_sync_clone.py tests/test_sync_storage_jobs.py tests/test_sync_service.py -q` → **207 passed**, no collateral breakage in the wider sync test files.
- Read `_sync_writeback.py:118-166` (`writeback_create_config_in_manifest`) and `sync_service.py:2098-2161` (`_resolve_branch_id`) directly and hand-traced the exact numeric values for all three clone shapes (plain, `--branch`, git-branching production) against the writeback's `branch_id or 0` match — matches the test assertions and the two prior reviews' independent reproductions.
- Did NOT hand-mutate the working tree to check "test teeth" (e.g. reverting to the old code to confirm the new tests fail on it) — that would violate the read-only mandate on this worktree; teeth were instead established by hand-tracing the exact old-vs-new numeric mismatch (confirmed identical to the two prior independent reviews' own reproduction output) and by the fact that `repoint_default_branch_configs` / `_default_branch_dir` are wholly new functions with dedicated unit tests.
- Wrote and ran an independent, standalone script (not part of the repo) driving the real Typer app via `CliRunner` with a mocked `SyncService` to check `folder_lookup_failed` rendering across all three pull output paths (single-project, `--all-projects` default, `--all-projects --verbose`) — found the `[NB-1-residual]` gap above.
- Fetched full PR review history (`gh api repos/keboola/cli/pulls/753/reviews` + inline comments): confirmed the four issues match zajca's reviews and kbagent re-review `5196362682` verbatim; confirmed the PR description was already updated after that review's NIT-1 (no longer describes the stale pre-fix design) — no further action needed there.
- Could not reproduce against a live Keboola project (no `E2E_API_TOKEN`/`E2E_URL`, no `/tmp/kbagent-e2e` in this environment); all four issues are manifest/branch-id bookkeeping bugs fully reproducible by driving the real Python functions and the real Typer CLI (via `CliRunner`) without a live API, which was done throughout.

(none)

src/keboola_agent_cli/commands/sync.py
```diff
@@ -161,6 +161,11 @@ def sync_init(
 
 def _format_pull_result(formatter: Any, result: dict) -> None:
     """Format a single-project pull result for human output."""
+    if result.get("folder_lookup_failed"):
```
zajca, line 164: Reviewed by Sonnet.

The new `if result.get("folder_lookup_failed")` block in `_format_pull_result` (the verbose/single-project human formatter) has no dedicated test asserting the warning text is printed, unlike its counterpart in `_pull_one_liner` which gained `test_pull_one_liner_flags_folder_lookup_failed`. The logic is simple and low-risk, but per this repo's own convention of testing all new behavior, a quick formatter-level test (e.g. asserting `formatter.console.print` is called with the warning when `folder_lookup_failed` is True, and not called when False/absent) would close the gap and guard against a future refactor silently dropping it.

zajca (changes requested):

src/keboola_agent_cli/commands/sync.py
soustruh: Fixed in 527d0aa. Added a formatter-level test for the `_format_pull_result` warning. It asserts the warning prints when `folder_lookup_failed` is true, and stays silent when the flag is false or absent.

Also added a clone regression guard, `test_clone_folder_no_duplicate_manifest_entry`. It pins the original symptom — after a clone:
- the manifest keeps exactly one entry
- push POSTs the folder
- the entry sits on the push branch

zajca (approved):

PR #757, opened 2026-09-11
feat(telemetry): add CLI version and agent/human marker to usage events (CLI-12)

soustruh:
## Why

The telemetry team (KIDS-522) reports on kbagent usage from the `ext.keboola.cli.` events. Their report needs the CLI version and an agent-vs-human marker. The stored event body carried neither. Elasticsearch keeps the request body, not the server-stamped envelope or the headers, so neither value reached the team.

## What

- `_send_event` writes `params.cliContext` on every usage event. It holds `userAgent` (the `keboola-cli/<version> (<os>; <arch>; <impl> <pyver>)` User-Agent, which carries the version and platform) and, when set, `conversationId`.
- The CLI path reads the conversation id from `KBAGENT_CONVERSATION_ID` / `--conversation-id`. The serve middleware reads it from the request's `X-Conversation-ID` header. A request that sends the header is marked. One without it is not.
- `gotchas.md` documents `params.cliContext`, tagged `(since vNEXT)`.

## Tests

- `test_telemetry.py`: the success event carries `cliContext.userAgent`, and a set `KBAGENT_CONVERSATION_ID` lands in `cliContext.conversationId`.
- `test_serve_telemetry.py`: the request's `X-Conversation-ID` reaches `send_serve_event`, and its absence sends `None`.

## Related

- Ticket: CLI-12 (blocks KIDS-522).
- Event schema: keboola/event-schema#33.

### DIFF
--- a/src/keboola_agent_cli/server/app.py
+++ b/src/keboola_agent_cli/server/app.py
@@ -708,6 +708,15 @@ class _UsageTelemetryMiddleware:
             values = parse_qs(scope.get("query_string", b"").decode("latin-1")).get("project")
             project_alias = values[0] if values else None
 
+        # The caller's X-Conversation-ID (set by kbagent's own clients) rides into
+        # params.cliContext so the telemetry reader tells an agent request from a
+        # human one; a web-UI request sends no such header and stays unmarked (CLI-12).
+        conversation_id: str | None = None
+        for key, value in scope.get("headers") or []:
+            if key == b"x-conversation-id":
+                conversation_id = value.decode("latin-1") or None
+                break
+
         await asyncio.to_thread(
             telemetry.send_serve_event,
             self._config_store,
@@ -717,6 +726,7 @@ class _UsageTelemetryMiddleware:
             status_code=status_code,
             duration_s=duration_s,
             project_alias=project_alias,
+            conversation_id=conversation_id,
         )
 
 
--- a/src/keboola_agent_cli/telemetry.py
+++ b/src/keboola_agent_cli/telemetry.py
@@ -5,6 +5,11 @@ events (``POST /v2/storage/events``), the same mechanism the kbc CLI uses. The
 event is stored as ``ext.keboola.cli.`` (CLI) or ``ext.keboola.cli.serve``
 (serve REST API) and stamped server-side with the caller's token + user agent.
 
+The event body also carries ``params.cliContext``: the kbagent User-Agent
+(version + OS/arch/Python) and, when set, the conversation id. The events store
+keeps the request body, not the server-stamped envelope, so a downstream reader
+reads the version and an agent-vs-human marker off the event itself (CLI-12).
+
 This is telemetry, NOT an audit trail: it is voluntary (the two env vars below
 disable it) and best-effort (a failed or blocked events endpoint never affects
 the command). Mutating operations are recorded server-side by Connection as
@@ -27,12 +32,14 @@ import typer
 from .auth.sentinel import is_session_token
 from .config_store import ConfigStore, resolve_config_dir
 from .constants import (
+    ENV_CONVERSATION_ID,
     ENV_DISABLE_TELEMETRY,
     ENV_DO_NOT_TRACK,
     TELEMETRY_COMPONENT_ID,
     TELEMETRY_SERVE_CONFIG_ID,
     TELEMETRY_TIMEOUT,
 )
+from .http_base import build_user_agent
 from .services.base import make_telemetry_client
 from .services.project_service import ProjectService
 
@@ -217,6 +224,7 @@ def emit_cli_invocation(
             duration_s=duration_s,
             configuration_id=None,
             extra_params=None,
+            conversation_id=os.environ.get(ENV_CONVERSATION_ID) or None,
         )
     except Exception as exc:
         logger.debug("usage event failed: %s", exc)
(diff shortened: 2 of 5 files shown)


src/keboola_agent_cli/telemetry.py
```diff
@@ -301,10 +313,17 @@ def _send_event(
...
     extra_params: dict[str, Any] | None,
+    conversation_id: str | None,
 ) -> None:
     params: dict[str, Any] = {"command": command}
     if extra_params:
         params.update(extra_params)
+    # KIDS reads the CLI version (auto-update adoption) and an agent-vs-human
+    # marker off the stored event body, so both ride in params.cliContext (CLI-12).
+    cli_context: dict[str, Any] = {"userAgent": build_user_agent()}
+    if conversation_id:
```
zajca: Reviewed by Opus.

The new marker is purely presence-based: any non-empty conversation id makes the event look like an agent run (`gotchas.md:5019-5021` states this explicitly). But `kbagent serve` unconditionally exports a session id into its own process env — `commands/serve.py:270-271` sets `KBAGENT_CONVERSATION_ID=serve-<UTC-ts>-<8hex>` — and `_default_conversation_id`'s docstring (`commands/serve.py:36-38`) says the `serve-` prefix is precisely how observability filters a **human-driven** session. The serve startup banner (`commands/serve.py:354`, `:373`) then tells the operator to `export KBAGENT_CONVERSATION_ID=serve-...` in a second terminal for their own `kbagent http` workflow.

Failure scenario: a human runs `kbagent serve`, copies the banner's `export KBAGENT_CONVERSATION_ID=serve-20260912T101500Z-ab12cd34` into a second terminal, and runs `kbagent config list`. `emit_cli_invocation` (telemetry.py:227) reads that env var, `_send_event` writes `cliContext.conversationId`, and the KIDS report counts a hand-typed command as an agent run. The same applies to serve's own scheduled `kbagent http` child processes, which inherit the env var. The signal this PR exists to deliver is therefore biased upward for exactly the flow the CLI itself recommends.

Consider either excluding `serve-`-prefixed ids from the marker, or emitting an explicit source/interface field instead of inferring intent from the id's presence — and at minimum documenting the `serve-` caveat in the `gotchas.md` block so the report author does not read presence as agent.

src/keboola_agent_cli/server/app.py
```diff
@@ -708,6 +708,15 @@ async def _emit(self, scope: Any, status_code: int, duration_s: float) -> None:
             values = parse_qs(scope.get("query_string", b"").decode("latin-1")).get("project")
             project_alias = values[0] if values else None
 
+        # The caller's X-Conversation-ID (set by kbagent's own clients) rides into
+        # params.cliContext so the telemetry reader tells an agent request from a
+        # human one; a web-UI request sends no such header and stays unmarked (CLI-12).
+        conversation_id: str | None = None
+        for key, value in scope.get("headers") or []:
+            if key == b"x-conversation-id":
+                conversation_id = value.decode("latin-1") or None
```
zajca, line 717: Reviewed by Opus.

The header value is decoded and forwarded unmodified into `params.cliContext.conversationId` (`telemetry.py:325`), which is POSTed to `/v2/storage/events` and stored in the customer project's event log. There is no length cap and no control-character filtering, even though the module deliberately bounds the other free-text field it stores — `results.error` is truncated at `_MAX_ERROR_LEN = 1000` specifically to keep the event under the events API's 200 KB cap (`telemetry.py:52-54`, `_resolve_error_text`).

Failure scenario: an authenticated serve client (a buggy integration, or a caller abusing the endpoint) sends `X-Conversation-ID` with several KB of arbitrary bytes — up to the ~16 KB h11 header budget, including control characters that survive the latin-1 decode. Every mutating request then writes that blob into the project's Storage event log, bloating events and injecting junk into whatever dashboard reads `cliContext`; past the API cap the POST is rejected and swallowed at DEBUG, silently dropping serve telemetry.

A short truncation plus a printable-character filter on the value (and, for symmetry, on the `KBAGENT_CONVERSATION_ID` read at `telemetry.py:227`, which is likewise unbounded and unstripped) would match the treatment the module already gives `results.error`.

zajca (changes requested):

src/keboola_agent_cli/telemetry.py
soustruh: Fixed in e17903f (documentation). You're right that presence alone is not a clean agent marker: `kbagent serve` generates a `serve-<ts>-<hex>` id and exports it, so a `serve-` id is a serve session, not an external agent.

I took the "document the caveat" path rather than excluding `serve-` ids or adding a source field, because the event carries the actual id string, not a boolean. A report can already split by the id's shape (`serve-*` vs a harness id vs none), so no signal is lost. Excluding `serve-` ids would instead drop the attribution of serve's own scheduled agent tasks, which are agent-driven.

The `gotchas.md` note now says to read the id's shape, not its mere presence, and names the `serve-` origin explicitly.

src/keboola_agent_cli/server/app.py
soustruh: Fixed in e17903f. `conversationId` now gets the same treatment as `results.error`: `_send_event` runs it through `_clean_conversation_id`, which strips non-printable characters and caps it at 200 chars (`_MAX_CONVERSATION_ID_LEN`). I put the guard in `_send_event` rather than at each read site, so it covers both the CLI env var and the serve `X-Conversation-ID` header in one place. Added `test_conversation_id_is_bounded_and_stripped` to cover it.

zajca (approved):

PR #759, opened 2026-09-14
feat: accept browser-login (session) tokens for the commands whose backend supports them (CLI-13)

soustruh:
## Why

kbagent registers a browser-login project with a `kbc-session://` sentinel token, but only the Storage and Manage clients accepted it. Every other client required a static Storage token and refused a session, even though those backends already accept an OAuth bearer token (rollout under [DMD-1593](https://linear.app/keboola/issue/DMD-1593), Done). The block was only on the kbagent side and was tracked nowhere.

## What changed

`services/base.py` adds `make_session_aware_client_factory(config_store, client_cls)`: it builds any `BaseHttpClient` subclass with `BearerAuth` for a `kbc-session://` project and with a static token otherwise. `make_client_factory` delegates to it. Five service clients (`SchedulerClient`, `StreamClient`, `DataScienceClient`, `MetastoreClient`, `AiServiceClient`) accept `http_auth` and drop the `X-StorageApi-Token` header when it is set, the same as `ManageClient`.

A browser-login project now reaches these interfaces:

- **Scheduler** – `flow schedule`, `flow schedule-remove`
- **Data Streams** – the `stream` command group
- **Data Science** – the `data-app` command group
- **Metastore (semantic layer)** – the `semantic-layer` command group, except `semantic-layer token --encrypt` (see Out of scope)
- **AI Service** – `docs query`, `config new`, `config examples`, `component detail`, `flow new` / `update` / `validate`
- **Storage bucket sharing** – `sharing share`, `sharing unshare`: `resolve_master_token` no longer refuses a session, and the Storage API enforces the master privilege

`config oauth-url` and `data-app git-credentials` already worked for a master session: both check the token's real `isMasterToken` privilege, not whether it is static or a session. No change.

The five clients no longer declare `SESSION_AUTH_FEATURE`, and they are recorded in the `BEARER_CAPABLE_CLIENTS` allowlist that `check-sentinel-guards` enforces. `SESSION_UNSUPPORTED_FEATURES` shrinks to the three paths that still need a static token, and `auth login` / `auth register-projects` disclose that shorter list.

## Still to do

- `kbagent kai`: it goes through the external `kai_client` library, which takes a plain Storage token and has no bearer path, so `kai_service` keeps its static-token guard until the library accepts one.

## Out of scope

- `semantic-layer token --encrypt` stays static-token-only by design: it encrypts the project's own token so a metastore component can store and reuse it later, and a `kbc-session://` sentinel is not a reusable credential.
- The importable SDK (`keboola_agent_cli.Client`) stays static-token-only by design: it is a stateless entry point with no config directory and no session store to read a bearer from.
- Manage API commands (`org setup`, `feature *`, `project refresh` / `invite` / `member-*`). A session has no Manage API access by design ([DMD-1711](https://linear.app/keboola/issue/DMD-1711)); these need a manage token, the same as a static-token project.

## Testing

`make check` passes. `TestSessionAwareServiceClients` builds each of the five clients through the session-aware factory and checks that a static token keeps the `X-StorageApi-Token` header, while a session token drops it and gets a bearer. Live-verified on a real stack: the scheduler (read → 200; the activate write path reaches its permission check → 404, not 403), and `stream` / `data-app` / `semantic-layer` / `docs query` read commands return ok with a session token where the released build refuses with "does not support browser-login".

Refs: CLI-13; backend [scheduler#145](https://github.com/keboola/scheduler/pull/145), [mcp-server#406](https://github.com/keboola/mcp-server/pull/406), [sandboxes-service#625](https://github.com/keboola/sandboxes-service/pull/625), [keboola-as-code#2621](https://github.com/keboola/keboola-as-code/pull/2621).


scripts/check_sentinel_guards.py
```diff
@@ -123,6 +123,13 @@ class a direct `KeboolaClient(stack_url, project.token)` would use --
     "ManageClient",  # Manage: the supported bearer path
     "AuthClient",  # talks to the auth service itself, token=""
     "DeveloperPortalClient",  # own username/password identity, never a project token
+    # CLI-13: these reach their backend over a bearer via
+    # make_session_aware_client_factory; the backends accept OAuth tokens.
+    "SchedulerClient",
```
zajca, line 155: Reviewed by Opus.

The five clients are added to `BEARER_CAPABLE_CLIENTS` (so Check 2 skips them) and their `SESSION_AUTH_FEATURE` declarations are deleted (so the `http_base.py:131` runtime guard is inert for them), but they are **not** added to `PROJECT_CREDENTIAL_CLIENTS` (line 116), which is the only seed Check 4 walks (`_descendants_of(root, set(PROJECT_CREDENTIAL_CLIENTS))`, line 392). These clients are constructed from `project.token`, which is exactly the value that can be a `kbc-session://` sentinel — the situation Check 4's own docstring describes for `KeboolaClient`.

Verified on this head commit:

```
$ uv run python -c "from keboola_agent_cli.ai_client import AiServiceClient; \
  c=AiServiceClient(stack_url='https://connection.keboola.com', token='kbc-session://10105'); \
  print(repr(c._client.headers.get('x-storageapi-token')))"
'kbc-session://10105'
```

Before this PR the same call raised `SessionAuthUnsupportedError`. Today every call site goes through `make_session_aware_client_factory`, so nothing is broken now — but the PR removes the runtime net without extending the static one, so the next direct construction or hand-rolled default factory lands silently and sends the literal sentinel string as a credential (an opaque 401), which is precisely the drift `check_sentinel_guards.py` exists to stop.

Suggested fix: add `SchedulerClient`, `StreamClient`, `DataScienceClient`, `MetastoreClient` and `AiServiceClient` to `PROJECT_CREDENTIAL_CLIENTS`. Note that `client/stream.py:34` (`StreamClient(stack_url=..., token=self._token, http_auth=self._http_auth)`) will then be flagged and needs an explicit, reasoned entry in the allowlist — it inherits the parent client's bearer hook and is safe.

src/keboola_agent_cli/metastore_client.py
```diff
@@ -87,15 +87,13 @@ class MetastoreClient(BaseHttpClient):
...
         self._stack_url = stack_url.rstrip("/")
         base_url = self._derive_service_url(self._stack_url, "metastore")
-        headers = {
-            "X-StorageApi-Token": token,
-        }
-        super().__init__(base_url=base_url, token=token, headers=headers)
+        headers: dict[str, str] = {}
+        if http_auth is None:
+            headers["X-StorageApi-Token"] = token
+        super().__init__(base_url=base_url, token=token, headers=headers, http_auth=http_auth)
```
zajca, line 103: Reviewed by Opus.

`MetastoreClient._do_request` (lines 104-137) exists specifically because the metastore answers a **valid** non-master token with HTTP 401 `"Failed to create project scope"` — a permission gate, not a credential failure, remapped to `MISSING_MASTER_TOKEN`. Routing the client through `http_auth` puts `BearerAuth.auth_flow` (`auth/token_provider.py:468-475`) underneath it, and that hook retries **any** 401 after calling `provider.force_refresh(token)`, which performs a real refresh-token exchange and rotates the refresh token in `auth.json`.

Failure scenario: a browser-login user whose session is not project-admin runs `kbagent semantic-layer list`. The metastore returns the project-scope 401; `BearerAuth` burns a network refresh + rotation and resends; the second 401 arrives and only then is the diagnosis produced. Every such call costs one pointless rotation and a doubled round trip. Worse, if that forced refresh itself fails (concurrent rotation by another kbagent process, revoked session), the user gets a "session expired — run `kbagent auth login`" error instead of the accurate `MISSING_MASTER_TOKEN` remedy, sending them to re-login for a problem re-login cannot fix.

The same 401-retry semantics already applied to Storage/Manage, but the metastore is the one client with a documented 401 that is *not* an auth failure, so it is worth either short-circuiting the retry for this exception string or noting the interaction where the `http_auth` wiring is added. Related: the remapped message still interpolates `self._masked_token`, which is `"***"` for a session client (`mask_token("")`), and advises `kbagent project edit --token ...`, which is not the right remedy for a session project.

zajca (changes requested):

zajca (approved):

PR #762, opened 2026-09-14
fix(semantic-layer): derive dataset fqn from the table's Storage location (#761)

### LINKED ISSUE #761: semantic-layer: dataset fqn hardcodes a "KEBOOLA" database that does not exist in the project (add dataset / build)
## Summary

`semantic-layer add dataset` and `semantic-layer build` derive a dataset's `fqn` as `"KEBOOLA"."<bucket-id>"."<table>"` with the database name hardcoded. On the Snowflake projects I checked, the project database is `KBC_<STACK>_<PROJECT_ID>` and `"KEBOOLA"` does not resolve. The model still looks valid, but any consumer that pastes the `fqn` into SQL fails (Kai, data apps, AI SQL generation from semantic context).

## Where

- `src/keboola_agent_cli/services/semantic_layer_service.py`, `_derive_fqn` (lines 109-138 in v0.93.1, unchanged on `main`):
  ```python
  return f'"KEBOOLA"."{schema}"."{table}"'
  ```
- `src/keboola_agent_cli/services/_semantic_layer_internals.py` (~line 847 on `main`): the column-type lookup used by `build --types-workspace` / `--auto-types-workspace` queries `"KEBOOLA".INFORMATION_SCHEMA.COLUMNS`, so it presumably fails the same way (not run).

## Reproduction (kbagent 0.93.1)

Verified on two stacks: GCP us-east4 and GCP europe-west3.

1. `kbagent --json storage bucket-detail --project <alias> --bucket-id <bucket-id>` returns `snowflake_database: KBC_USE4_<project-id>` (us-east4) and `KBC_EUW3_<project-id>` (europe-west3).
2. In a read-only workspace of that project, run through `kbagent workspace query`:
   ```sql
   SELECT COUNT(*) FROM "KEBOOLA"."<bucket-id>"."<table>"
   ```
   Result: `SQL compilation error: Database 'KEBOOLA' does not exist or not authorized.`
3. The same query against `"KBC_USE4_<project-id>"."<bucket-id>"."<table>"` (or the `KBC_EUW3_…` equivalent) succeeds in the same workspace.

Existing models are affected too. A model on europe-west3 has all 47 datasets with a `"KEBOOLA"` fqn, and that fqn does not resolve from a workspace in the same project.

## Expected

The dataset `fqn` uses the bucket's real Snowflake database, which `storage bucket-detail` already exposes as `snowflake_database`. The INFORMATION_SCHEMA type lookup should use the same database.

## Suggested fix

- Resolve the database from the bucket detail when deriving the fqn. It is one call per bucket and can be cached per build.
- Add an optional `--fqn` override to `semantic-layer add dataset`.
- Consider a `validate --deep` check that each dataset `fqn` resolves (e.g. `SELECT 1 FROM <fqn> LIMIT 0` via Query Service), so existing models with a broken fqn get detected. I have not checked whether `validate --deep` currently flags such models.

## Workaround

Write the snapshot JSON with the correct `fqn` and load it with `semantic-layer import`. `import` posts item attributes verbatim, so the fqn is preserved.

ondrejhlavacek:
Closes #761.

## Release Notes

- `semantic-layer add dataset` and `semantic-layer build` now read the dataset `fqn` from the table's real warehouse location in Storage. Before, they wrote a hardcoded `"KEBOOLA"` database that exists in no project, so any consumer pasting the fqn into SQL failed (Kai, data apps, AI SQL generation).
- New `--fqn FQN` on `semantic-layer add dataset` (and `fqn` in the REST body) stores a value verbatim.
- `storage table-detail` returns two new keys: `backend_path` and `sql_path`.
- `semantic-layer validate --deep` warns `FQN_MISMATCH` on datasets whose stored fqn differs from the table's location. Models built before this fix are flagged this way.

## Plans for customer communication
None.

## Impact analysis

**Root cause.** `_derive_fqn` returned `"KEBOOLA"."<bucket-id>"."<table>"`. The Snowflake `INFORMATION_SCHEMA` query behind `build --types-workspace` also targeted `"KEBOOLA"`.

**Why not just swap the database name.** I checked this in a read-only workspace on a GCP us-east4 project:

| fqn | result |
|---|---|
| `"KEBOOLA"."out.c-azure-costs-copy"."2025-reservations-final"` | `Database 'KEBOOLA' does not exist` |
| `"KBC_USE4_5024"."out.c-azure-costs-copy"."2025-reservations-final"` | works |
| `"KBC_USE4_5024"."in.c-aws-costs-raw"."reports"` (linked bucket, own database) | `Schema ... does not exist` |
| `"KBC_USE4_337"."in.c-keboola-ex-aws-s3-…"."reports"` (linked bucket, source database and schema) | works |

A linked bucket's tables live in the source project's database and schema. The Storage API returns exactly that in `bucket.backendPath`, so the fix uses it verbatim and never rebuilds the path from the tableId.

**Changes**
- `services/_table_detail.py`: `table_sql_path()` quotes the location per backend. Snowflake gives `"db"."schema"."table"`. BigQuery gives `` `dataset`.`table` ``, with the project prepended when `databaseName` is set; this matches `bucket-detail`. It returns `None` for an unsupported backend, a missing `backendPath`, or a quote character inside an identifier.
- `services/_semantic_layer_fqn.py` (new): `derive_dataset_fqn` raises `VALIDATION_ERROR` when there is no usable location, pointing at `semantic-layer add dataset --fqn` (the same message serves `build`, which has no `--fqn`). `append_fqn_mismatch_warnings` powers the new `validate --deep` warning.
- `add_dataset` fetches table detail once, which also serves `--deep-fields`. With `--fqn` alone it makes no Storage call.
- `build_information_schema_sql` takes `backend_path`. The Snowflake query uses its database and schema. The BigQuery query uses `backendPath[0]` instead of rebuilding the dataset name from the bucket id.

**Behavior changes**
- `add dataset` without `--fqn` now fails when the table does not exist, or when Storage reports no location for it. It used to store a broken fqn silently. The E2E tests that use synthetic tableIds now pass `--fqn`.
- `build` aborts before any write when a table's location cannot be derived.
- Existing models are not rewritten. `gotchas.md` documents the repair path: export, fix the fqn and metric sql, then `import --overwrite`.

**Live verification (kbagent from this branch)**
- `build --dry-run --types-workspace` on a normal table and a linked alias table: both fqns match the working paths above. `type_resolution_errors: []`, and the linked table got its types from `"KBC_USE4_337".INFORMATION_SCHEMA`.
- `validate --deep` on the two europe-west3 models from the issue: 41 `FQN_MISMATCH` warnings on each. The other 6 of the 47 datasets hit the existing `DEEP_FETCH_FAILED`.

**Not covered.** BigQuery was not verified live; no BigQuery project was available. A plain alias table in the project's own (non-linked) bucket was not verified live either: none of the available projects had one. It takes its own bucket's `backendPath`, the same as a regular table. `build` still emits Snowflake `sql_dialect` and column quoting on BigQuery, and its BigQuery `INFORMATION_SCHEMA` lookup omits the GCP project (unchanged from before). Both are tracked as a separate follow-up.

**Reviews on this PR**
- Standards review (sub-agent against CONTRIBUTING.md and CLAUDE.md): no hard violations. Applied: documented the new `table-detail` keys in `AGENT_CONTEXT`, `CLAUDE.md` and `keboola-expert.md`; removed a duplicate quote check; trimmed comments that argued for the change.
- Spec review (sub-agent against #761): the `build` error advised `--fqn`, which `build` does not have. Fixed. A live `SELECT 1 FROM <fqn>` check in `validate --deep` was only suggested in the issue; this PR compares against the Storage location instead, which needs no workspace.
- Gemini 3 Pro second opinion: no defects in the change. Its one note (BigQuery `INFORMATION_SCHEMA` without the GCP project) predates this PR and joins the BigQuery follow-up.

## Change type
Bug fix (plus the additive `--fqn` option and new `table-detail` keys).

## Justification
Every semantic model created through kbagent carried an fqn that resolves nowhere, which breaks every downstream SQL consumer of the semantic layer.

## Deployment
Merge & automatic deploy.

## Rollback plan
Revert of this PR.

## Post release support plan
None.


ondrejhlavacek:
@keboola-pr-reviewer-bot review

soustruh (approved):
Approved. Root-cause fix: the dataset fqn is derived from the table's real Storage backendPath instead of a hardcoded KEBOOLA database. The change is additive and backward-compatible, and CI is green.

PR #763, opened 2026-09-15
fix(release): re-enable the winget release job

Matovidlo:
## Summary
- Re-enables the `winget` job in `.github/workflows/release-kbagent.yml`, restoring its original trigger condition (`needs.version.outputs.IS_PRERELEASE == 'false' && startsWith(github.ref, 'refs/tags/')`) in place of the hardcoded `if: false`.
- Updates the job's comments, which had gone stale: they claimed `Keboola.KeboolaCLI2` had "never had its first submission" to `microsoft/winget-pkgs`, blocking `wingetcreate update` from having anything to bump.
- That's no longer true — the one-time bootstrap `wingetcreate submit` was already done by hand on 2026-09-07 (`microsoft/winget-pkgs` PR #409446), creating the `Keboola.KeboolaCLI2` manifest starting at version `0.79.0`. Nobody flipped the job back on afterward, so every release since (up through the current `0.93.2`) has silently skipped winget publishing even though the blocker was gone.
- `WINGET_TOKEN` is already present as a `release`-environment secret (set 2026-06-15), so no secret changes are needed here.

## Test plan
- Workflow-only change; no application code paths affected.
- Verified via the GitHub API that the `Keboola.KeboolaCLI2` manifest exists on `microsoft/winget-pkgs@master` (`manifests/k/Keboola/KeboolaCLI2/0.79.0`) and that `WINGET_TOKEN` is set in this repo's `release` environment.
- Can't fully dry-run the `winget` job locally (it publishes to an external repo and only runs on a real version tag); the real verification is watching the job on the next tagged release and confirming a new `microsoft/winget-pkgs` PR opens with the correct version/URL/hash.

## Related
- Follow-up to #710 (docs-only PR adding a manual for creating `WINGET_TOKEN`), whose text about the bootstrap submission being outstanding is now stale — this PR is the actual fix that makes the job run again.
- No Linear issue.


soustruh (approved):
Approved. Minimal, scoped fix that re-enables the winget release job. CI is green.

PR #768, opened 2026-09-16
chore(deps): drop pip-audit, gate CI on uv audit, require uv >=0.12.15 (CLI-16)

soustruh:
> **Update (addressing the review):** the audit moved from a step in the `check` job to a **separate, non-blocking** CI job. Superseded lines below are ~~struck through~~ with the revision after them. See the follow-up commit.

## Why

- `pip-audit` entered the `dev` group in v0.6.0 ([dcc5413](https://github.com/keboola/cli/commit/dcc54130e84c3b61e790c6cfe2a1eb289d325585)). No CI step, Makefile target, pre-commit hook, document, or agent instruction ever ran it. The check covered the full git history.
- It is the largest dependency subtree in `uv.lock`: 29 of the 80 locked packages. Nothing else uses 19 of those 29. `pip` itself enters the lockfile only through `pip-audit` and `pip-api`.
- `uv audit` finds the same vulnerability on `main` as pip-audit does. Funnily enough, the vulnerable package is `pip` itself. For a project built on uv, `uv audit` fully replaces pip-audit: it reads `uv.lock` directly and needs no pip. Dependabot already watches the lockfile, so the audit runs as a complement, not a blocking gate.

## What

- `pyproject.toml` drops `pip-audit` from `[dependency-groups].dev`. The re-locked `uv.lock` removes 19 packages and adds none.
- ~~`.github/workflows/ci.yml` adds a `Dependency audit` step to the `check` job (`uv audit --frozen`), right after `uv sync`. The step prints an expected preview warning line.~~ `.github/workflows/ci.yml` adds a **separate `audit` job** that runs `uv audit --frozen`. It is **not** a required check: a finding shows red for visibility but does not block a merge. Its verdict depends on the OSV database over time, not on the diff, so it stays out of the `check` job to keep that job deterministic.
- ~~`Makefile` adds an `audit` target and includes it in `make check`, which the CI job mirrors. `CONTRIBUTING.md` lists it under "Running CI Locally".~~ `Makefile` adds an `audit` target for local use; it is **not** part of `make check`, so `make check` stays offline-capable and mirrors the CI `check` job. `CONTRIBUTING.md` documents the new job under "CI workflows" and lists `make audit`.
- `ci.yml`, `e2e.yml`, `release.yml`, `release-kbagent.yml`, and `.github/actions/setup-build` now pin uv to 0.12.15. `[tool.uv] required-version = ">=0.12.15"` makes 0.12 the new minimum. `uv audit`'s CLI interface differs between 0.11 and 0.12: the `--preview-features` flag value was renamed from `audit` to `audit-command`. `docs/adr/0003` names 0.11.16 as the pin at decision time. This PR leaves that line unchanged.

## Verified

- `uv lock --check` and `uv lock --dry-run` with uv 0.12.15 report no lockfile changes.
- `uv audit --frozen` with uv 0.12.15 on the branch reports no known vulnerabilities in 61 packages and exits 0.
- A cross-check against pip-audit 2.10.1 (`uvx --with pip pip-audit -r <uv export>`) uses both the PyPI and OSV services. It finds the same result: the finding on `main`, nothing on the branch. pip-audit also skips `colorama`, a Windows-only marker. `uv audit` covers it too.
- With uv 0.12.15, all of these pass:
  - `uv sync --extra server`
  - `ruff check`
  - `ruff format --check`
  - `ty check` (only the known, downgraded hatchling `unresolved-import` warning)
  - `uv build --wheel`
  - `make audit`
  - `make -n check`

  `required-version` refuses uv 0.11.16 with a clear message.
- Follow-up (separate-job revision): `uv audit --frozen` works on a clean checkout with no `uv sync`, so the job needs no install step. `ci.yml` parses as valid YAML with jobs `check`, `audit`, `test`, `build-windows`. `make -n check` no longer runs the audit; `make -n audit` does.

## Future

In the future we might consider dropping uv audit entirely in favor of Dependabot, which already runs on this repository. However, as it is a built-in feature of uv now, we will just perform the switch this time and make it at least remotely useful.

Closes CLI-16. Dependabot [#727](https://github.com/keboola/cli/pull/727) (pip 26.1.2 to 26.2) becomes obsolete after merge.


.github/workflows/ci.yml
```diff
@@ -70,6 +70,10 @@ jobs:
         # croniter so the test suite can import `keboola_agent_cli.server.*`.
         run: uv sync --extra server
 
+      - name: Dependency audit
+        # Checks locked packages against OSV; the preview warning is expected.
+        run: uv audit --frozen
```
zajca: Reviewed by Opus.

`uv audit --frozen` is now a hard gate in the `check` job, but its result is not a function of the diff. It queries the OSV service over the network and re-evaluates all 61 locked packages against advisories published *since* the last run, so a newly disclosed CVE in any transitive dependency — or an OSV outage — turns every open PR red with no relation to its contents.

This cuts directly against the design the surrounding comments in this same file state. The job header at lines 37-42 describes these gates as "deterministic and interpreter-independent", and lines 44-48 explain that `changelog-check` is *deliberately* excluded from CI precisely because it needs network and audits state outside the diff ("a release-time concern, not a per-PR one"). A dependency audit has exactly that shape.

There is also no waiver path. Neither the CI step nor the `audit` target in `Makefile:71` passes `--ignore` or `--ignore-until-fixed`, and nothing documents how to handle an advisory with no fixed version available. When that happens the only way to unblock merges is to edit the workflow or Makefile under time pressure. Separately, `make check` is now network-bound, which weakens the "run `make check` before pushing" instruction at `CONTRIBUTING.md:871` for anyone working offline.

Worth considering: run the audit on a schedule (or as a `continue-on-error` / separate non-required job) so a fresh advisory opens a tracking issue instead of blocking unrelated work, and document the `--ignore-until-fixed` escape hatch next to the target.

Minor related note: `uv audit` still self-identifies as experimental ("warning: `uv audit` is experimental and may change without warning"), and this PR itself documents that its CLI changed incompatibly between uv 0.11 and 0.12. Passing `--preview-features audit-command` explicitly would make the preview opt-in deliberate and silence the warning the comment currently apologises for.

CONTRIBUTING.md
```diff
@@ -852,10 +852,11 @@ foot-gun.
 ## Running CI Locally
 
 ```bash
-make check              # CI parity: lint + format + typecheck + skill + version + command-sync + endpoints + changelog + error-codes + sentinel-guards + test
+make check              # CI parity: lint + format + typecheck + audit + skill + version + command-sync + endpoints + changelog + error-codes + sentinel-guards + test
 make lint               # Just the ruff linter
 make format             # Auto-format code
 make typecheck          # Static type check (Astral `ty`)
+make audit              # Audit locked dependencies for known vulnerabilities (uv audit, OSV)
```
zajca, line 859: Reviewed by Opus.

The "Running CI Locally" block was updated here (lines 855 and 859), but the "CI workflows" section further down was not. `CONTRIBUTING.md:881-887` enumerates what the `check` job runs — "lint, format, `ty` type-check, SKILL.md freshness, version consistency, the command-sync and version-gate silent-drift gates, the serve endpoint-reference freshness check, and the error-code enum check" — and still omits the dependency audit that `.github/workflows/ci.yml:73-75` now adds.

That same paragraph also asserts the job's gates "are deterministic and interpreter-independent", which no longer holds once an OSV lookup is part of it (see C001). Given how much this repo invests in keeping documentation and gates in sync, a contributor reading that section will not learn that a CI failure can now come from a dependency advisory rather than from their diff. Adding the audit to the enumeration — and noting why it can fail without any code change — keeps the section accurate.

zajca (changes requested):

zajca (approved):

PR #770, opened 2026-09-16
fix(sync): restore the DS client factory default after CLI-13

Matovidlo:
## Summary
- Fixes `main`'s own tip being CI-red (`check`, `test (3.12)`, and the Windows wheel-build job all failing) since PR #759 (CLI-13, session-token support) merged.
- Root cause: PR #759 replaced `data_app_service.py`'s module-level `_default_ds_client_factory` with `make_session_aware_client_factory(config_store, DataScienceClient)`, but left `sync_service.py` importing the now-removed name. Any code path that imports `keboola_agent_cli.cli` (almost the entire CLI) or the FastAPI server app failed at import time with `ImportError: cannot import name '_default_ds_client_factory'`.
- Fix: `SyncService` now builds its DS client factory default through the same `make_session_aware_client_factory` helper `DataAppService` already uses, so a browser-login session project's Storage/DS calls during `sync pull`/`push` work the same way CLI-13 intended everywhere else. The helper is imported locally inside `__init__` (not hoisted into the top-level `.base` import) specifically to keep `services/sync_service.py` within its grandfathered file-size ceiling — `scripts/file_size_baseline.json` caps it at 1655 code lines, `main` was already at 1654, and a straightforward top-level-import fix would have pushed it to 1661 and failed the file's own `tests/test_check_file_size.py` gate. This version lands at 1653.
- This is what made PR #763 (an unrelated `winget` release-job fix) show red CI after rebasing onto `main` — that PR's diff only touches a GitHub Actions workflow file and was never the cause.

## Test plan
- `uv run ruff check src/ tests/ scripts/`, `uv run ruff format . --check`, `uv run ty check` — all clean.
- `uv run python scripts/check_file_size.py` — `OK: 240 modules within budget`, `services/sync_service.py` at 1653/1655.
- `uv run python scripts/sync_version.py`, `scripts/check_command_sync.py`, `scripts/check_version_gates.py`, `scripts/gen_endpoint_reference.py` — no drift.
- `uv run pytest tests/ -q` — full local suite: **6787 passed, 184 skipped, 0 failed** (previously 62 failed + 124-127 collection errors on `main`).

## Related
- Fixes a regression introduced by PR #759 (CLI-13, session-token support).
- Unblocks PR #763 (`fix(release): re-enable the winget release job`), whose CI was red only because of this bug on `main`.
- No separate Linear issue; part of the CLI-13 ticket's follow-up.


Matovidlo:
Superseded by #764, which fixed this exact same bug (same root cause: sync_service.py importing the removed `_default_ds_client_factory` after #759) and already merged to main. Closing this in favor of that fix.

